PR #17, opened 2017-03-03
Removed numpy directory from include directory

lchauvin:
swiggfaiss_wrap.cxx already include <numpy/arrayobject.h>, if compiled with -I/usr/lib64/python2.7/site-packages/numpy/core/include/numpy, file will be searched at /usr/lib64/python2.7/site-packages/numpy/core/include/numpy/numpy/arrayobject.h and will not be found.


lchauvin:
Done

mdouze:
Thanks for the observation. It turns out that on the machines where I tested it, there was a `/usr/include/numpy/arrayobject.h` file. And indeed the `numpy.get_include()` does not include the `numpy/` part.

Issue #29, opened 2017-03-09
nvcc fatal   : Don't know what to do with '/usr/lib/lapack/liblapack.so.3.0' make: *** [../python/_swigfaiss_gpu.so] 


harveyaot:
I encountered the following error when make [**../python/_swigfaiss_gpu.so**].

> /usr/local/cuda-7.5//bin/nvcc -I..  -I /usr/local/cuda-7.5//targets/x86_64-linux/include/ -Xcompiler -fPIC -Xcudafe --diag_suppress=unrecognized_attribute -gencode arch=compute_35,code="compute_35" -gencode arch=compute_52,code="compute_52" -gencode arch=compute_52,code="compute_52" --std c++11 -lineinfo -ccbin g++ -DFAISS_USE_FLOAT16 -I/usr/include/python2.7/ -I/usr/lib64/python2.7/site-packages/numpy/core/include/ -shared -o ../python/_swigfaiss_gpu.so ../python/swigfaiss_gpu_wrap.cxx libgpufaiss.a ../libfaiss.a \
> 	-Xcompiler -fopenmp -lcublas \
> 	-Xlinker  /usr/lib/libopenblas.so.0 /usr/lib/lapack/liblapack.so.3.0
> nvcc fatal   : Don't know what to do with '/usr/lib/lapack/liblapack.so.3.0'
> make: *** [../python/_swigfaiss_gpu.so] Error 1

**I use cuda-7.5. Is cuda-8.0  necessary for compiling gpu version?**
System is 14.04.1-Ubuntu

harveyaot:
Already resolved by [#issue15](https://github.com/facebookresearch/faiss/issues/15)
Thanks.

PR #30, opened 2017-03-09
quote the multiple BLASLDFLAGS values

harveyaot:
To resolve the [#issues29](https://github.com/facebookresearch/faiss/issues/29), just add quotation marks for multiple BLASLDFLAGS values in example_makefiles/the makefile.inc.Linux


mdouze:
Hi 

Yes it is the second time this issue arises. The other solution is to add -Xlinker in front of each of the BLASLDFLAGS, but I think putting the quotation marks will not break anything.

daniellevy:
The quotes broke it for me, when I tried building (couldn't find the files), but removing the quotes worked fine.

mdouze:
Hmmm...

Thanks for the report. I am going to revert the quotes and add a comment about several -Xlinkers...

PR #75, opened 2017-04-10
add initial cmake support

hlzz:
add the cmake build scripts, some Find scripts copied from caffe (OpenBlas). Currently it only supports openblas as the BLAS library, will add MKL and ATLAS later.

TODO:
1. python support (with swig)
2. more BLAS options.
3. doc generation in cmake, if there is any.


mdouze:
Hi @hlzz 

Thanks for the contribution. Will you be able to support people who use cmake, ie. answer to cmake-specific issues and keep the CMakeFiles up-to-dat wrt. faiss?

Could you add a section to the INSTALL file explaining how to use cmake to build faiss?

@mdouze

hlzz:
Hi @mdouze ,
Ok, I will try my best to support the cmake build. I've also added a section for cmake build.

mdouze:
I tried to use cmake on the mac, but I get 

https://gist.github.com/mdouze/46daf5fb494be70c7dd8ab19698021d0

Any idea what could be wrong?

hlzz:
Hi @mdouze ,
Answered in the gist.

hlzz:
Revised the pr again, now FindOpenBLAS.cmake use openblas_config.h to find OpenBLAS, which exists in both mac and ubuntu. Could you have a check whether macports have that one? Also, can we move the discussion here? I cannot receive notifications from that gist.

mdouze:
Thanks @hlzz 

The cmake compilation compiles on the Mac/macports and on CentOS. As a bonus it compiles the tests that are ignored by the regular cmake.

Could you update the cmake section of the INSTALL saying:
- the BLAS implementation must be openblas
- the python interface is not supported
- how to include the produced library in another cmake project

Then I will accept the PR.

hlzz:
Hi @mdouze ,
The INSTALL is revised as you requested. Thanks.

mdouze:
Sorry for the late reaction @hlzz, I thought I merged the pull request, but I apparently forgot to click  a button.

Thanks for your contribution.

hlzz:
@mdouze Cool, thanks for your work!

PR #76, opened 2017-04-11
Update makefile.inc.Linux

olalonde:


mdouze:
Hi 

Thanks for the PR. Normally installing swig should not be necessary since the swig-generated files are included in the git repo. Why do you think this should be added?

olalonde:
I needed it to compile on a fresh Ubuntu server. You can see that it's also installed in the Dockerfile and used in the Makefile:

https://github.com/facebookresearch/faiss/blob/master/Dockerfile#L5
https://github.com/facebookresearch/faiss/blob/master/example_makefiles/makefile.inc.Linux#L100

Maybe it's not the right place to put it but I thought it might save someone some time.

Issue #77, opened 2017-04-11
Another way to install with docker


jubenjum:
Hi,

In the "docker instructions" section in the INSTALL file, I think you can  also build the docker image by doing:

nvidia-docker build -t 'faiss' https://github.com/facebookresearch/faiss

then you won't need to clone all the repository. 

Juan

mdouze:
Hi 

Thanks for the information. @pletessier, any comment about this?

pletessier:
Hi,
That's true, I wasn't aware of this feature, thanks for the info @jubenjum ! Note that with my version of Docker (1.13.0), I was needed to remove the "protocol part" of the url to make it work. It seems to be the good syntax according to https://docs.docker.com/engine/reference/commandline/build/#build-with-url.
So the command was `nvidia-docker build -t 'faiss' github.com/facebookresearch/faiss`

But, by doing that, you still need to to bring the sources inside the container. Giving an URL in the build step of docker is just giving a build context. 

So the best way to build the image would probably be to ADD the sources from the context (URL or local path) rather than cloning a second time from github. It would also be easier to make changes in the code this way.

It would also be preferable to write an Ubuntu 16.04 dedicated makefile.inc to avoid the "dirty sed replacing" of the LDFLAGS vars.

I will take care of that in a few days if nobody took the ticket !

Pierre

mdouze:
Seems to be taken into account. Closing issue.

PR #88, opened 2017-04-17
Docker improvements

pletessier:
Hi @mdouze ! Here are some changes discussed in #77 with @jubenjum.


mdouze:
Thanks for the update @pletessier!

I am a bit reluctant to make separate makefile examples for different minor versions of operating systems, because it means there is a lot of code duplication between them, which is hard to maintain. In that respect the sed command that you had before was better.

pletessier:
I totally understand your concerns about code duplication, but I'm also concerned about the sed command (Murphy's law...), so here is my new proposal : To use an env var $BLASLDFLAGS in the Dockerfile which will overwrite (if defined) the corresponding var in the makefile.inc.Linux file.

What do you think ?

mdouze:
Thanks, that looks better to me. Could you try with `?=` instead of the ifeq? It is more compact. Also, I don't think it is useful to do the test for commented out code.

pletessier:
Wasn't aware of the `?=` operator, that's nice !

mdouze:
Looks good. Thanks!

PR #133, opened 2017-06-12
added makefile info

mbstrange:
I didn't realize I needed both makefiles.  For more seasoned pros this was probably obvious, but my linux administration skills aren't super sharp so this might be helpful to have.


mdouze:
Sorry, now outdated.

PR #199, opened 2017-08-31
Delete --no-as-needed MKL config. Fixes #198

### LINKED ISSUE #198: GPU build fails with MKL
Specifically, this line will fail:

```make
test/demo_ivfpq_indexing_gpu: test/demo_ivfpq_indexing_gpu.cpp \
	$(LIBNAME).a ../libfaiss.a
	$(NVCC) $(NVCCFLAGS) -o $@ $^ -Xcompiler -fopenmp -lcublas \
	$(BLASLDFLAGSNVCC)
```

With the error:

```
ar: creating libgpufaiss.a
/usr/local/cuda-8.0/bin/nvcc  (many flags)
-Xlinker -Wl,--no-as-needed -L/opt/intel/compilers_and_libraries/linux/mkl//lib/intel64   -lmkl_intel_ilp64 -lmkl_core -lmkl_gnu_thread -ldl -lpthread
/usr/bin/ld: unrecognized option '-Wl'
```

This is because

`BLASLDFLAGSNVCC=-Xlinker $(BLASLDFLAGS)`, and

`BLASLDFLAGS=-Wl,--no-as-needed -L$(MKLROOT)/lib/intel64   -lmkl_intel_ilp64 \
 -lmkl_core -lmkl_gnu_thread -ldl -lpthread`

from `makefile.inc` when one is compiling with MKL.

However, `nvcc` does not recognize the `-Wl,--no-as-needed` option ([see the second comment under this SO post](https://stackoverflow.com/a/7221187/1884420)).

I am not sure what the best option here is. Deleting `-Wl,--no-as-needed` works for me, but I guess we could also add an extra variable? I'll submit a PR with the solution that worked for me, but let me know if you know of a better way.

una-dinosauria:
See https://github.com/facebookresearch/faiss/issues/198

### DIFF
--- a/example_makefiles/makefile.inc.Linux
+++ b/example_makefiles/makefile.inc.Linux
@@ -44,7 +44,7 @@ FAISSSHAREDFLAGS=-shared
 #
 # MKLROOT=/opt/intel/compilers_and_libraries/linux/mkl/
 #
-# BLASLDFLAGS=-Wl,--no-as-needed -L$(MKLROOT)/lib/intel64   -lmkl_intel_ilp64 \
+# BLASLDFLAGS=-L$(MKLROOT)/lib/intel64   -lmkl_intel_ilp64 \
 # -lmkl_core -lmkl_gnu_thread -ldl -lpthread
 #
 # BLASCFLAGS=-DFINTEGER=long


mdouze:
Setting MKL flags is a horrible experience, they depend on the compiler version, the MKL version and the phase of the moon. The current setting works for some people (including me) and other ones are expected to be able to figure out the flags themselves. So I prefer not to touch the current flags.

una-dinosauria:
>  they depend on the compiler version, the MKL version and the phase of the moon

😂 fair enough

Issue #219, opened 2017-09-30
Problem with Clustering::train when nredo>1


Subenle:
In function [Clustering::train](https://github.com/facebookresearch/faiss/blob/master/Clustering.cpp#L121)
At the end of each redo`for (int redo = 0; redo < nredo; redo++)`

there is a judge whether the objective is improved
```
        if (nredo > 1) {
            if (err < best_err) {
                if (verbose)
                    printf ("Objective improved: keep new clusters\n");
                centroids = buf_centroids;
                best_err = err;
            }
            index.reset ();
        }
```

when `nredo > 1`, index.reset() will always be  executed, so the index will be empty instead of saving the centroids.

In my test case, when I set `index.cp.nredo=3;`, the quantizer will be empty after training and index will add nothing into it when execute `index.add(n,xb);`. Some of my output is listed as following:
```
after train: quantizer ntotal: 0, index ntotal: 0
IndexIVFFlat::add_core: added 0 / 100000 vectors


after add: quantizer ntotal: 0, index ntotal: 0
[3.373 s] start search,probe = 1
[3.382 s] end search,probe = 1
I=
query 9995:   -1    -1    -1    -1 
query 9996:   -1    -1    -1    -1 
query 9997:   -1    -1    -1    -1 
query 9998:   -1    -1    -1    -1 
query 9999:   -1    -1    -1    -1 
```

mdouze:
Should be fixed in latest version.

PR #221, opened 2017-10-06
fix bug in Clustering::train when nredo > 1

Subenle:
Fix the bug https://github.com/facebookresearch/faiss/issues/219

when nredo > 1, we should set index with centroids, it saves the best centroids with best_err.

### DIFF
--- a/Clustering.cpp
+++ b/Clustering.cpp
@@ -195,10 +195,12 @@ void Clustering::train (idx_t nx, const float *x_in, Index & index) {
                 centroids = buf_centroids;
                 best_err = err;
             }
-            index.reset ();
         }
     }
-
+    if (nredo > 1) {
+       index.reset ();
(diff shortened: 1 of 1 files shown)


mdouze:
Hi 
Thanks for the fix. 
Could you also add a function to 

https://github.com/facebookresearch/faiss/blob/master/tests/test_build_blocks.py

that verifies that by setting eg. nredo = 10 it gets a lower quantization error than with nredo = 1 ?
Thanks

Subenle:
I find a mistake in my code , `index.reset ();` should not be deleted. I'll fixed this and add a test function to verifies that by setting eg. nredo = 10 it gets a lower quantization error than with nredo = 1.

mdouze:
The nredo implementation was updated in the last version + added a test.

PR #232, opened 2017-10-24
Fix makefile.inc.Mac.brew

Ivanopolo:
Removing old flag `-Dnullptr=NULL` that causes issues with installation on osx with brew.


mdouze:
Hi
Thanks for spotting this!

Issue #72, opened 2017-04-10
ld: library not found for -lomp


olalonde:
On MacOS 10.12.4, I am getting the following:

```
~/code/faiss$  cp example_makefiles/makefile.inc.Mac.brew makefile.inc
~/code/faiss$ make tests/test_blas
/usr/local/opt/llvm/bin/clang++ -fPIC -m64 -Wall -g -O3 -msse4 -mpopcnt -fopenmp -Wno-sign-compare -Dnullptr=NULL -I/usr/local/opt/llvm/include -Doverride= tests/test_blas.cpp -o tests/test_blas -framework Accelerate -DFINTEGER=int
ld: library not found for -lomp
clang-4.0: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [tests/test_blas] Error 1
```

```
/usr/local/opt/llvm/bin/clang++ --version                                                            master
clang version 4.0.0 (tags/RELEASE_400/final)
Target: x86_64-apple-darwin16.5.0
Thread model: posix
InstalledDir: /usr/local/opt/llvm/bin
```

olalonde:
Managed to compile by adding $(LDFLAGS) to the makefile target but running it gives an error:

```
~/code/faiss$ ./tests/test_blas                               fix-issue-72 ✗
dyld: Library not loaded: @rpath/libomp.dylib
  Referenced from: /Users/olalonde/code/faiss/./tests/test_blas
  Reason: image not found
[1]    32263 abort      ./tests/test_blas
```

olalonde:
This seems to magically solve the issue but not sure why

```
install_name_tool -add_rpath /usr/local/opt/llvm/lib tests/demo_ivfpq_indexing
```

```
$ install_name_tool -add_rpath /usr/local/opt/llvm/lib tests/demo_ivfpq_indexing
~/code/faiss$ ./tests/demo_ivfpq_indexing                                       master ✗
[0.000 s] Generating 100000 vectors in 128D for training
[0.197 s] Training the index
Training IVF quantizer on 100000 vectors in 128D
Training IVF residual
  Input training set too big (max size is 65536), sampling 65536 / 100000 vectors
computing residuals
training 4x256 product quantizer on 65536 vectors in 128D
Training PQ slice 0/4
Clustering 65536 points in 32D to 256 clusters, redo 1 times, 25 iterations
  Preprocessing in 0.00 s
[...]

PR #241, opened 2017-11-03
Allow brew version to find libomp.dylib

MisterTea:
llvm on brew doesn't expose libomp.dylib causing build issues (see https://github.com/facebookresearch/faiss/issues/72 ).  This PR fixes that issue.


MisterTea:
Actually this causes warnings, let me find a better way

PR #261, opened 2017-11-22
add a working setup.py

soumith:
as title says

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,51 @@
+from __future__ import print_function
+from setuptools import setup, find_packages
+import os
+import shutil
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+check_fpath = os.path.join("python", "_swigfaiss.so")
+if not os.path.exists(check_fpath):
+    print("Could not find {}".format(check_fpath))
+    print("Have you run `make` and `make py` "
+          "(and optionally `cd gpu && make && make py && cd ..`)?")
+
+# make the faiss python package dir
+shutil.rmtree("faiss", ignore_errors=True)
+os.mkdir("faiss")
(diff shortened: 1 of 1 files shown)


mdouze:
Hi 
Could you add a section to the INSTALL to say how the setup.py should be used? 
I understand it is to move the files to the right location, but it does not do the compilation by itself.

Stonesjtu:
I‘m working on adding python installation support for CMAKE build. I prefer to place the `setup.py` in `python` directory, and then the setup procedure will be invoked by CMAKE.

mdouze:
Hi @Stonesjtu 
Go ahead and put the setup.py where you prefer.

t-vi:
Thank you!
I think the imports of swigfaiss / swigfaiss_gpu would need to be relative at least in python3 (i.e. in `__init__.py` it should be `from .swigfaiss import *`

Best regards

Thomas

beauby:
LGTM – @mdouze any objections? If not, I'll merge and add an `install` target calling this from Makefile, plus the python3 fix @t-vi mentioned.

mdouze:
The case that I can think of that would break is when you access faiss with a PYTHONPATH, ie.

```
export PYTHONPATH=/my_faiss_install_dir/ 
python -c "import faiss" 
```

Could you check if that still works?

beauby:
@mdouze It will indeed not work (without some added logic) when referencing the source directory, but as you can see in #469, it still works if you set `PYTHONPATH=/my_faiss_src_dir/python/build/lib` (which contains the actual generated python package before install), instead of formerly `PYTHONPATH=/my_faiss_src_dir/python`.

beauby:
Finally merged as part of #469. Thanks @soumith.

PR #316, opened 2018-01-18
use brew to install OpenBLAS rather than port

isrsal:


mdouze:
Hi 
Looks reasonable.

Issue #293, opened 2017-12-21
Plain C interface


Enet4:
Faiss is quite an impressive library, and I was hoping to be able to use it in more development environments other than C++ or Python. I'm particularly interested in using this library in Rust. Writing bindings for Rust is easy when a library provides a plain C API. Unfortunately, SWIG does not support Rust, and as far as I can tell, there is currently no C interface available which Rust could link against.

With that in mind, I'm slowly making a C wrapper of Faiss by hand ([branch](https://github.com/Enet4/faiss/tree/c_api)). I got enough of it covered in order to build and run the example program below. I'm writing an issue here in case more people are interested in this becoming part of the main repository eventually. I'm also open to feedback, since I might be taking an unnecessarily tedious approach. Edit: I found [this](https://github.com/facebookresearch/faiss/issues/228#issuecomment-341038240) in one of the other issues, which is an interesting minimalist approach, but I'm hoping for a _de facto_ wrapper with good error handling and completeness.

```c
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

#include "error_c.h"
#include "Index_c.h"
#include "IndexFlat_c.h"
#include "AutoTune_c.h"

inline double drand() {
    return (double)rand() / (double)RAND_MAX;
}

int main() {
    time_t seed = time(NULL);
    srand(seed);
    printf("Generating some data...\n");
    int d = 128;                           // dimension
    int nb = 100000;                       // database size
    int nq = 10000;                        // nb of queries
    float *xb = malloc(d * nb * sizeof(float));
    float *xq = malloc(d * nq * sizeof(float));

    for(int i = 0; i < nb; i++) {
        for(int j = 0; j < d; j++) xb[d * i + j] = drand();
        xb[d * i] += i / 1000.;
    }
    for(int i = 0; i < nq; i++) {
        for(int j = 0; j < d; j++) xq[d * i + j] = drand();
        xq[d * i] += i / 1000.;
    }

    printf("Building an index...\n");

    FaissIndex* index = NULL;
    int c = faiss_index_factory(&index, d, "Flat", METRIC_L2);  // use factory to create index
    if (c) {
        fprintf(stderr, "%s", faiss_get_last_error());
        exit(-1);
    }
    printf("is_trained = %s\n", faiss_Index_is_trained(index) ? "true" : "false");
    faiss_Index_add(index, nb, xb);                     // add vectors to the index
    printf("ntotal = %ld\n", faiss_Index_ntotal(index));

[...]

PR #317, opened 2018-01-18
Add C API

Enet4:
As previously mentioned in #293, this PR provides a pure C interface to Faiss, which can subsequently be used either in pure C programs or to produce bindings for programming languages with Foreign Function Interface (FFI) support. Although Python doesn't need this, some languages (e.g. Rust and Julia) do not have SWIG support.

This is not a complete API, but provides enough constructs for most tasks, and can be incrementally extended in future contributions.

### Building and Using

The new contents are in the "c_api" folder. In particular, header files in the format "###_c.h" and "gpu/###_c.h" are the pure C APIs that users would include in their project. "###_c.cpp" are their respective wrapper implementations. "macros_impl.h" is a C++ header file containing helper macros.

After building the static library "libfaiss.a", enter the directory and run:

```sh
make
```

This builds a dynamic library `faiss_c", containing the full implementation of Faiss and the necessary wrappers for the C interface. It does not depend on libfaiss.a or the C++ standard library. A C example is included, which can be built as thus:

```sh
make bin/example_c
```

For GPU support, a separate dynamic library in the "c_api/gpu" directory needs to be built and linked.

```sh
make
make bin/example_gpu_c
```

This is a substantial addition, which admittedly may take some time to evaluate. I do have a few concerns which might have to be addressed before merging:

- Although the new code is fairly documented, the building instructions were not included yet. Should there be a new section in INSTALL.md?
- Should the dynamic library with GPU support work standalone? Maybe I should change it so that "libgpufaiss_c.so" can work without "libfaiss_c.so"? This would also make the example easier to run, since it needs to link with both files at the moment.
- There is still no test suite. I could look into this if deemed important.
- It relies on Makefiles rather than CMake. While I understand that there's been a shift towards CMake, this what could be arranged in short-term.
- Not exactly an issue in this PR, but one can identify some hard asserts in the main implementation which would be more useful if they were recoverable at this level. For example, certain GPU index functions could throw an exception if a CUDA function failed instead of aborting. Would you consider such changes in the future?

### DIFF
--- a/example_makefiles/makefile.inc.Linux
+++ b/example_makefiles/makefile.inc.Linux
@@ -9,9 +9,11 @@
 # tested on CentOS 7, Ubuntu 16 and Ubuntu 14, see below to adjust flags to distribution.
 
 
-CC=g++
+CC=gcc
+CXX=g++
 
-CFLAGS=-fPIC -m64 -Wall -g -O3 -mavx -msse4 -mpopcnt -fopenmp -Wno-sign-compare -std=c++11 -fopenmp
+CFLAGS=-fPIC -m64 -Wall -g -O3 -mavx -msse4 -mpopcnt -fopenmp -Wno-sign-compare -fopenmp
+CXXFLAGS=$(CFLAGS) -std=c++11
 LDFLAGS=-g -fPIC  -fopenmp
 
 # common linux flags
@@ -124,7 +126,7 @@ PYTHONCFLAGS=-I/usr/include/python2.7/ -I/usr/lib64/python2.7/site-packages/nump
 
 
 # a C++ compiler that supports c++11
-CC11=g++
+CXX11=g++
 
 # root of the cuda 8 installation
 CUDAROOT=/usr/local/cuda-8.0/
@@ -141,7 +143,7 @@ NVCCFLAGS= $(CUDAFLAGS) \
    -gencode arch=compute_52,code="compute_52" \
    -gencode arch=compute_60,code="compute_60" \
    --std c++11 -lineinfo \
-   -ccbin $(CC11) -DFAISS_USE_FLOAT16
+   -ccbin $(CXX11) -DFAISS_USE_FLOAT16
 
 # BLAS LD flags for nvcc (used to generate an executable)
 # if BLASLDFLAGS contains several flags, each one may
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+/bin/
+/c_api/bin/
+/c_api/gpu/bin/
 /makefile.inc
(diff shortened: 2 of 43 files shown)


example_makefiles/makefile.inc.Linux
```diff
@@ -9,9 +9,11 @@
 # tested on CentOS 7, Ubuntu 16 and Ubuntu 14, see below to adjust flags to distribution.
 
 
-CC=g++
+CC=gcc
+CXX=g++
```
mdouze: Ok to replace CC with CXX.
I will replace CXX11 with CXX since c++11 is also required to compile the CPU lib now.

mdouze:
Hi, 

The diff looks fine, and quite easy to accept because it does not impact the main Faiss. 

The installation instructions will probably be quite long, so please write a separate INSTALL.md and point to it via the toplevel INSTALL.md

There are no automatic tests for Faiss currently, but please write a self-contained binary that can be used to check if compile was ok (and document it in INSTALL.md)

The default way of compiling Faiss is via Makefiles, so it's ok to provide only these.

We are indeed considering making Cuda failures exceptions as well.

Enet4:
I have pushed a few additions and changes. "c_api/INSTALL.md" should contain the necessary details about building and using the API. In addition, "libgpufaiss_c.so" now works without "libfaiss_c.so", and the necessary CUDA libraries are linked as dependencies. This seems to be what the Python API does as well.

mdouze:
Thanks!
I tested the two exampes, that work fine. 
Could you just do a tiny change:
```
--- a/gpu/Makefile
+++ b/gpu/Makefile
@@ -74,7 +74,7 @@ CUOBJ=      impl/BroadcastSum.o \


 .cpp.o:
-       $(CXX11) -fPIC -m64 -Wall -g -Wno-sign-compare -O3 -fopenmp \
+       $(CXX11) $(CXXFLAGS) -fPIC -m64 -Wall -g -Wno-sign-compare -O3 -fopenmp \
          -c $< -o $@ $(EXTRAFLAGS) $(CUDACFLAGS)

 .cu.o:
```
Then I'll merge the PR.

Enet4:
The suggested change was made. I also found a few issues that I had overlooked, and are indeed necessary for things to work properly.

Enet4:
@mdouze I have rebased this branch to resolve all conflicts, so it's ready to merge again. Although I did not make direct tests of the API, I managed to obtain some coverage through the tests in the Rust bindings ([Travis build](https://travis-ci.org/Enet4/faiss-rs)). Let me know if you'd like the commits squashed.

beauby:
LGTM, merging. Thanks for your work @Enet4.

PR #326, opened 2018-02-02
Update makefile.inc.Linux

jwz-ecust:


mdouze:
Hi 
This change is only for anaconda.

PR #342, opened 2018-02-19
[WIP] Update tutorial

beauby:
- [x] Remove transitional print import.
- [x] Add C++ code example for one GPU.
- [x] Add python/C++ code examples for several GPUs.

### DIFF
--- a/tutorial/cpp/4-GPU.cpp
+++ b/tutorial/cpp/4-GPU.cpp
@@ -10,6 +10,8 @@
 #include <cstdlib>
 
 #include <faiss/IndexFlat.h>
+#include <faiss/gpu/GpuIndexFlat.h>
+#include <faiss/gpu/StandardGpuResources.h>
 
 
 int main() {
@@ -32,9 +34,8 @@ int main() {
         xq[d * i] += i / 1000.;
     }
 
-    faiss::gpu::GpuResources *res = new faiss::gpu::StandardGpuResources;
-
-    faiss::GpuIndexFlatL2 index(res, d);   // call constructor
+    faiss::gpu::StandardGpuResources res;
+    faiss::gpu::GpuIndexFlatL2 index(&res, d);   // call constructor
 
     printf("is_trained = %s\n", index.is_trained ? "true" : "false");
     index.add(nb, xb);                     // add vectors to the index
--- a/tutorial/cpp/4-GPU.cpp
+++ b/tutorial/cpp/4-GPU.cpp
@@ -1,0 +1,102 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD+Patents license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <cstdio>
+#include <cstdlib>
+
+#include <faiss/IndexFlat.h>
+
+
+int main() {
+    int d = 64;                            // dimension
+    int nb = 100000;                       // database size
+    int nq = 10000;                        // nb of queries
+
+    float *xb = new float[d * nb];
+    float *xq = new float[d * nq];
+
+    for(int i = 0; i < nb; i++) {
+        for(int j = 0; j < d; j++)
+            xb[d * i + j] = drand48();
+        xb[d * i] += i / 1000.;
+    }
+
+    for(int i = 0; i < nq; i++) {
+        for(int j = 0; j < d; j++)
+            xq[d * i + j] = drand48();
+        xq[d * i] += i / 1000.;
+    }
+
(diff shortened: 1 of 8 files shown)


mdouze:
Have you tested with py2 + py3?

beauby:
@mdouze Yes, python2.7 and python3.6.

tutorial/cpp/4-GPU.cpp
```diff
@@ -0,0 +1,102 @@
...
+    faiss::GpuIndexFlatL2 index(res, d);   // call constructor
+
+    printf("is_trained = %s\n", index.is_trained ? "true" : "false");
+    index.add(nb, xb);                     // add vectors to the index
+    printf("ntotal = %ld\n", index.ntotal);
+
+    int k = 4;
+
+    {       // sanity check: search 5 first vectors of xb
+        long *I = new long[k * 5];
```
Enet4: When relying on C++11, one can wrap these data allocations around a `unique_ptr`. Not only it would be automatically freed when falling out of scope, it would protect also prevent leaks in case of exceptions.

```c++
#include <memory>

// ...

auto I = std::unique_ptr<long[]>(new long[k * 5]);
auto D = std::unique_ptr<float[]>(new float[k * 5]);

index.search(5, xb, k, D.get(), I.get());
```

tutorial/cpp/5-Multiple-GPUs.cpp
```diff
@@ -0,0 +1,118 @@
...
+    int ngpus = faiss::gpu::getNumDevices();
+
+    std::vector<faiss::gpu::GpuResources*> res;
+    std::vector<int> devs;
+    for(int i = 0; i < ngpus; i++) {
+        res.push_back(new faiss::gpu::StandardGpuResources);
+        devs.push_back(i);
+    }
+
+    faiss::Index *index =
```
Enet4: Was this in a more complicated context, wouldn't want the index to leak. I believe a `unique_ptr` around the pointer to the GPU index works fine here.

```c++
auto index = std::unique_ptr<faiss::Index>(
    faiss::gpu::index_cpu_to_gpu_multiple(
        res,
        devs,
        new faiss::IndexFlat(d)
    ));
```

beauby:
@Enet4 Thanks for your comments. Agreed that providing examples with good practices is a plus. I modified the example on one GPU to avoid using `new`, as it was totally unnecessary.

Regarding the (lack of) use of `unique_ptr`s, the rationale is to avoid scaring off people who are not familiar with it, and focus on the library usage.

tutorial/cpp/5-Multiple-GPUs.cpp
```diff
@@ -0,0 +1,122 @@
...
+        faiss::gpu::index_cpu_to_gpu_multiple(  // call constructor
+            res,
+            devs,
+            new faiss::IndexFlatL2(d)
```
mdouze: This will not be deallocated, so it should be declared separately.

tutorial/cpp/5-Multiple-GPUs.cpp
```diff
@@ -0,0 +1,122 @@
...
+        );
+
+    printf("is_trained = %s\n", index->is_trained ? "true" : "false");
+    index->add(nb, xb);                     // add vectors to the index
+    printf("ntotal = %ld\n", index->ntotal);
+
+    int k = 4;
+
+    {       // sanity check: search 5 first vectors of xb
```
mdouze: Remove the sanity check, we will not comment it twice

tutorial/cpp/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+-include $(MAKEFILE_INC)
+
+all: cpu gpu
+
+cpu: tutorial1 tutorial2 tutorial3
+
+gpu: tutorial4 tutorial5
+
+tutorial1: 1-Flat.cpp ../../libfaiss.a
```
mdouze: could you call the executables 1-Flat, 2-IVFFlat, etc. ?

tutorial/cpp/5-Multiple-GPUs.cpp
```diff
@@ -0,0 +1,98 @@
...
+            xq[d * i + j] = drand48();
+        xq[d * i] += i / 1000.;
+    }
+
+    int ngpus = faiss::gpu::getNumDevices();
+
+    std::vector<faiss::gpu::GpuResources*> res;
+    std::vector<int> devs;
+    for(int i = 0; i < ngpus; i++) {
+        res.push_back(new faiss::gpu::StandardGpuResources);
```
wickedfoo: Use `std::unique_ptr` at the least if you keep it by pointer, so you don't need the free at the end.

PR #343, opened 2018-02-21
Make tutorial examples use dynamic linkage.

beauby:


gpu/Makefile
```diff
@@ -85,6 +85,9 @@ CUOBJ=      impl/BroadcastSum.o \
 $(LIBNAME).a: $(CPPOBJ) $(CUOBJ)
 	ar r $@ $^
 
+$(LIBNAME).$(SHAREDEXT): $(CPPOBJ) $(CUOBJ)
```
Enet4: Just a quick heads up, the `clean` target in 119 should probably consider this file as well.

gpu/Makefile
```diff
@@ -85,6 +85,9 @@ CUOBJ=      impl/BroadcastSum.o \
...
+	$(CC) $(LDFLAGS) $(FAISSSHAREDFLAGS) -o $(LIBNAME).$(SHAREDEXT) $^ $(CUDAFLAGS)
```
Enet4: Ok, just found a slightly more serious concern: I was trying to build the GPU examples, but it was constantly failing to link:

```
/opt/cuda//bin/nvcc  -I /opt/cuda//targets/x86_64-linux/include/ -Xcompiler -fPIC -Xcudafe --diag_suppress=unrecognized_attribute -gencode arch=compute_37,code="compute_37" -gencode arch=compute_61,code="compute_61" --std c++11 -lineinfo -ccbin g++-6 -DFAISS_USE_FLOAT16 -o 4-GPU 4-GPU.cpp -Xcompiler \"-Wl,-rpath=../../:../../gpu/\" -L../.. -L../../gpu -lfaiss -lgpufaiss \
-I../../../ -Xcompiler -fopenmp -lcublas -Xlinker /usr/lib64/libopenblas.so.0 \

ld: 4-GPU: hidden symbol `cudaMemcpyAsync' in /opt/cuda//bin/..//lib64/libcudart_static.a(libcudart_static.a.o) is referenced by DSO
ld: final link failed: Bad value
make: *** [Makefile:28: 4-GPU] Error 1
```

After some research, I found that it can be related to an attempt of a dynamic library to use a statically defined reference from an object that it was not properly linked against, or something along these lines. What solves the problem here is to make `libgpufaiss.so` dynamically link to `cudart` and `cublas` explicitly:

```makefile
    $(CC) -L$(CUDA_ROOT)/lib64 -lcublas -lcudart $(LDFLAGS) $(FAISSSHAREDFLAGS) -o $(LIBNAME).$(SHAREDEXT) $^ $(CUDAFLAGS)
```

This also makes the `-lcublas` flags in the examples redundant, and can be removed.

beauby: Thanks for the writeup @Enet4. I pushed the suggested fixes – could you give it a try?

Enet4: @beauby Yes, it works now. :+1:

PR #344, opened 2018-02-22
Add (un)install Makefile targets for CPU version.

beauby:
+ Add `installdirs` target ;
+ add `install` target ;
+ add `uninstall` target.

All targets support customization via `DESTDIR` and `PREFIX`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ MAKEFILE_INC=makefile.inc
 
 PREFIX ?= /usr/local/
 
-all: .env_ok libfaiss.a tests/demo_ivfpq_indexing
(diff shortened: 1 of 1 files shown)


beauby:
@mdouze `PREFIX` is dependant on the OS/package manager (usually `/usr/local` on linux, `/opt/local` on Mac OS X with macports, etc.), and `DESTDIR` is a GNU convention (c.f. https://www.gnu.org/prep/standards/html_node/DESTDIR.html), used for staged installations (as macports for instance does: first install everything in a temp directory (but keeping the directory structure), and then move it to its final destination).

PR #348, opened 2018-02-23
Make tests compatible with py3

ailzhang:
I'm building conda packages for faiss and encountered test failure with py3. 
Making tests compatible with py3.


mdouze:
That works, thanks!

PR #349, opened 2018-02-23
updating dockerfile after PR #347 (sync w/ FB) 

nathandouglas:
* Updated the build steps to reflect what is in the current INSTALL.md after #347.
* Updated PYTHONPATH to include /opt/faiss directory where `_swigfaiss.so` and `_swigfiass_gpu.so` are located.


mdouze:
Hi 
thanks for the Fix!

PR #357, opened 2018-03-03
Fixed missing malloc.h include error on osx

ifeherva:
Added precompiler macros around malloc.h since it does not exist on osx and causes a build failure. Tested on OSX 10.12.6 (Sierra)

Another solution would be to change
#include <malloc.h>
to
#include "malloc.h"

Not sure which one is preferred.

### DIFF
--- a/IndexScalarQuantizer.cpp
+++ b/IndexScalarQuantizer.cpp
@@ -11,7 +11,11 @@
 #include <cstdio>
 #include <algorithm>
 
+#ifdef __APPLE__
+#include <stdlib.h>
+#else
 #include <malloc.h>
+#endif
 
 #include <omp.h>
 


mdouze:
hi 
I have never seen an error with malloc.h missing

ifeherva:
Hmm, not sure on what system you are testing it, on osx it is either #include <malloc/malloc.h> or you use the stdlib.h (homebrew llvm)

Enet4:
I found a relevant SO question: https://stackoverflow.com/q/12973311/1233251

Including `<stdlib.h>` (or `<cstdlib>` in C++ for the namespaced counterpart) is the standard way to obtain the `malloc` function. While `malloc.h` might still work on Linux, I wouldn't be surprised if it didn't work on other compilers. I would actually recommend that all inclusions of `malloc.h` are replaced, regardless of the target platform.

mdouze:
Sorry, you are right actually. I removed the malloc.h include from the includes.

PR #358, opened 2018-03-05
Adding working setup.py version that includes compilation of Faiss

amrotork:
Hi,
I have created a version of setup.py that enables installation of faiss on a Conda environment with MKL. I'm planning to support more compilation options like openBLAS and others.

Regards,
Amro


mdouze:
Hi 
Thanks for the work. It looks interesting. 
Could you please diff against the latest Faiss version? It seems that some differences are already included in the diff.

amrotork:
Hi Matthijs,
Yes, they are. But for some reason, github keep telling me that those are different than the one in the master branch on facebookresearch repo.

Regards,
Amro

felipegasparini:
Hi! 
Is there any ETA to have this PR merged in master?
I'm interested in this, so I can help if there is something still missing.

mdouze:
Hi 
Thanks for the PR. However, the supported way of building with conda is via the pytorch channel, see https://github.com/facebookresearch/faiss/blob/master/INSTALL.md#install-via-conda

amrotork:
I think I could customize it to auto detect the system that it's into and update the compilation paths accordingly. This way it could support all OS systems, not just anaconda.

PR #363, opened 2018-03-13
update Readme to include conda install info

ailzhang:
FAISS conda packages has been uploaded to pytorch anaconda channel here. https://anaconda.org/pytorch/repo.
Updating the README to include this installation method.


mdouze:
Thanks for the conda package! It works for my environment.
Please move the conda instructions to INSTALL.md instead of README.md, and say in the intro of INSTALL.md that this is the easiest way of installing Faiss for Python.

PR #380, opened 2018-03-28
Fix compile error in cpp tutorial

myl2821:
Hello,
It seems that the cpp tutorial is kind of outdated, so that I make a small patch to fix it.


mdouze:
Hi!
Thanks for spotting this and fixing it!

Issue #371, opened 2018-03-20
GPU issue when installing from conda


hminle:
# Summary

I install Faiss from conda (GPU version)
![image](https://user-images.githubusercontent.com/16201681/37644231-a7c399d0-2c55-11e8-99c7-396b59dcb5b7.png)

And I got ```ImportError: No module named 'swigfaiss'```
Could you guys help me out?
Did I forget anything?

# Platform

OS: Ubuntu

Faiss version: 

Faiss compilation options: 

Running on :
- [ ] CPU
- [x] GPU

# Reproduction instructions

![image](https://user-images.githubusercontent.com/16201681/37644254-bcc2b9ce-2c55-11e8-907a-8cb6e8467f8b.png)

hminle:
I install faiss:
faiss-gpu: 1.2.1-py35_cuda9.0.176_1 pytorch

my cuda version
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2017 NVIDIA Corporation
Built on Fri_Nov__3_21:07:56_CDT_2017
Cuda compilation tools, release 9.1, V9.1.85
```

hminle:
I managed to overcome swigfaiss issue, but got this one 
![image](https://user-images.githubusercontent.com/16201681/37654993-93bb48a4-2c76-11e8-8007-c649641c3a55.png)

It does not recognize my GPU

ailzhang:
@hminle , could you list all the items under $YOUR_CONDA_HOME/lib/python3.6/site-packages/faiss?

hminle:
Here is all the items
![image](https://user-images.githubusercontent.com/16201681/37659504-c2c31f6c-2c82-11e8-8710-b3ca370e430e.png)

ailzhang:
Inside that folder, could you try 'python; import faiss' and let me know what happens?

hminle:
![image](https://user-images.githubusercontent.com/16201681/37659665-1fc5744e-2c83-11e8-832a-3024cdb1cda1.png)
it's show the same issue

ailzhang:
That looks weird, since .so file it tries to import is there. 
Could you try rm -rf $YOUR_CONDA_HOME/lib/python3.6/site-packages/faiss*, and then reinstall the faiss-gpu package again?

hminle:
nothing happens, it's the same issue, fall back to CPU version. 
[...]

PR #392, opened 2018-04-04
Mention CUDA version required to run faiss-gpu

ashishmadeti:
As discussed in #371

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,7 +15,7 @@ You can easily install it by
 ```
 # CPU version only
 conda install faiss-cpu -c pytorch
-# GPU version requires CUDA to be installed, otherwise it falls back to CPU version
+# GPU version requires CUDA 9.0 to be installed, otherwise it falls back to CPU version
 conda install faiss-gpu -c pytorch
 ```
 


wickedfoo:
Faiss GPU only requires CUDA 7.5 or greater. The problem you seem to be having is specific to 9.1?

Enet4:
@wickedfoo I believe the issue here is that the conda package is pre-compiled specifically for CUDA 9.0, so it will only work for this version.

ailzhang:
Hi, actually we've added more support including CUDA8.0/CUDA9.0/CUDA9.1 and updating the INSTALL.md here. #395

ashishmadeti:
@wickedfoo I am using CUDA 9.0 with the conda package

@ailzhang #395 makes more sense. But by default, conda installs the CUDA 9.0 version.

mdouze:
addressed by #392

ailzhang:
@ashishmadeti  I tried on several computer which has CUDA9.0 installed, if I don't have cuda90 feature in my conda list, the default package is cuda8.0 one.

PR #407, opened 2018-04-12
Implement remove_ids in IndexPQ

myl2821:
Hi,

I found remove_ids was not implemented in IndexPQ when playing with faiss and made a patch of it like what was done in IndexFlat, hope you can take a review.


mdouze:
Hi 
This looks good to me.

PR #425, opened 2018-04-27
[C API] Additional API coverage: IndexLSH, I/O, meta-indexes

Enet4:
This PR extends the C API with some some functions and types for accessing features in the library which may be useful in bindings. It is composed of 3 major contributions (feel free to let me know if you would like separate pull requests instead):

- Bridged `IndexLSH` to the C interface, which to my understanding isn't accessible through the index factory.
- Added most of "index_io.h" to the C header file "index_io_c.h", so as to enable saving and loading indices to/from disk. To serve as an example, "example_c.c" was extended to save the index in the end.
- Added most of "MetaIndexes.h" to the C counterpart "MetaIndexes_c.h", thus allowing users to create index ID maps and index shards.

Although this is not yet fully tested, generating the low-level Rust bindings containing these changes was successful and did not interfere with the rest of the library ([Travis build](https://travis-ci.org/Enet4/faiss-rs/builds/372026868)).

### DIFF
--- a/c_api/IndexFlat_c.cpp
+++ b/c_api/IndexFlat_c.cpp
@@ -24,6 +24,9 @@ using faiss::IndexFlatL2BaseShift;
 using faiss::IndexRefineFlat;
 using faiss::IndexFlat1D;
 
+DEFINE_DESTRUCTOR(IndexFlat)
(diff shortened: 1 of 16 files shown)


mdouze:
This looks like a legitimate extension of the C api. Thanks!

PR #427, opened 2018-04-30
read/write index with std::function wrapper

dengoswei:
1. add std::function wrapper to read/write index; 
   In our case, we may read/write index from/to hdfs. 

2. add IndexIVF::access;

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,9 +3,9 @@ cmake_minimum_required(VERSION 2.8.7)
 # faiss project
 project(faiss C CXX)
 
-option(BUILD_TUTORIAL "Build tutorials" ON)
-option(BUILD_TEST "Build tests" ON)
-option(BUILD_WITH_GPU "Build faiss with gpu (cuda) support" ON)
+option(BUILD_TUTORIAL "Build tutorials" OFF)
+option(BUILD_TEST "Build tests" OFF)
+option(BUILD_WITH_GPU "Build faiss with gpu (cuda) support" OFF)
 option(WITH_MKL "Build with MKL if ON (OpenBLAS if OFF)" OFF)
 
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules)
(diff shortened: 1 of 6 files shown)


mdouze:
Hi 

Thanks for the PR. The modification to the I/O functions to make them generic is something we have on our todo list for a long time, so it is useful. So I will consider to accept it if: 

- the access function is removed (it is unrelated to the I/O part)

- no `std::function`. There should be a virtual class with virtual functions `write` and `read`, that can be implemented as FILE* objects or otherwise. 

Thanks

dengoswei:
why not std::function ? Virtual class with virtual functions `write` and `read` can be wrapped with std::function, and write_index won't need any read interface.

mdouze:
Please follow the Faiss conventions.

dengoswei:
well. 
https://github.com/facebookresearch/faiss/compare/master...dengoswei:master

mdouze:
Thanks for the update. It looks almost good for acceptance. 

Could you: 

- remove the PFN_ prefix from the macros and delete the old versions of the macros

- there are some functions that require a file id for memory mapping. Could you add a get_fileno() method to the IOWriter/IOReader objects and that FAISS_THROW_MSG()  by default but returns a proper file descriptor for the FileIOXXX versions

- remove unrelated changes in tests/test_pairs_decoding.cpp

Thanks a lot this will be a useful diff!

dengoswei:
- remove the PFN_ prefix from the macros and delete the old versions of the macros
   DONE
- remove unrelated changes in tests/test_pairs_decoding.cpp
   DONE
- there are some functions that require a file id for memory mapping
  check out read_InvertedLists;

mdouze:
Thanks, you can also rename `pfn_read` and `pfn_write` with `f`. The diff will be easier to read.

dengoswei:
up;

mdouze:
Thanks. We will pass the diff through our internal tests next week.

mdouze:
Sorry for the delay. The tests pass, so I am accepting the PR.

PR #443, opened 2018-05-11
- Add install target for the faiss library and headers to the

mohnishkodnani:
CMakeLists.txt, so we can install the libs and headers with make
install.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,3 +55,8 @@ endif(BUILD_TUTORIAL)
 if(BUILD_TEST)
     add_subdirectory(tests)
 endif(BUILD_TEST)
+# Install libraries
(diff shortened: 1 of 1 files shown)


mdouze:
@hlzz, any comment?

mohnishkodnani:
I am not putting the gpu headers because I currently dont have an environment where I can compile the gpu version of faiss and also install those. 
But that can come as another pull request i guess, if this one makes sense.

mohnishkodnani:
@hlzz install target is what i have added.

mohnishkodnani:
Let me know what I can do now ?

mdouze:
Thanks @mohnishkodnani and @hlzz, then I accept the PR.

PR #448, opened 2018-05-14
[C API] Fix libgpufaiss_c with extended API

Enet4:
I just noticed a tiny issue in the previous extension to the C API, which this PR corrects.
The new object files from #425 were not being included in the GPU shared object because `c_api/gpu/Makefile` was keeping its own replicated list of object files.
I adjusted the libgpufaiss_c.so building process so that it fetches the complete archive file built from c_api/Makefile, thus potentially avoiding these mistakes in the future.

- Change `libfaiss_c.a` and `libgpufaiss_c.a` targets to only include the C interface's wrapper implementations
- Include `libfaiss_c.a` target on `c_api`'s `make all`
- Build `libgpufaiss_c.so` with `libfaiss_c.a` instead of the "*_c.o" object files in the parent directory


mdouze:
Hi 
Accepting because it touches only the C interface.

PR #453, opened 2018-05-17
Support for index.assign in Python

Enet4:
I noticed that the `assign` method in the Python interface was not adapted to a high level API, like the other operations (`add`, `search`, `reconstruct`, ...). Steps to reproduce the issue:

```python
import numpy as np
import faiss
index = faiss.index_factory(2, "Flat")
index.add(np.array([[1, 2], [3, 4]], dtype=np.float32))
# can't call `assign` like this because it's still a low-level SWIG function
labels = index.assign(np.array([[0, 0]], dtype=np.float32), 1)
```

```txt
TypeError: assign() missing 1 required positional argument: 'labels'
```

With these changes, the code above now works as expected. One could indeed use `search` instead, but I did not need to fetch the distances in my use case.

### DIFF
--- a/faiss.py
+++ b/faiss.py
@@ -105,6 +105,13 @@ def handle_Index(the_class):
         assert ids.shape == (n, ), 'not same nb of vectors as ids'
         self.add_with_ids_c(n, swig_ptr(x), swig_ptr(ids))
 
+    def replacement_assign(self, x, k):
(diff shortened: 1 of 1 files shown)


beauby:
LGTM. Any remarks @mdouze?

mdouze:
ok

PR #457, opened 2018-05-22
Add TravisCI config.

beauby:
Builds on:
+ Linux with g++ libblas/libblapack
+ Linux with g++ libatlas
+ Linux with g++ libopenblas
+ Linux with clang++ libopenblas
+ OSX with g++ libopenblas
+ OSX with clang++ libopenblas


mdouze:
Thanks for the PR 

Why did you change the comments in makefile.inc.Linux? They could be useful for some people.

beauby:
@mdouze I put back the comments about the required packages. The intention was to remove the static linkage bits.

mdouze:
Do you plan to add the python tests before committing?

beauby:
Rebased on top of #466.

Enet4:
Just gave this a try on my machine. Is compiling the GPU version supposed to work at the moment? It would be nice if `./configure` would also take care of identifying the right CUDA flags.

I also noticed that some changes are required in the C API after this PR, but I can send a fix once it has been merged.

PR #466, opened 2018-05-31
Refactor makefiles and add configure script

beauby:
#### TL;DR: `./configure && make && make install`
+ add configure script to generate `makefile.inc`
+ separate python wrapper from main source
+ cleanup tests
+ add `make test`/`make -C python test` targets
+ refactor makefiles (remove redundant flags, etc.)


mdouze:
My understanding is that it will not work anymore without installing faiss, because swigfaiss is created in python/ and faiss.py is in the subdirectory. This is inconvenient when adding new stuff to the C++ part. Is that correct?

Is there a rule to install the dev version of Faiss? With the include files in 

```
some_prefix/include/faiss/Index.h
some_prefix/include/faiss/IndexFlat.h
....
```

beauby:
> My understanding is that it will not work anymore without installing faiss, because swigfaiss is created in python/ and faiss.py is in the subdirectory. This is inconvenient when adding new stuff to the C++ part. Is that correct?

Everything python-related now resides in the `python/` subdirectory, but the top level Makefile still has a `make py` target.

> Is there a rule to install the dev version of Faiss? With the include files in [...]

Yes, the `install` target adds the headers in `/usr/local/include/faiss/` by default.

beauby:
@Enet4 I added support for cuda, could you give it a shot? By defaults it looks for it in `/usr/local/cuda`, but you can specify the location via `./configure --with-cuda=/usr/local/cuda-8.0`.

@mdouze I think this is ready for merging (after squashing the commits). I tested on osx/linux with openblas/MKL, and on gpu. Last thing to automate would be determining the C type of fortran integers, but this can be done at a later time.

PR #469, opened 2018-06-02
Python install

beauby:
Based on #261.

+ Add `build` and `install` targets to python `Makefile`.
+ Fix faiss.py to work with python3.


Makefile
```diff
@@ -64,7 +64,8 @@ depend:
 
 test: libfaiss.a py
 	make -C tests run
-	PYTHONPATH=./python $(PYTHON) -m unittest discover tests/ -v
+	PYTHONPATH=./python/build/`ls python/build | grep lib` \
```
mdouze: why is this grep necessary? What is the exact filename?

beauby: The exact file name is platform-dependent (e.g. `lib.linux-x86`). The clean way would be to have `setup.py` run the tests, but that would imply having the tests within the `python/` subdir.

PR #477, opened 2018-06-04
Make c_api compatible with configure'd makefile

Enet4:
Here are the necessary adjustments to the building process of the C API after #466. It is mostly a change of environment variable flags. I also removed the use of `inline` in the respective example files, which was not needed and could bring compilation issues.


beauby:
Thanks for following up on this :+1:

PR #489, opened 2018-06-13
Fix linking order for misc/test_blas in Makefile.

beauby:
Fixes #487.

Issue #487, opened 2018-06-12
make misc/test_blas failed in ubuntu14.04


shutcode:
# Summary
after running configure, failed to make misc/test_blas
I'm sure openblas is installed(apt-get and install from source),and the link flags are correctly set. error info as follows
```
test_blas.cpp:(.text+0x187)：undefined reference to `sgemm_'
test_blas.cpp:(.text+0x346)：undefined reference to `sgeqrf_'
collect2: error: ld returned 1 exit status
```
I have tried solutions from #15  and #131 

# Platform

OS: ubuntu 14.04
Faiss version: 1fe2872013685092d697f08a2a48e110acd25b2b

Faiss compilation options: 
```
g++ -std=c++11  -fopenmp  -lopenblas -llapack  -o misc/test_blas misc/test_blas.cpp
#or 
g++ -std=c++11  -fopenmp  -lopenblas -llapack -L/opt/OpenBLAS/lib -I/opt/OpenBLAS/include -o misc/test_blas misc/test_blas.cpp
````

Running on :
- [x] CPU

# Reproduction instructions

```
./configure
make misc/test_blas.cpp
```
 
```
test_blas.cpp:(.text+0x187)：undefined reference to `sgemm_'
test_blas.cpp:(.text+0x346)：undefined reference to `sgeqrf_'
collect2: error: ld returned 1 exit status
ldconfig -p |grep openblas
	libopenblas.so.0 (libc6,x86-64) => /usr/local/lib/libopenblas.so.0
	libopenblas.so.0 (libc6,x86-64) => /usr/lib/libopenblas.so.0
	libopenblas.so (libc6,x86-64) => /usr/local/lib/libopenblas.so
	libopenblas.so (libc6,x86-64) => /usr/lib/libopenblas.so
 ldconfig -p |grep lapack
	liblapack.so.3 (libc6,x86-64) => /usr/lib/liblapack.so.3
	liblapack.so (libc6,x86-64) => /usr/lib/liblapack.so
```

**Comment by beauby** [comment]
1. What does `g++ --version` say?
2. What happens if you put `-lopenblas -llapack` at the end of the line? i.e. g++ -std=c++11  -fopenmp  -o misc/test_blas misc/test_blas.cpp -lopenblas -llapack.

**Comment by shutcode** [comment]
1. g++ version
```
Target: x86_64-linux-gnu
[...]

PR #490, opened 2018-06-13
[C API] Rearrange lib linking order

Enet4:
Turns out that the Makefiles in the C interface also had the library linking order issue as the one in #487, and solved in #489. It seemed to work locally, but it failed when trying to get Travis to compile the Rust bindings.


beauby:
LGTM. Thanks for fixing this @Enet4.

PR #503, opened 2018-06-26
Use madvice instead of iteration over data

notimesea:


notimesea:
Using OnDiskInvLists prefetching consume a lot of CPU while doing nothing. I suggest using [madvice](http://man7.org/linux/man-pages/man2/madvise.2.html) instead of iteration over data since O/S doesn't guarantee that pages would be in memory in any cases except [mlock](http://man7.org/linux/man-pages/man2/mlock.2.html)

mdouze:
Hi 
This seems to be exactly the system call that we need. 
Would it be possible to change the PR so that the behavior of prefetch() can be set with a boolean `use_madvise`: 

- true: don't spawn threads, just call madvise on the lists (because I assume it returns immediately)

- false: current behaviour

Then we could benchmark. Do you have a setting that stresses the OnDisk enough for this? Otherwise I can make one.

notimesea:
Ok, I will change the API to support this. I haven't such sort of benchmarks, only for my purposes when almost whole index could fit into the memory.

notimesea:
I have added option to use madvice. @mdouze can you make benchmarks?

notimesea:
@mdouze any updates?

mdouze:
Hmmm... I will not have time to do benchmarks for the time being. 
Since the code will not break anything, I'll accept it right away.

PR #504, opened 2018-06-28
Facebook sync

beauby:
Changelog:
+ Support for binary indexes (`IndexBinaryFlat`, `IndexBinaryIVF`)
+ Support fp16 encoding in scalar quantizer 
+ Support for deduplication in `IndexIVFFlat` (Duplicate vectors are identified and removed at add time and added back at search time. factory: "FlatDedup")
+ Support for index serialization
+ Add a method to add preassigned embeddings 

Bugs:
+ Fix MMAP bug for normal indexes 
+ Fix propagation of io_flags in read func 
+ CUDA 9 Faiss k-selection fixes 

TODO:
- doxygen

### DIFF
--- a/python/swigfaiss.swig
+++ b/python/swigfaiss.swig
@@ -83,6 +83,9 @@ extern "C" {
 #include "MetaIndexes.h"
 #include "FaissAssert.h"
 
+#include "IndexBinaryFlat.h"
+#include "IndexBinaryIVF.h"
+
 #include "index_io.h"
 
 #include "utils.h"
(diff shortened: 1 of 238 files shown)


python/swigfaiss.swig
```diff
@@ -118,13 +121,16 @@ namespace std {
         size_t size();
         T at (size_t n) const;
         void resize (size_t n);
+        void swap (vector<T> & other);
```
mdouze: could this be the source of the CI error?

tests/test_binary_io.py
```diff
@@ -0,0 +1,86 @@
...
+
+        index = faiss.IndexBinaryFlat(d)
+        index.add(self.xb)
+        D, I = index.search(self.xq, 3)
+
+        tmpnam = os.tempnam()
+        faiss.write_index_binary(index, tmpnam)
+
+        index2 = faiss.read_index_binary(tmpnam)
+
```
mdouze: add try: finally: block to os.remove(tmpnam)

python/swigfaiss.swig
beauby: Yes. C.f. https://github.com/swig/swig/issues/568

PR #511, opened 2018-07-06
Add conda packages metadata.

beauby:


mdouze:
What is the status of this? Are we supporting conda ourselves?

beauby:
@mdouze If we want to own the conda build/release process, then we need this.

beauby:
Closing in favor of #769.

Issue #491, opened 2018-06-14
[docker deploy] Dockerfile compiling problem


alues:
# Summary

# Platform

OS: *`docker`*-> `nvidia/cuda:8.0-devel-ubuntu16.04`

Faiss version: [faiss:master](https://github.com/facebookresearch/faiss/commit/dd6c9ebf2da3123e055ffdc99f9b597e8b97de04)

Faiss compilation options: 
```bash
./configure && \
    make -j $(nproc) && \
    make -j $(nproc) -C gpu && \
    make -j $(nproc) -C python gpu && \
    make -j $(nproc) -C python build && \
    make -j $(nproc) -C python install
```
Running on :
- [x] CPU
- [x] GPU

# Reproduction instructions
`python -c "import _swigfaiss_gpu"` in faiss's python folder
```bash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: ./_swigfaiss_gpu.so: undefined symbol: omp_set_lock
```

how can i fix it?

beauby:
Could you post the output of `./configure`?

alues:
```bash
checking for g++... g++
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking whether g++ supports C++11 features with -std=c++11... yes
checking for gcc... gcc
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking how to run the C preprocessor... gcc -E
checking whether make sets $(MAKE)... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for python... python
checking for python-config... python-config
checking for Python C flags... -I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7
checking for Python ld flags... -L/usr/lib/python2.7/config-x86_64-linux-gnu -L/usr/lib -lpython2.7 -lpthread -ldl  -lutil -lm  -Xlinker -export-dynamic -Wl,-O1 -Wl,-Bsymbolic-functions
checking for numpy headers path... /usr/lib/python2.7/dist-packages/numpy/core/include
checking for swig... swig
checking how to run the C++ preprocessor... g++ -std=c++11 -E
[...]

PR #526, opened 2018-07-17
Correct docker build

pletessier:
Hi,

Here is a correction solving the broken docker build. The solution is mainly based on the work of @beauby : https://gist.github.com/beauby/813791b8d511d7801c85f4cf73bc6307

@mdouze, it could solve issues #470 #491.

Pierre

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM nvidia/cuda:8.0-devel-ubuntu16.04
 MAINTAINER Pierre Letessier <pletessier@ina.fr>
 
-RUN apt-get update -y
-RUN apt-get install -y libopenblas-dev python-numpy python-dev swig git python-pip wget
+RUN apt-get update -y && \
+    apt-get install -y libopenblas-dev python-numpy python-dev swig python-pip curl
 
 RUN pip install matplotlib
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,37 +1,29 @@
...
+    apt-get install -y libopenblas-dev python-numpy python-dev swig python-pip curl
 
 RUN pip install matplotlib
 
 COPY . /opt/faiss
 
 WORKDIR /opt/faiss
 
-ENV BLASLDFLAGS /usr/lib/libopenblas.so.0
+ENV BLASLDFLAGS=/usr/lib/libopenblas.so.0 \
```
beauby: This is a no-op.

Dockerfile
```diff
@@ -1,37 +1,29 @@
...
+    PYTHONPATH=$PYTHONPATH:/opt/faiss \
```
beauby: This shouldn't be needed since `make install` is called.

Dockerfile
```diff
@@ -1,37 +1,29 @@
...
+    PYTHON=python
```
beauby: This should not be needed.

Dockerfile
```diff
@@ -1,37 +1,27 @@
...
 RUN pip install matplotlib
 
 COPY . /opt/faiss
 
 WORKDIR /opt/faiss
 
-ENV BLASLDFLAGS /usr/lib/libopenblas.so.0
+ENV BLASLDFLAGS=/usr/lib/libopenblas.so.0
 
 RUN mv example_makefiles/makefile.inc.Linux ./makefile.inc
```
beauby: This is not needed anymore, since `./configure` will generate `makefile.inc`.

mdouze:
@beauby, what should we do with this PR?

beauby:
Thanks for your work Pierre. Closing this as a new Dockerfile is available in #769.

Issue #170, opened 2017-08-01
Add to pypi


kahnvex:
Are there any plans to add this project to [pypi](https://pypi.python.org/pypi) so others may install it and use it via python with ease?

Willing to help with this.

mdouze:
Hi 

There are no plans that I know of. The process to add a new install method is that a contributor submits a PR and commits to support it when Faiss is updated, then we integrate it.

kyamagu:
Hi, I have set up CPU-only wheel package at PyPI:

https://pypi.org/project/faiss-cpu/

It seems other unofficial packages also exist:

https://pypi.org/project/faiss/
https://pypi.org/project/faiss-centos/

My packaging scripts are [here](https://github.com/kyamagu/faiss-wheels).
Let me know if there is any plan to officially support pypi distribution.

kyamagu:
I've also made [a faiss-gpu wheel package available at PyPI](https://pypi.org/project/faiss-gpu/):

```
pip install faiss-gpu
```

For anyone wishing to tune the performance, there is a build instruction for source distribution.
https://github.com/kyamagu/faiss-wheels

DrChrisLevy:
If we could get an official PYPI package (as an alternative to conda) that would be great. Many developers have existing stacks with docker/pipenv/pip so being able to simply pip install faiss officially would be very nice. I dont want to move everything to conda just b/c I want to add faiss module.

I have a set up with docker and pipfile and everything is installed into the base python on deployment. I tried also getting conda in the Docker file and installing faiss that way, but it always installs it in a miniconda location. I want it to be in /usr/local/lib/python3.6/site-packages/. Not in some mini conda location. I also tried copying files from the miniconda location to the /usr/local/lib/python3.6/site-packages/ location but got some `ModuleNotFoundError: No module named '_swigfaiss'` errors.

Alternatively, if someone could provide instructions on downloading faiss from official conda but then getting into the base python  /usr/local/lib/python3.6/site-packages/, that would be welcomed too.

OlegSmelov:
Is there any particular reason you don't want to use [the unofficial build](https://github.com/facebookresearch/faiss/issues/170#issuecomment-512107346)? We've been using it for quite some time now without any issues.

DrChrisLevy:
@OlegSmelov , yes I have been using the unofficial build for quite some time and it’s great. If the unofficial one keeps being supported then great. I was just thinking more long term and as updates come out etc.

mdouze:
[...]

PR #536, opened 2018-07-20
Support PyPI friendly build using setuptools

kyamagu:
This PR replaces `python/setup.py` with `setup.py` that includes extension build script. Python users can easily install faiss with `pip`.

On Ubuntu 16.04 LTS:

GPU support is automatically built when nvcc compiler is available. Set
`CUDA_HOME` environment variable to specify where CUDA is installed.

    apt-get install swig libblas-dev liblapack-dev
    pip install numpy setuptools
    python setup.py bdist_wheel

On macOS (requires Homebrew):

    brew install llvm swig openblas
    pip install numpy setuptools
    python setup.py bdist_wheel

#170


mdouze:
Hmm... Not sure we want to change these build files again. @beauby?

beauby:
@kyamagu Thank you for your work. However, I am not sure I understand the benefits of building the shared library with python, rather than Makefile? Especially considering the fact that, as is, your version requires more effort to build against MKL, which is the recommended BLAS/LAPACK implementation.

kyamagu:
@beauby Thanks for reviewing this PR. The main point is the ability to host PyPI package in [a standard approach](https://packaging.python.org/tutorials/packaging-projects/). I have an environment where I cannot easily use `conda`. If Makefile is able to build up sdist as well as bdist to cover various environment, that would be fine.

As for MKL and other build options, I can adapt the setup to [PyTorch approach](https://github.com/pytorch/pytorch/blob/master/setup.py).

beauby:
@kyamagu I see. Note that using the `Makefile` along with the `configure` script with the right options, it is indeed possible to build the library for all supported platforms, and then the current `setup.py` can be used to generate the `sdist` / `bdist`.

ahkarami:
Dear @beauby,
Please add PyPI package for your fantastic repository.

tests/test_binary_io.py
```diff
@@ -76,7 +76,6 @@ def test_ivf_flat(self):
         D, I = index.search(self.xq, 3)
 
         tmpnam = tempfile.NamedTemporaryFile().name
-
```
OlegSmelov: Looks like unnecessary change

PR #544, opened 2018-07-25
[WIP] Make version available from within python.

beauby:
TL;DR: Faiss version is available in python as a string as `faiss.__version__`. 
It will also be available from C++ as `faiss::version::MAJOR` / `faiss::version::MINOR` / `faiss::version::PATCH` (integers), and `faiss::version::STRING` (string).


mdouze:
I think this is too complicated. I would like to avoid adding a py file just to set a version number.
The c++ version is available at faiss.cvar.MAJOR, the __version__ variable in faiss.py could just be set from that. 
Also I thought there would be some git magic to automatically set the version in the C++ source like cvs' keywords http://www.badgertronics.com/writings/cvs/keywords.html

beauby:
@mdouze AFAIK there is no such thing in git. However, we can have a target in the `Makefile` that "prepares the release", including updating the version according to the latest version tag.

PR #573, opened 2018-08-29
Facebook sync

beauby:
## Changelog:
Features: 
+ automatic tracking of C++ references in Python 
+ non-intel platforms supported -- some functions optimized for ARM
+ override nprobe for concurrent searches
+ support for floating-point quantizers in binary indexes

Bug fixes: 
+ no more segfaults in python (I know it's the same as the first feature but it's important!)
+ fix GpuIndexIVFFlat issues for float32 with 64 / 128 dims
+ fix sharding of flat indexes on GPU with index_cpu_to_gpu_multiple


mdouze:
checks are passing!!

PR #626, opened 2018-10-26
PYTHON executable location is missing.

esdotzed:


zamai:
Hey @esdotzed, I've also noticed that there are no Python executables in the Mac example files, can you also add them to your PR, so that your MR solve this issue for both systems, otherwise I'll open another PR for that.
Thanks!

beauby:
Thanks for the PR @esdotzed.

PR #644, opened 2018-11-23
use more clear code and add lacking header

cyyever:


gpu/impl/Distance.cu
```diff
@@ -19,6 +19,7 @@
 #include "../utils/BlockSelectKernel.cuh"
 
 #include <memory>
+#include <algorithm>
```
beauby: What is the rationale behind adding this header?

Enet4: To my understanding, this file is using at least `std::min`, which is defined by this header. Can't quite justify why GCC can still compile without this line though.

PR #661, opened 2018-12-14
IVFPQ filtering with polysemous codes

half-empty:
Query data is coded when "by_residuals" are closed, but it's empty when "by_residuals" are opened, it's just a mistake.
So, I added "compute_residual" and "compute_code" at "init_list".


mdouze:
Thanks for the PR. Normally the issue should be fixed in 1.5.0.

PR #676, opened 2018-12-25
Improve arm64 support.

tolysz:
Make library compile and run on arm64 8.2 (Nvidia Xavier arm+CUDA)
`configure` is regenerated

### DIFF
--- a/utils.cpp
+++ b/utils.cpp
@@ -21,15 +21,6 @@
 
 #include <omp.h>
 
-#ifdef __SSE__
-#include <immintrin.h>
-#endif
-
-#ifdef __aarch64__
-#include  <arm_neon.h>
-#endif
-
-
 #include <algorithm>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


mdouze:
Unfortunately I don't have an ARM platform for testing anymore. 
Travis builds seem broken.
@beauby could you take a look at the changes in the configure?

utils.cpp
```diff
@@ -21,7 +21,14 @@
...
 #include <omp.h>
 
+#ifdef __SSE__
 #include <immintrin.h>
+#endif
+
+#ifdef __aarch64__
+#include  <arm_neon.h>
+#endif
+
```
beauby: Could you remove all includes for simd, as it is not needed in this file?

beauby:
The `configure` part seems legit. Would you mind rebasing @tolysz?

mdouze:
@beauby, is this failing with the current faiss as well?

tolysz:
thanks!

PR #683, opened 2019-01-04
 add verbose check when train_q1

hxfxjun:
add verbose check when train_q1


mdouze:
The proper way of setting the clustering verbosity is via `index.cp.verbose`, so this does not seem necessary.

Issue #669, opened 2018-12-21
Packaging details for building against FAISS C++ API


cjnolet:
The documentation describes that FAISS builds shared & static libraries but these do not appear to end up in the installed Python packaging. Rather, the Python conda packaging only installs the swig-converted artifacts. 

If a codebase, dependent on FAISS, is being built against the libfaiss.so and/or libfaiss_gpu.so artifacts, this leaves developers with the option of forcing contributors to build FAISS themselves, or including the libs in the codebase. 

Further clarification on the ideal way of packaging against FAISS C++ API's would be ideal.

cjnolet:
Assuming the users build your source before building libraries that depend on it, Would it make sense to have FAISS install the so files into a common directory already on the LD path (e.g. `/user/lib64/faiss`)?

We would also need to install the include files in `/usr/include` as well. This would allow other projects to depend on FAISS and just point their users to the FAISS source.

mdouze:
Absolutely. IMO the proper paths would be 
```
/usr/lib64/faiss.so
/usr/include/faiss/Index.h
/usr/include/faiss/IndexFlat.h
...
```
I will talk to @beauby on where we can standardize this an if we can distribute a binary package.

cjnolet:
Sounds good. I certainly don't mind providing a PR.

cjnolet:
@mdouze @beauby, I'm preparing to get some changes merged in a project that depend on FAISS. Do you have any objections to changing the default libname / includedir in the build to be /usr/lib64 & /usr/include?

beauby:
@cjnolet The current default values for `libdir`/`includedir` come from autoconf (see [here](https://github.com/facebookresearch/faiss/blob/master/makefile.inc.in#L49). I do not feel strongly about it though, but in the meantime you can do `./configure --libdir=/usr/lib64 --includedir=/usr/include`.

PR #688, opened 2019-01-08
For #669. Adding install target to gpu Makefile

cjnolet:
This trivial change installs both the FAISS non-gpu and gpu library artifacts into the same output location so that downstream projects can build against them.

### DIFF
--- a/gpu/Makefile
+++ b/gpu/Makefile
@@ -85,4 +85,12 @@ depend:
 	    $(CXXCPP) $(CPPFLAGS) -x c++ -MM $$i; \
 	done > depend
 
+install: libgpufaiss.a libgpufaiss.$(SHAREDEXT) installdirs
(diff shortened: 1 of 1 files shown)


gpu/Makefile
```diff
@@ -85,4 +85,12 @@ depend:
 	    $(CXXCPP) $(CPPFLAGS) -x c++ -MM $$i; \
 	done > depend
 
+install: libgpufaiss.a libgpufaiss.$(SHAREDEXT) installdirs
+	cp libgpufaiss.a libgpufaiss.$(SHAREDEXT) $(DESTDIR)$(libdir)
+	cp *.h $(DESTDIR)$(includedir)/faiss/gpu
+	cp --parents **/**.h $(DESTDIR)$(includedir)/faiss/gpu
```
beauby: The `--parents` option of `cd` is not standard but as this target will likely only run on Linux for now it's not a big deal.

Issue #710, opened 2019-01-27
Issue while building the Docker image.


Arun-George-Zachariah:
```nvidia-docker build -t faiss github.com/facebookresearch/faiss``` fails with the error :
```
Step 5/12 : RUN pip install matplotlib
 ---> Running in f15f72d724b6
Collecting matplotlib
  Downloading https://files.pythonhosted.org/packages/89/0c/653aec68e9cfb775c4fbae8f71011206e5e7fe4d60fcf01ea1a9d3bc957f/matplotlib-3.0.2.tar.gz (36.5MB)
    Complete output from command python setup.py egg_info:
    
    Matplotlib 3.0+ does not support Python 2.x, 3.0, 3.1, 3.2, 3.3, or 3.4.
    Beginning with Matplotlib 3.0, Python 3.5 and above is required.
    
    This may be due to an out of date pip.
    
    Make sure you have pip >= 9.0.1.
    
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-Gmv72I/matplotlib/
You are using pip version 8.1.1, however version 19.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
The command '/bin/sh -c pip install matplotlib' returned a non-zero code: 1
```

beauby:
A new docker image is in the making. I'll keep this issue open until it's ready.

mdouze:
@beauby any news?

beauby:
@mdouze This specific issue was fixed by merging #710, and the new Dockerfile will be ready by EOW.

PR #711, opened 2019-01-27
Upgrading pip in the Docker Container (#710)

Arun-George-Zachariah:
Changes done for the issue #710


mdouze:
@beauby could you take a look?

Issue #709, opened 2019-01-26
reconstruct not supported with IndexShards->IDMap2->IVFFlat


bvarghese1:
# Summary

I am trying to perform a reconstruct on an IndexShard which is having an IDMap2 index pointing to an IVFFlat index.

Running on:
- [ Y] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [ ] Python

# Reproduction instructions

Here is my code snippet:
```
num_shards = 10
index = faiss.IndexShards(D, True, False) # D - dim, True - threaded, False - successive_ids
for i in range(num_shards):
     shard = faiss.read_index(os.path.join(index_path, 'shard_{}'.format(i)), faiss.IO_FLAG_MMAP)
     index.add_shard(shard)

_id = 814
embedding = index.reconstruct(_id)
```

Error:
```
index.reconstruct(cid)
*** RuntimeError: Error in virtual void faiss::Index::reconstruct(idx_t, float *) const at Index.cpp:56: reconstruct not implemented for this type of index
```

To overcome this, I am doing the following in my python code:
```
if _id > index.ntotal:
    raise Exception('{} is invalid'.format(_id))
embedding = None
for i in range(num_shards):
    try:
        embedding = index.at(i).reconstruct(cid)
        break
    except:
        pass
if embedding is None:
    print ('{} not found'.format(_id))
```

Is it possible to implement something similar in the IndexShards.cpp as a convenience method?

bvarghese1:
@beauby I would like to work on this.

PR #715, opened 2019-01-29
Add support for reconstruct in IndexShards (#709)

bvarghese1:


MetaIndexes.cpp
```diff
@@ -564,7 +564,21 @@ void IndexShards::search (
...
-
+void IndexShards::reconstruct (idx_t key, float * recons) const
+{
+    for (int i = 0; i < shard_indexes.size(); i++) {
+        try {
+            shard_indexes[i]->reconstruct (key, recons);
+            if (verbose)
+                printf ("key %ld found in shard %d\n", key, i);
+            return;
+        } catch (const std::out_of_range& e) {
```
Enet4: Is it accurate to assume an exception of this type being raised when the key is not found? From what I saw in one of the implementations, [`IndexPQ`](https://github.com/facebookresearch/faiss/blob/master/IndexPQ.cpp#L126), `reconstruct` might throw a `FaissException` instead.

bvarghese1: @Enet4 Thanks! Would it useful to catch both the exceptions and have different messages for it in verbose mode or simply replace out_of_range with FaissException for brevity?

mdouze:
Faiss only throws `FaissException`s. 
I reject this PR because we are working on a way to properly propagate exceptions from `IndexShards` sub-indices.

bvarghese1:
@mdouze makes sense. Thanks.

mdouze:
Yes, let's give it another try then.

PR #723, opened 2019-02-09
ondisk demo bug fix: data needs to be added with ids

seominjoon:
When data is being added to each index block, it needs to explicitly identify ids. Otherwise, the ids are automatically assumed to be the index of each individual data block, so every block has the same ids (in the range of 0 to 250,000).


mdouze:
Thanks for the fix! It is unclear why we did not spot this before, probably just not checked the accuracy of the resulting index.

PR #756, opened 2019-03-28
Facebook sync (Mar 2019)

beauby:
## Changelog

- MatrixStats object
- option to round coordinates during k-means optimization
- alternative option for search in HNSW
- moved stats and imbalance_factor of IndexIVF to InvertedLists object
- range search for IVFScalarQuantizer
- direct unit8 codec in ScalarQuantizer
- renamed IndexProxy to IndexReplicas and moved to main Faiss
- better support for PQ code assignment with external index
- support for IMI2x16 (4B virtual centroids!)
- support for k = 2048 search on GPU (instead of 1024)
- most CUDA mem alloc failures throw exceptions instead of terminating on an assertion
- support for renaming an ondisk invertedlists
- interrupt computations with ctrl-C in python


IndexShards.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+namespace faiss {
+
+// subroutines
+namespace {
+
+typedef Index::idx_t idx_t;
+
+/// callback + thread management to train 1 shard
+struct TrainJob {
```
mdouze: this is not the latest FB version

benchs/bench_all_ivf/bench_all_ivf.py
```diff
@@ -1,9 +1,3 @@
-# Copyright (c) 2015-present, Facebook, Inc.
-# All rights reserved.
-#
-# This source code is licensed under the BSD+Patents license found in the
-# LICENSE file in the root directory of this source tree.
```
mdouze: unclear why this (c) notice disappears

benchs/link_and_code/README.md
```diff
@@ -11,14 +11,12 @@ Link & code is an indexing method that combines HNSW indexing with
 compression and exploits the neighborhood structure of the similarity
 graph to improve the reconstruction. It is described in
 
-```
```
mdouze: do not remove the quotes

demos/demo_ondisk_ivf.py
```diff
@@ -4,6 +4,12 @@
 # This source code is licensed under the BSD+Patents license found in the
 # LICENSE file in the root directory of this source tree.
 
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
```
mdouze: duplicate license

Issue #751, opened 2019-03-26
GPU compilation fails: Tensor.cuh(362): type name is not allowed


Steffen911:
# Summary

I follow the Install.md instructions from step 1 to step 3. General compilation and Python setup worked fine, but the GPU compilation fails with:

```
$ make
/usr/local/cuda/bin/nvcc -I /usr/local/cuda/targets/x86_64-linux/include/ -Xcompiler -fPIC -Xcudafe --diag_suppress=unrecognized_attribute -gencode arch=compute_35,code="compute_35" -gencode arch=compute_52,code="compute_52" -lineinfo -ccbin g++ -std=c++11 -DFAISS_USE_FLOAT16 -g -O3 -c impl/BinaryDistance.cu -o impl/BinaryDistance.o
impl/../utils/Tensor.cuh(362): error: type name is not allowed

impl/../utils/Tensor.cuh(362): error: expected an expression

2 errors detected in the compilation of "/tmp/tmpxft_000010e7_00000000-4_BinaryDistance.cpp4.ii".
make: *** [Makefile:71: impl/BinaryDistance.o] Error 2
```
The same error occurs multiple times if I follow the recommendation to use multiple threads during make.

# Platform

OS:  Fedora 29

Faiss version:  a9959bf63b930da491d3e2bce00ff1a675d3ddcf

mdouze:
Which versions of cuda and g++ are you using ?

Steffen911:
g++:
```
$ g++ --version
g++ (GCC) 8.3.1 20190223 (Red Hat 8.3.1-2)
```
cuda:
```
$ /usr/local/cuda/bin/nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2019 NVIDIA Corporation
Built on Fri_Feb__8_19:08:17_PST_2019
Cuda compilation tools, release 10.1, V10.1.105
```

redmajor:
This error happened to me as well, and I fixed it by adding the "template" keyword after arg on the faulty line.

return arg.template canUseIndexType<IndexType>() &&
    canUseIndexType(args...);

mdouze:
@beauby can we repro this with the compilers used for Travis?
@redmajor, @Steffen911 would you mind submitting a PR with the suggested change so that we can check if it breaks something else?

beauby:
[...]

PR #771, opened 2019-04-04
fix: gpu compilation with cuda 10.1

Steffen911:
The GPU compilation fails with g++ 8.3.1 and cuda 10.1
With this proposed fix the compilation works locally

Contributes to #751


beauby:
Merging, thanks!

Issue #816, opened 2019-05-05
Unable to read a written untrained index


pletessier:
# Summary

A written index cannot be read if it has not been trained.

# Reproduction instructions

```
import numpy as np
import faiss

d = 16
nlist = 4
m = 8

quantizer = faiss.IndexFlatL2(d)
index = faiss.IndexIVFPQ(quantizer, d, nlist, m, 8)

n_training_features = 10000
training_features = np.random.random((n_training_features, d)).astype('float32')

# Without this step, it won't work
# index.train(training_features)

faiss.write_index(index, "myindex")
new_index = faiss.read_index("myindex")
```

beauby:
Closing this as the corresponding PR has been merged.

PR #817, opened 2019-05-05
Fix unable to read untrained index

pletessier:
Fix proposal for #816


beauby:
LGTM, merging. Thanks!

PR #858, opened 2019-06-12
[C API] Update for compatibility with v1.5

Enet4:
The recent major updates to faiss included a few API changes which made the C API no longer compile. The changes made (small summary below) aim to make things work again, rather than exposing the latest features. The latter can be done in future PRs, and I'll be listening to related issues in the event that someone is interested in them.

- Move `IndexShards` (out of MetaIndexes_c) into a dedicated module IndexShards_c
- Remove getter/setters to unreachable fields and non-existing methods in the latest API
- Reimplement faiss_IndexIVF_imbalance_factor (to retrieve the same output from the inner `invlists`)
- Remove `FaissQueryResult`, provide `FaissRangeQueryResult` instead
- Minor documentation tweaks
- The C API was relicensed to MIT for consistency with the rest of the project. Hopefully this change is desirable, do you agree?

### DIFF
--- a/c_api/AutoTune_c.cpp
+++ b/c_api/AutoTune_c.cpp
@@ -1,8 +1,7 @@
 /**
- * Copyright (c) 2015-present, Facebook, Inc.
- * All rights reserved.
+ * Copyright (c) Facebook, Inc. and its affiliates.
  *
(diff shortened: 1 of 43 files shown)


beauby:
Thanks @Enet4 – this is indeed desirable.

Issue #848, opened 2019-05-31
Support distances from `scipy.spatial.distance`


cjnolet:
There are several algorithms in cuML which need support for different distances. For instance, our NearestNeighbors algorithm should be able to support the distances that sklearn's NearestNeighbors supports. 

UMAP-contrib uses nearest neighbor descent and is able to support all of the distances in `scipy.spatial.distance, using a pairwise evaluation function for those which are not L_p based. 

It would be really helpful if we were able to do this within FAISS, both supporting more L_p variants within the brute force kNN computation and supporting more distance types in the ANN algorithms overall.

wickedfoo:
I see. I will push this up on the priority list and write a pairwise distance kernel (akin to GEMM but more general) that will have a pluggable functor for the implementation, but these functors will be chosen at compile time.

mdouze:
Supporting different types of distances should not be too hard, the issues are: 
- some metrics need an additional parameter (eg. L_p), that has to be carried around
- will IP remain the unique metric that has to be maximized rather than minimized?

mdouze:
From scipy.spatial.distance: 

- braycurtis(u, v[, w]): OK (=> can be supported)
- canberra(u, v[, w]): OK
- chebyshev(u, v[, w]): OK
- cityblock(u, v[, w]): OK
- correlation(u, v[, w, centered]): is a max. CEntering should be done as pre-processing
- cosine(u, v[, w]): equivalent to IP with preprocessing
- euclidean(u, v[, w]): already supported (L2) 
- jensenshannon(p, q[, base]): OK
- mahalanobis(u, v, VI): equivalent to L2 with preprocessing
- minkowski(u, v[, p, w]): requires an additonal parameter
- seuclidean(u, v, V): unclear what that is ??? 
- sqeuclidean(u, v[, w]): equivalent of L2 with post-processing
- wminkowski(u, v, p, w): equivalent of Minkowski with pre-processing

Observations: 
- many distances are equivalent up to a pre- or post-processing. These should be carried out by the caller. 
- most metrics take a weight vector in addition to the vectors to compare -> is a pre-processing
- in fact only L_p has an additional parameter (aka minkowski)

cjnolet:
I wanted to check in on this in hopes there may be a timeline to get these distances supported. We could really use this feature in UMAP on RAPIDS cuML.

wickedfoo:
@cjnolet I'm starting this week on this, sorry for the delay.

[...]

Issue #836, opened 2019-05-23
Why does it take more time to search the entire xq than to search sequentially for a subset of all the shards of the xq?


chenyihang1993:
There are 1B samples in xq. I get the searching result by two different ways:
1. Using IVF65536_HNSW32,PQ64 index, train it and add whole xq in it, then search.
2. Using IVF65536_HNSW32,PQ64 index, train it and cut xq into 10 subsets(there are 100M samples in it), then add subsets of xq separately to 10 indexes and search separately, finally merge the result.
The first way takes more time than the second. Please tell me the reason. THX.

mdouze:
With as baseline case 1:
- Searching a dataset of 100M that is a subset of the 1B should be faster because the inverted lists are shorter.
- Searching 10 datasets of 100M should be slower because it is performing the same amount of comparisons but the overhead is higher.
- Doing the same but in parallel (eg. via an IndexShards) could be faster if you are performing searches in one thread. 

If you are not in one of those cases, please comment.

chenyihang1993:
The strange thing is that searching 10 datasets of 100M is faster than searching a dataset of 1B. This is my case.

chenyihang1993:
There is a demo. The length of xb is 1M, and the index is 'IVF100_HNSW32,PQ64'.
```
import numpy as np
import faiss
import time

d = 64  # dimension
nb = 1000000  # database size
nq = 1000  # nb of queries
nt = 10000
np.random.seed(1234)  # make reproducible
xb = np.random.random((nb, d)).astype('float32')
xq = np.random.random((nq, d)).astype('float32')
xt = np.random.random((nt, d)).astype('float32')
k = 100

index_1 = faiss.index_factory(d, 'IVF100_HNSW32,PQ64')
index_1.train(xt)
index_1.add(xb)
faiss.ParameterSpace().set_index_parameter(index_1, 'nprobe', 64)
start = time.time()
index_1.search(xq, k)
print('time of searching: ' + str(time.time() - start))

index_1.reset()
faiss.write_index(index_1, 'base.index')

subset_len = 100000
sum_time = 0
for i in range(0, nb, subset_len):
    index = faiss.read_index('base.index')
    faiss.ParameterSpace().set_index_parameter(index, 'nprobe', 64)
    index.add(xb[i:i + subset_len])
    start = time.time()
    index.search(xq, k)
    sum_time = sum_time + time.time() - start

print('time of searching separately:' + str(sum_time))
```
The result is 
```
[...]

Issue #780, opened 2019-04-10
[Feature Request] IdMap/IdMap2 for IndexBinary


bduclaux:
Hello,

As far as I know, there is currently no way to use an IndexIDMap / IndexIDMap2 with binary indexes, as the IDMap classes derive from the Index class, and not the IndexBinary class.

Such code does not compile:
```
faiss::IndexBinaryHNSW * hnsw=new faiss::IndexBinaryHNSW(dim, 32);
faiss::IndexIDMap2 * hnsw_index=new faiss::IndexIDMap2(hnsw);
```
It would be nice to have IndexBinaryIDMap / IndexBinaryIDMap2 classes in a future release, with serialization support in index_io.cpp as for other indexes.

The current workaround is to store mappings between ids and record numbers in a separate binary file.

Best

mdouze:
Yes we are progressively extending the meta-indexes to the binary indexes as well.

mdouze:
Implemented in 1.5.3

Issue #758, opened 2019-03-29
 std::bad_alloc


linghuang:
# Summary

# Platform

OS: Ubuntu Server 16.04 

Faiss version:  GPU version

Faiss compilation options:  conda install faiss-gpu -c pytorch 

Running on:
- [ ] GPU

Interface: 
- [ ] Python

# minimal code to reproduce the issue：
import faiss
import numpy as np

d = 2048
index = faiss.IndexFlatIP(d)

#img_feature_list is a numpy array with dimension of 30000 * 2048
img_feature_list = np.load("feature.npy")      

for i in range(40): index.add(img_feature_list)
# Reproduction instructions

RAM size is 32G

Run this code, It throws the following error:
“terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)”   

But the  memory hasn't ran out. The memory is only occupied about  67%. 
what to do with this ?

mdouze:
@linghuang please provide a proper bug report
https://github.com/facebookresearch/faiss/wiki/Troubleshooting#reporting-bugs

mdouze:
Sorry, I missed the "minimal code" section. 
Memory allocation is done with the `std::vector` `insert` function 
https://github.com/facebookresearch/faiss/blob/master/IndexFlat.cpp#L32
the memory reallocation policy of `std::vector` is to increment by powers of 2. So if you already use 67% RAM, it will go to 2*67% > 100% RAM. 
There is no clear workaround here unfortunately. Giving up the geometric reallocation would have an impact on performance.

mdouze:
Implemented in 1.5.3

PR #862, opened 2019-06-13
Facebook sync (Jun 2019)

beauby:
### Bugfixes:
- slow scanning of inverted lists (#836).

### Features:
- add basic support for 6 new metrics in CPU `IndexFlat` and `IndexHNSW` (#848);
- add support for `IndexIDMap`/`IndexIDMap2` with binary indexes (#780).

### Misc:
- throw python exception for OOM (#758);
- make `DistanceComputer` available for all random access indexes;
- gradually moving from `long` to `int64_t` for portability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
             - libblas-dev
             - liblapack-dev
             - python-numpy
+            - python-scipy
             - python-dev
 #            - swig3.0
       env:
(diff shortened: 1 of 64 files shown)


mdouze:
Good to go IMO.

mdouze:
@Enet4 thanks for flagging. We forgot to test without the -mavx2 flag, which exposes this bug.

beauby:
Only test left failing on OSX is a matter of system python not finding Homebrew-installed scipy.

PR #869, opened 2019-06-21
[C API] June 2019 update

Enet4:
This is a set of changes to adapt the C API to the latest major updates. Quick summary:

- Update Makefiles in c_api to include the relevant flags from makefile.inc
- Type changes: use `idx_t` or `size_t` where appropriate instead of `long`
- Remove a redundant declaration of `FaissIndexIVF`


beauby:
Thanks @Enet4 – merging.

PR #964, opened 2019-09-25
Replace print with logging

kyamagu:
`print()` has no control over how messages are displayed. Use `logging` module instead.

### DIFF
--- a/python/faiss.py
+++ b/python/faiss.py
@@ -16,6 +16,10 @@ import inspect
 import pdb
 import platform
 import subprocess
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 def instruction_set():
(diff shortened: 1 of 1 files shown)


mdouze:
This may be a good idea, the new import is a bit verbose.

python/faiss.py
```diff
@@ -35,15 +39,15 @@ def instruction_set():
...
         from .swigfaiss_avx2 import *
     else:
-        print("Loading faiss.", file=sys.stderr)
+        logger.info("Loading faiss.")
         from .swigfaiss import *
 
 except ImportError:
     # we import * so that the symbol X can be accessed as faiss.X
-    print("Loading faiss.", file=sys.stderr)
+    logger.error("Loading faiss.")
```
konstin: Does this need to be on error level? Afaik this is expected to happen if the platform support avx2 but faiss is compiled without `swigfaiss_avx2`.

mdouze:
While we are at it, it could be useful to add a message that says whether GPU is supported and how many GPUs are detected (this does not change during the process lifetime).

python/faiss.py
kyamagu: All of the logging could be info or lower level. What is the expectation of swigfaiss_avx2 build for general use? I couldn't find documentation on avx2 support in `INSTALL.md`

mdouze:
Current plan is to add this change to 1.6.1

PR #987, opened 2019-10-15
Use intrinsics for float16 in GPU scalar quantizer.

beauby:


mdouze:
@beauby what is the status of this PR?

PR #943, opened 2019-09-10
Facebook sync (2019-09-10)

beauby:

PR #989, opened 2019-10-15
[C_API] Pass compilation in current state

xolott:
The current version of Faiss has c_api compiling issues due to the last v1.6.0 changes.

This pull request only changes the necessary to compile without problem.


c_api/index_io_c.cpp
```diff
@@ -15,35 +15,40 @@
...
+int faiss_read_index_fname(const char *fname, int io_flags, FaissIndex **p_out)
+{
+    try
+    {
         auto out = faiss::read_index(fname, io_flags);
-        *p_out = reinterpret_cast<FaissIndex*>(out);
-    } CATCH_AND_HANDLE
-}
-
-int faiss_clone_index (const FaissIndex *idx, FaissIndex **p_out) {
```
Enet4: `clone_index` still appears to be in the library, declared in [clone_index.h](https://github.com/facebookresearch/faiss/blob/master/clone_index.h).

xolott:
@Enet4 I'm using visual studio code with some auto-formatting extension. Sorry about that, I will revert the formating to align with the rest of the c_api.
 I have a question (I'm not a C/C++ expert): Why in every ###_new function is expected to receive a double pointer to `Index` and not a single pointer.

Enet4:
> I have a question (I'm not a C/C++ expert): Why in every ###_new function is expected to receive a double pointer to `Index` and not a single pointer.

It is indeed a double pointer, to serve as an output parameter to a function which provides an extra "return" value. It would be used like this:

```c
FaissIndex* index = NULL;
int e = FlatIndex_new(&index, other_args); // passed a FaissIndex**
```

In other words, we're passing a place for the function to write a pointer to a newly created instance. How the actual instance is allocated is left as an implementation detail. Were there a single indirection point, the pattern would have to be different: the consumer would have to allocate space for the index and the function would need to construct the instance on that space. For the Faiss indexes, I felt the former approach to be more intuitive.

xolott:
This is the list of changes:

- Fixed formatting issues.
- Based on [this log](https://github.com/facebookresearch/faiss/pull/943#issuecomment-532590627):
```
index_io_c.h: split off clone_index_c.h
AuxIndexStructures_c.h: moved to impl/AuxIndexStructures_c.h
AutoTune_c.h: split off index_factory_c.h
```
- Updated Makefile with new files and modified paths
- Updated `example_c.c` to include `clone_index_c.h`
- Included `faiss_Index_compute_residual_n` to `Index_c.h`

c_api/impl/AuxIndexStructures_c.h
```diff
@@ -126,8 +126,24 @@ int faiss_RangeSearchPartialResult_set_lims(
 int faiss_RangeSearchPartialResult_new_result(
     FaissRangeSearchPartialResult* res, idx_t qno, FaissRangeQueryResult** qr);
 
+
+FAISS_DECLARE_CLASS(DistanceComputer)
```
Enet4: The respective implementations of the functions below appear to be missing. Please follow the pattern seen in the other opaque types to fill in the missing definitions. You can also copy the rest of the documentation to each function where applicable.

c_api/impl/AuxIndexStructures_c.h
```diff
@@ -126,8 +126,24 @@ int faiss_RangeSearchPartialResult_set_lims(
...
+int faiss_DistanceComputer_new(FaissDistanceComputer** p_dc);
+int faiss_DistanceComputer_set_query(FaissDistanceComputer *dc, const float *x);
+int faiss_DistanceComputer_vector_to_query_dis(FaissDistanceComputer *dc, idx_t i, float *qd);
+/** 
+ * Compute distance between two stored vectors
+ *
+ * @param dc          opaque pointer to distance computer object
+ * @param key         output distance
+ **/
+int faiss_DistanceComputer_symmetrik_dis(FaissDistanceComputer *dc, idx_t i, idx_t j, float *vd);
```
Enet4: It's `symmetric_dis` in the main API, it might be better to stick to the same name.

```suggestion
int faiss_DistanceComputer_symmetric_dis(FaissDistanceComputer *dc, idx_t i, idx_t j, float *vd);
```

c_api/impl/AuxIndexStructures_c.h
```diff
@@ -126,8 +126,24 @@ int faiss_RangeSearchPartialResult_set_lims(
...
+
+FAISS_DECLARE_CLASS(DistanceComputer)
+int faiss_DistanceComputer_new(FaissDistanceComputer** p_dc);
+int faiss_DistanceComputer_set_query(FaissDistanceComputer *dc, const float *x);
+int faiss_DistanceComputer_vector_to_query_dis(FaissDistanceComputer *dc, idx_t i, float *qd);
+/** 
+ * Compute distance between two stored vectors
+ *
+ * @param dc          opaque pointer to distance computer object
+ * @param key         output distance
```
Enet4: It's named `vd` below.

```suggestion
 * @param vd          output distance
```

c_api/impl/AuxIndexStructures_c.h
xolott: I just finish the definitions. But I have a trouble with `DistanceComputer::operator()`... I changed the name to `faiss_DistanceComputer_vector_to_query_dis`, but I want to read your suggestions about a better name.

c_api/impl/AuxIndexStructures_c.h
xolott: My bad... a typo

c_api/impl/AuxIndexStructures_c.h
Enet4: Well, I don't have a strong opinion regarding the name of the operator, but it's worth documenting that this function defers to the overloaded function call operator, so that users know where it comes from in the native API.

xolott:
All changes are already in the branch. If there is no other observation, this could be merge :smiley:

c_api/impl/AuxIndexStructures_c.h
```diff
@@ -126,8 +126,24 @@ int faiss_RangeSearchPartialResult_set_lims(
...
     FaissRangeSearchPartialResult* res, idx_t qno, FaissRangeQueryResult** qr);
 
+
+FAISS_DECLARE_CLASS(DistanceComputer)
+/// called before computing distances
+int faiss_DistanceComputer_set_query(FaissDistanceComputer *dc, const float *x);
+
+/**
+ * Compute distance of vector i to current query.
+ * This function correspond to the call operator DistanceComputer::operator()()
```
Enet4:
```suggestion
 * This function corresponds to the function call operator: DistanceComputer::operator()
```

c_api/impl/AuxIndexStructures_c.h
```diff
@@ -126,8 +126,24 @@ int faiss_RangeSearchPartialResult_set_lims(
...
+ * */
```
Enet4:
```suggestion
 */
```

xolott:
Done!

xolott:
No problem. Metrics updated

mdouze:
Merging, since @Enet4 approves. 
Sorry for the lag.

PR #992, opened 2019-10-17
Fixes to enable building with -mavx, for Sandybridge CPUs.

mattg-sp:
The -mf16c (i.e. packed-half conversion) instructions weren't added until the
Ivy Bridge CPU generation (https://en.wikipedia.org/wiki/F16C#CPUs_with_F16C).
Therefore, code dependent on those intrinsics must be conditioned on __F16C__ -
not __AVX__.

Without this patch, the code will not build with -mavx and not also -mf16c or -march=haswell (or later).  After patching, it builds with -mavx and still enables AVX in pure-AVX1 codepaths, while also enabling f16c instructions when an appropriate -m flag is specified.

### DIFF
--- a/impl/ScalarQuantizer.cpp
+++ b/impl/ScalarQuantizer.cpp
@@ -43,6 +43,9 @@ namespace faiss {
 #define USE_AVX
 #endif
 
+#ifdef __F16C__
+#define USE_F16C
+#endif
 
 
 namespace {
@@ -183,7 +186,7 @@ struct Codec6bit {
 
 
 
-#ifdef USE_AVX
+#ifdef USE_F16C
 
 
 uint16_t encode_fp16 (float x) {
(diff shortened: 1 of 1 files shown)


mdouze:
Looks good. Let's see what @beauby thinks of it...

mattg-sp:
Thanks for reviewing.

I chased down the issue after I got a SIGILL.  Once I found a f16c intrinsic being used, I changed the compiler flags and converted that #ifdef from USE_AVX to USE_F16C.  Then, as I hit more compiler errors, I just started switching more, one-by-one, until it compiled for Sandybridge.

However, to be honest, I didn't really investigate whether there was a way to achieve better performance on Sandybridge - I figured if at least some of the USE_AVX are still intact, then at least it has some benefit over the generic version.

Anyway, thanks for the great library.

PR #997, opened 2019-10-18
Pass forward prefix option to python install target

Purg:
The python Makefile install target did not include specifying the same install prefix that is given at configure time and used for the native library and headers. This can cause the python module to be installed into a different location than the libraries/headers when a custom install target is specified. Adding ``--prefix=$(prefix)`` tells the ``setup.py install`` to also install in the same root context as the library/headers.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -30,7 +30,7 @@ build: _swigfaiss.so faiss.py
 	$(PYTHON) setup.py build
 
 install: build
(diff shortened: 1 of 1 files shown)


beauby:
Thanks for the fix. However, as the build system evolved, this PR is no longer relevant. Please feel free to reopen if needed.

PR #1006, opened 2019-10-29
Fix GPUSQ copyto bug

JinHai-CN:
In  gpu/GpuIndexIVFScalarQuantizer.cu copyTo function. Index->code_size isn't set. And the code_size will be used in creating new invertedlist, which will lead to error.

Solution:
code_size is stored in scalar_quantizer, just assign it to index->code_size.


mdouze:
looks useful to me. @wickedfoo ?

lubas569:
How i can check pls

PR #1011, opened 2019-11-01
Fix GpuIndexIVF::setNumProbes bug

JinHai-CN:
nprobe = nprobe; won't set object member: nprobe.


mdouze:
Accepting commit now. We will roll out a new version with this fix and a few others soon.

JinHai-CN:
@mdouze，when the new version is released?

PR #1060, opened 2019-12-11
add header for faiss::gpu::index_cpu_to_gpu_multiple(line 52)

gameofdimension:
image


mdouze:
Thanks for this diff. 
I fixed the multi-gpu example. 
For the rest, could you describe what this diff is about: what is the coverage of the Java wrapper? If the swig + makefiles are unique it is perhaps not useful to put it in Faiss itself.

gameofdimension:
> Thanks for this diff.
> I fixed the multi-gpu example.
> For the rest, could you describe what this diff is about: what is the coverage of the Java wrapper? If the swig + makefiles are unique it is perhaps not useful to put it in Faiss itself.

you are right. thank you

PR #1065, opened 2019-12-19
Add support to use selected GPUs by id

aruslantsev:
One can pass parameter gpus to index_cpu_to_gpus_list to use only selected GPU's, i.e. if we pass gpus=[1, 3, 4], we will use 1-st, 3-rd and 4-th graphical cards

### DIFF
--- a/python/faiss.py
+++ b/python/faiss.py
@@ -469,6 +469,8 @@ def index_cpu_to_gpu_multiple_py(resources, index, co=None, gpus=None):
     """ builds the C++ vectors for the GPU indices and the
     resources. Handles the case where the resources are assigned to
     the list of GPUs """
+    if gpus is None:
+        gpus = range(len(resources))
     vres = GpuResourcesVector()
     vdev = IntVector()
     for i, res in zip(gpus, resources):
--- a/python/faiss.py
+++ b/python/faiss.py
@@ -481,9 +481,33 @@ def index_cpu_to_gpu_multiple_py(resources, index, co=None):
 def index_cpu_to_all_gpus(index, co=None, ngpu=-1):
     if ngpu == -1:
         ngpu = get_num_gpus()
-    res = [StandardGpuResources() for i in range(ngpu)]
-    index2 = index_cpu_to_gpu_multiple_py(res, index, co)
-    return index2
+    res = [StandardGpuResources() for _ in range(ngpu)]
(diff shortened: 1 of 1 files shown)


python/faiss.py
```diff
@@ -481,9 +481,33 @@ def index_cpu_to_gpu_multiple_py(resources, index, co=None):
 def index_cpu_to_all_gpus(index, co=None, ngpu=-1):
     if ngpu == -1:
         ngpu = get_num_gpus()
-    res = [StandardGpuResources() for i in range(ngpu)]
```
mdouze: please call index_cpu_to_gpus_list

python/faiss.py
```diff
@@ -481,9 +481,33 @@ def index_cpu_to_gpu_multiple_py(resources, index, co=None):
...
-    index2 = index_cpu_to_gpu_multiple_py(res, index, co)
-    return index2
+    res = [StandardGpuResources() for _ in range(ngpu)]
+    index_gpu = index_cpu_to_gpu_multiple_py(res, index, co)
+    return index_gpu
+
+
+def index_cpu_to_gpu_selected_py(gpus, resources, index, co=None):
```
mdouze: please merge this function with index_cpu_to_gpu_multiple_py or have index_cpu_to_gpu_multiple_py call this function to avoid code duplication.

python/faiss.py
aruslantsev: done

python/faiss.py
aruslantsev: done

python/faiss.py
```diff
@@ -465,25 +465,35 @@ def replacement_function(*args):
...
 
-def index_cpu_to_gpu_multiple_py(resources, index, co=None):
-    """builds the C++ vectors for the GPU indices and the
-    resources. Handles the common case where the resources are assigned to
-    the first len(resources) GPUs"""
+def index_cpu_to_gpu_multiple_py(resources, index, co=None, gpus=None):
+    """ builds the C++ vectors for the GPU indices and the
+    resources. Handles the case where the resources are assigned to
+    the list of GPUs """
     vres = GpuResourcesVector()
```
mdouze: Does not work with the default gpus=None

aruslantsev: fixed

mdouze:
Thanks!

PR #1067, opened 2019-12-24
Enable Tensor Core for cublasHgemm

Jie-Fang:
This PR makes faiss can use Tensor Core to accelerate computation when useFloat16 and useFloat16Accumulator are set.

In my testing, FP16 computation on Tensor Core is 2.2 faster than FP32 computation.

### DIFF
--- a/gpu/utils/Float16.cu
+++ b/gpu/utils/Float16.cu
@@ -20,6 +20,11 @@ bool getDeviceSupportsFloat16Math(int device) {
           (prop.major == 5 && prop.minor >= 3));
 }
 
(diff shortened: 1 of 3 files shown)


wickedfoo:
Looks fine to me! Will merge internally at FB first then do this here.

wickedfoo:
The next Faiss push from the FB version will include this change.

wickedfoo:
I did this change in a different way and it is now part of the repo.

https://github.com/facebookresearch/faiss/blob/master/gpu/StandardGpuResources.cpp#L254

PR #1094, opened 2020-01-27
c_api: faiss_IndexIVF_invlists_get_ids function

guskovd:
I would like to be able to get invlists from c_api

### DIFF
--- a/c_api/IndexIVF_c.h
+++ b/c_api/IndexIVF_c.h
@@ -114,6 +114,8 @@ double faiss_IndexIVF_imbalance_factor (const FaissIndexIVF* index);
 /// display some stats about the inverted lists of the index
 void faiss_IndexIVF_print_stats (const FaissIndexIVF* index);
 
+/// get inverted lists ids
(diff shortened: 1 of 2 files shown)


c_api/IndexIVF_c.h
```diff
@@ -114,6 +114,8 @@ double faiss_IndexIVF_imbalance_factor (const FaissIndexIVF* index);
 /// display some stats about the inverted lists of the index
 void faiss_IndexIVF_print_stats (const FaissIndexIVF* index);
 
+/// get inverted lists ids
```
Enet4: I would prefer to document this function better, as it could make room for misuse. In particular, the user needs to first call `faiss_IndexIVF_get_list_size` to ensure that a large enough array is allocated, and only then call this function.

```suggestion
/// Get the IDs in an inverted list. IDs are written to `invlist`, which must be large enough
//// to accommodate the full list.
///
/// @param list_no the list ID
/// @param invlist output pointer to a slice of memory, at least as long as the list's size
/// @see faiss_IndexIVF_get_list_size(size_t) 
```

PR #1095, opened 2020-01-28
Replace memset on struct with floats

gaurav1086:
memset on struct with floating values does not zero out.

### DIFF
--- a/IndexIVF.cpp
+++ b/IndexIVF.cpp
@@ -889,7 +889,12 @@ IndexIVF::~IndexIVF()
 
 void IndexIVFStats::reset()
 {
-    memset ((void*)this, 0, sizeof (*this));
+    nq = 0;
+    nlist = 0;
+    ndis = 0;
+    nheap_updates = 0;
+    quantization_time = 0.0;
+    search_time = 0.0;
 }
 
 


mdouze:
Why do you think this is useful?

gaurav1086:
@mdouze , thanks a lot for the review. The memset approach is incorrect for structs or arrays of floating point values. If the initial intent here was to have them zero'd, then the new approach is better or else we can get rid of the memset() [O(n)] call altogether and just use a boolean value (within the struct) to indicate whether the entire memory of the struct is reset or not.

mdouze:
there are no structs or arrays here.

gaurav1086:
@mdouze , IndexIVFStats is a struct I am talking about.

Enet4:
I tried looking up on the subject of using `memset` to initialize a POD struct, and I'm honestly still a bit confused. [This SO question](https://stackoverflow.com/q/2773927/1233251) predates C++11, whereas [this one](https://stackoverflow.com/q/23123606/1233251) suggests that a POD type must not have a non-trivial default constructor (to the best of my knowledge calling `reset` makes the current default constructor non-trivial). While this operation should be safe in practice, given that the the struct is a plain composition of `size_t` and `double`, it is also known not to be a very recommended way to initialize a value. I am not entirely sure of what the standard mandates here, but if it is undefined behavior to manipulate `this` in such a fashion with this type, what we could be observing so far is a coincidence, favored by circumstances and assumptions, that the outcome of calling `memset` is the same as initializing each member manually.

All in all, it might not be important to reflect so deeply on this one. Assigning members one by one with 0 is as safe and intuitive as it can be, but if no one has detected runtime issues with the current implementation, I understand if y'all prefer not to change the code.

mdouze:
Of course it is possible to use memset. 
Discussions about how to do something "properly" in C++ are a waste of time and mental load.

PR #1121, opened 2020-02-29
Add .cuh files to install include dir. 

wxingda:
Add `.cuh` files to install include dir ,then users can use more APIs and make CUDA testing.


wickedfoo:
Looks good to me.

Issue #1139, opened 2020-03-12
v1.6.2 : Compilation of C API Failed


glutamatt:
# Summary

the C api compilation failed on `Clustering_c.cpp` compilation

Faiss version: [v1.6.2](https://github.com/facebookresearch/faiss/releases/tag/v1.6.2)

# Reproduction instructions

```Dockerfile
FROM ubuntu:18.04 as builder

ARG FAISS_HOST=https://github.com/facebookresearch/faiss
ARG FAISS_VERSION=1.6.2

RUN apt-get update && apt-get -y install gcc build-essential libblas-dev liblapack-dev
RUN apt-get install -y wget

RUN wget -nv -O faiss.tar.gz ${FAISS_HOST}/archive/v${FAISS_VERSION}.tar.gz
RUN tar -zxf faiss.tar.gz
WORKDIR /faiss-${FAISS_VERSION}
RUN ./configure --without-cuda && make
WORKDIR /faiss-${FAISS_VERSION}/c_api
RUN make
```

```
g++ -std=c++11 -DFINTEGER=int  -fopenmp  -fPIC -m64 -Wno-sign-compare -g -O3 -Wall -Wextra -I.. -I ../impl -DNDEBUG  -mpopcnt -msse4 -c error_impl.cpp -o error_impl.o
g++ -std=c++11 -DFINTEGER=int  -fopenmp  -fPIC -m64 -Wno-sign-compare -g -O3 -Wall -Wextra -I.. -I ../impl -DNDEBUG  -mpopcnt -msse4 -c Index_c.cpp -o Index_c.o
g++ -std=c++11 -DFINTEGER=int  -fopenmp  -fPIC -m64 -Wno-sign-compare -g -O3 -Wall -Wextra -I.. -I ../impl -DNDEBUG  -mpopcnt -msse4 -c IndexFlat_c.cpp -o IndexFlat_c.o
g++ -std=c++11 -DFINTEGER=int  -fopenmp  -fPIC -m64 -Wno-sign-compare -g -O3 -Wall -Wextra -I.. -I ../impl -DNDEBUG  -mpopcnt -msse4 -c Clustering_c.cpp -o Clustering_c.o
Clustering_c.cpp: In function 'void faiss_Clustering_obj(FaissClustering*, float**, size_t*)':
Clustering_c.cpp:85:72: error: 'struct faiss::Clustering' has no member named 'obj'
[...]

PR #1141, opened 2020-03-13
CI : job with allowed failure to build C api

glutamatt:
following of https://github.com/facebookresearch/faiss/issues/1139#issuecomment-598350870

I duplicated the first job, adding a env var to init C api compilation
I set this duplicated job as `allow_failures`

image

***

## existing job
image

***

## new job with `allow_failures`
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ env:
 
 matrix:
   include:
-    - os: linux
+    - &base_job
+      os: linux
       compiler: gcc
       addons:
         apt:
@@ -17,6 +18,10 @@ matrix:
 #            - swig3.0
       env:
         - PYTHON_CFLAGS="-I/usr/include/python2.7"
+    - <<: *base_job
+      env:
+        - PYTHON_CFLAGS="-I/usr/include/python2.7"
+        - BUILD_C_API="true"
     - os: linux
       compiler: gcc
       addons:
@@ -86,6 +91,9 @@ matrix:
           update: true
       env:
         - MATRIX_EVAL="which gcc; export CC=/usr/local/bin/gcc CXX=/usr/local/bin/g++"
+    - env:
+        - PYTHON_CFLAGS="-I/usr/include/python2.7"
+        - BUILD_C_API="true"
 
 before_install:
   - eval "$MATRIX_EVAL"
@@ -98,6 +106,7 @@ install:
   - ./configure --without-cuda
   - make
   - make -C python
+  - ./.travis/install_c.sh
 
 script:
   - make test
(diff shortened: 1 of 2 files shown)


glutamatt:
I had to change and move the condition test in the bash script in order to make it work on darwin environements ( if you compare the actual files diff and the screenshots )

mdouze:
@beauby any opinion on this PR?

beauby:
This would be great. As we moved to CircleCI, the PR would need to be reworked, but it shouldn't take much effort, and the C API build could be a dependent job of the C++ tests.

glutamatt:
It seems that CircleCI doesn't support the "allow_failures" feature : after some search in the circleci forum, there is plenty feature requests and some ugly workarounds.

Do I
 - make the success of C api build mandatory (no failure allowed) ?
 - I close this PR ?
 - I try some ugly workaround ?

glutamatt:
to sum up and check whether i got the context : 
 
**before, with travis**, macOS build success was allowed to fail : i added C api build test optional as well
 travis need the "optional" flag in order the keep the whole build successful even if optional jobs fail

**now with CircleCi**, each job seems independent (add its own github check), and a PR can be merged without all checks pass

correct me if i'm wrong

beauby:
@glutamatt This is correct.

glutamatt:
On master branch ([this commit](https://github.com/facebookresearch/faiss/commit/894116067888089c269f6668f539a4d98f72838e)), the build of `c_api` seems broken 💥 

I achieved to build the faiss library based of commands defined for [circleci](https://github.com/facebookresearch/faiss/blob/894116067888089c269f6668f539a4d98f72838e/.circleci/config.yml#L12-L41)
but then, following [documentation for c_api build](https://github.com/facebookresearch/faiss/blob/894116067888089c269f6668f539a4d98f72838e/c_api/INSTALL.md) does not work anymore

```
root@3005e5307bd1:~/faiss-master/c_api# make
Makefile:10: ../makefile.inc: No such file or directory
make: *** No rule to make target '../makefile.inc'.  Stop.
```

here we have an other example of the need of c_api build check
... maybe something related to the new usage of `cmake`

Enet4:
Yes, this is very likely due to the recent changes in the building system to use CMake instead of GNU autotools.

While I would be interested in migrating the C API to also use this building system, it is something that I am not that familiar with, and for which I cannot define a deadline. Assistance in this process would be greatly appreciated.

glutamatt:
@beauby , please,  your help is needed to get the C api build back on track
:pray:  thanks

glutamatt:
https://github.com/facebookresearch/faiss/pull/1678 seems to supersedes :+1:

PR #1160, opened 2020-03-25
Improve CPU architecture tolerance for CUDA build

hartb:
Depending on the host architecture type and the CUDA install
packaging type (e.g. .deb, .rpm, .run), CUDA include files can
end up in a couple different locations. Expand the search for
CUDA headers to add top-level 'include/' directory in the CUDA
root. And (on linux) adjust the target-specific search location
to match the build machine architecture.


beauby:
Thanks for the fix. As we're now relying on CMake to locate CUDA, this shouldn't be needed anymore, but feel free to reopen if needed.

PR #1243, opened 2020-06-03
fix free parameterspace

glutamatt:
ParameterSpace destructor implementation exists but declaration in header file is missing

https://github.com/facebookresearch/faiss/blob/v1.6.3/c_api/AutoTune_c.cpp#L37

it ends up in user code with error
```
faiss/writer/index.go:65:9: could not determine kind of name for C.faiss_ParameterSpace_free
```

the fix resolve this issue


mdouze:
Assuming this is ok @Enet4 ?

PR #1270, opened 2020-06-28
fix datatypes for mac

mdouze:
There are leftover longs in the tutorial codes.
long is 32 bit on mac (and on windows), so convert this to idx_t.


gpostelnicu:
Are these files built as part of the continuous build? If not, including them could help uncover these issues.

mdouze:
> Are these files built as part of the continuous build? If not, including them could help uncover these issues.

no. Perhaps a good idea to trigger compiles of the files (even if we don't test them). @beauby ?

PR #1271, opened 2020-06-29
python makefile should respect configured prefix

bhatiaharsh:
Python bindings should be installed in the prefix provided by the config, since the user may want to customize the installation.


beauby:
Thanks for the fix. As the build system evolved, this shouldn't be needed anymore. Please feel free to reopen if needed.

PR #1288, opened 2020-07-15
Properly handle depend file for CPU/CUDA build switch

kyamagu:
The current `Makefile` ignores `NVCC` variable and always generate `depend` file with CUDA objects.
Although this still passes the build process, results in tons of compiler errors in stdout. This PR fixes this issue.


beauby:
Thanks for the fix. As we moved to CMake for the build, this shouldn't be needed anymore. Please feel free to reopen if needed.

PR #1296, opened 2020-07-18
[C API] Fix GPU library build process

Enet4:
The changes in the building process of Faiss v1.5.1 affected the C API building process, which currently still relies on separate libfaiss and libgpufaiss. This was probably overlooked due to the lack of CUDA in CI.

I would be interested in improving the makefiles for the C API eventually, but for now this PR will make building the full C library work again.

### DIFF
--- a/c_api/example_c.c
+++ b/c_api/example_c.c
@@ -13,6 +13,7 @@
 #include <time.h>
 
 #include "error_c.h"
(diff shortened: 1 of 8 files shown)


mdouze:
It seems that we are moving to cmake as the build system because it's easier for Windows. @beauby ?

PR #1308, opened 2020-07-31
Facebook sync 2020-07-31

beauby:


Index.h
```diff
@@ -17,8 +17,8 @@
 #include <sstream>
 
 #define FAISS_VERSION_MAJOR 1
-#define FAISS_VERSION_MINOR 6
-#define FAISS_VERSION_PATCH 3
+#define FAISS_VERSION_MINOR 4
```
mdouze, line 20: what is the new version? 1.4.0?

IndexHNSW.cpp
```diff
@@ -289,29 +289,30 @@ void IndexHNSW::search (idx_t n, const float *x, idx_t k,
...
-#pragma omp parallel reduction(+ : nreorder)
+#pragma omp parallel
         {
             VisitedTable vt (ntotal);
 
             DistanceComputer *dis = storage_distance_computer(storage);
             ScopeDeleter1<DistanceComputer> del(dis);
 
-#pragma omp for
+#pragma omp declare reduction (stats_combine : HNSWStats : omp_out.combine(omp_in))
```
mdouze, line 307: not supported for openmp < 4. 
If it is not too much work, I'd suggest doing it with reduction + 's

benchs/bench_all_ivf/bench_all_ivf.py
```diff
@@ -1,10 +1,10 @@
+#!/usr/bin/env python2
```
mdouze, line 1: python3, maybe do a pass of search/replace. Most files are compatible with both, but python3 should be the default.

IndexHNSW.cpp
beauby: This PR is just a sync, so we shouldn't introduce diverging changes here.

benchs/bench_all_ivf/bench_all_ivf.py
beauby: As above, we can do this in a subsequent PR.

PR #1312, opened 2020-08-03
Move source code to faiss/ subdirectory.

beauby:


LowikC:
Is the CI failure expected?

beauby:
The CI failure is expected at this point: one test is failing because it is using a feature introduced in python 3.2 (but we'll move the CI to python 3 soon), and the other one is failing because there is no build support for the `faiss_contrib` module yet (but will be introduced soon).

PR #1320, opened 2020-08-07
fixed. compile error finite undefined

imlidian:
```
faiss/Clustering.cpp:291:37: error: use of undeclared identifier 'finite'; did you mean 'isfinite'?
            FAISS_THROW_IF_NOT_MSG (finite (x[i]),
                                    ^~~~~~
                                    isfinite
```

PR #1322, opened 2020-08-08
Fix demos compile failed

imlidian:
```
demo_imi_pq.cpp:138:33: error: cannot initialize a parameter of type 'const faiss::Index::idx_t *' (aka 'const long long *')
      with an rvalue of type 'std::__1::vector<long, std::__1::allocator<long> >::value_type *' (aka 'long *')
                                ids.data() + begin);
```

https://github.com/facebookresearch/faiss/pull/1320

### DIFF
--- a/demos/demo_imi_pq.cpp
+++ b/demos/demo_imi_pq.cpp
@@ -121,7 +121,7 @@ int main ()
                 elapsed() - t0, nb);
 
         std::vector <float> database (nb * d);
-        std::vector <long> ids (nb);
+        std::vector <int64_t> ids (nb);
         for (size_t i = 0; i < nb; i++) {
             for (size_t j = 0; j < d; j++) {
                 database[i * d + j] = drand48();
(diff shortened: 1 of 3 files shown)


mdouze:
Maybe hold on, we are in the process of moving the build chain to cmake.

demos/demo_imi_pq.cpp
```diff
@@ -121,7 +121,7 @@ int main ()
                 elapsed() - t0, nb);
 
         std::vector <float> database (nb * d);
-        std::vector <long> ids (nb);
+        std::vector <int64_t> ids (nb);
```
beauby, line 124: This should technically be `faiss::Index::idx_t`.

imlidian: thins for advice

beauby:
@imlidian  Please rebase and we can merge this.

mdouze:
This was probably fixex.

PR #1386, opened 2020-08-30
Bundle all files from contrib/ in python extension.

beauby:


mdouze:
All checks pass, this diff will be landed through a phabricator update.

PR #1397, opened 2020-09-09
Add IVF statistics info for search_with_parameters

JinHai-CN:
Signed-off-by: JinHai-CN <hai.jin@zilliz.com>

No, IVF completed statistics information when we use search_with_parameters in IVFlib.cpp to search the vector.

### DIFF
--- a/faiss/IVFlib.cpp
+++ b/faiss/IVFlib.cpp
@@ -335,8 +335,11 @@ void search_with_parameters (const Index *index,
     const IndexIVF *index_ivf = dynamic_cast<const IndexIVF *>(index);
     FAISS_THROW_IF_NOT (index_ivf);
 
+    double t0 = getmillisecs();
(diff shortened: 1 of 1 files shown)


JinHai-CN:
@mdouze 
'83 - TestGpuIndexIVFPQ.Float16Coarse ' is failed.
Do we need to run the CI again?

mdouze:
It would be better if the timing statistics were returned by the function, like `nb_dis_ptr`. The rationale is that this function is intended for use from multiple threads, where accessing the global var would cause thread conflicts.

JinHai-CN:
1. IVF indexes already use OpenMP in the search function. Will the search function really be used in multiple threads cases?

2. From the code structure. Yes, it will be better to avoid using global shared variables.

mdouze:
I integrated this modification to https://github.com/facebookresearch/faiss/pull/1438

PR #1410, opened 2020-09-15
fix ivflib

adream307:
if `ils = dynamic_cast<ArrayInvertedLists *> (index_ivf->invlists)` failed, `ils` would be `nullptr`.
so check if `ils` is `nullptr` before use it.


mdouze:
Looks good.

PR #1443, opened 2020-10-05
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.


mdouze:
Sounds good! 
TBH I thought we fixed all the python2 prints in the code, but these were apparently missing.
Thanks.

PR #1459, opened 2020-10-12
Mark ~BufferedIOWriter() noexcept(false)

beauby:
Summary: In C++11, destructors default to `noexcept(true)`. This destructor can throw (through `FAISS_THROW_IF_NOT()`), so marking it accordingly.

Differential Revision: D24253879

### DIFF
--- a/faiss/impl/io.h
+++ b/faiss/impl/io.h
@@ -50,7 +50,7 @@ struct IOWriter {
     // return a file number that can be memory-mapped
     virtual int fileno ();
 
-    virtual ~IOWriter() {}
+    virtual ~IOWriter() noexcept(false) {}
 };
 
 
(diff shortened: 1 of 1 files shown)


Enet4:
Is this a good idea, though? An exception thrown in a destructor can still have severe implications to a running program. Usually, one would prefer handling that exception in the destructor, and expose a `close` never function if necessary.

https://stackoverflow.com/q/130117

beauby:
@Enet4 Agreed, however this PR mainly aims to make the current situation explicit in the meantime. @mdouze what's your opinion?

mdouze:
I agree with both of you. I should add a `close()` or `flush()` function to the API. But in the meantime, let's make the exception throw explicit.

PR #1466, opened 2020-10-14
Skip TestShardedFlat for 1 GPU fixes #982

### LINKED ISSUE #982: make test_gpu: TestShardedFlat fails for 1 GPU
# Summary

`make test_gpu` fails if I have 1 GPU due to the TestShardedFlat test case where it asserts that the number of GPUs should be greater than 1. 

Does Faiss require more than 1 GPU to function as expected?

OS: Ubuntu 16.04

Faiss version: 31393764cccb3ae94e8a604e1fd891b331c46221

Faiss compilation options: 

Running on:
- [ ] CPU
- [X] GPU

Interface: 
- [ ] C++
- [X] Python

# Reproduction instructions

`make test_gpu` where you have 1 GPU. Line 224, `assert faiss.get_num_gpus() > 1` in `test_gpu_index.py` of course fails.

eyaltrabelsi:


beauby:
Thanks for the PR. Would you mind using `@unittest.skipIf()`instead?

eyaltrabelsi:
@beauby Done :)

faiss/gpu/test/test_gpu_index.py
```diff
@@ -219,11 +219,9 @@ def test_stress(self):
 
 
 class TestShardedFlat(unittest.TestCase):
-
+    
+    unittest.skipIf(faiss.get_num_gpus() == 1,"relevant for multiple GPU only")
```
beauby, line 223: Since `unittest.skipIf()` is a decorator, you need to call it as `@unittest.skipIf()`.

eyaltrabelsi: @beauby my bad updated :)

Issue #1503, opened 2020-11-04
Error on running bench_link_and_code.py


Hap-Hugh:
Running on:
- CPU
Interface: 
- Python

I used sift1M to run bench_link_and_code.py with the following commands:
`python bench_link_and_code.py \
   --db sift1M \
   --M0 6 \
   --indexkey OPQ36_144,HNSW32_PQ36 \
   --indexfile indexfile\
   --beta_nsq 4  \
   --beta_centroids /deep100M_PQ36_L6_nsq4.npy \
   --neigh_recons_codes $bdir/deep100M_PQ36_L6_nsq4_codes.npy \
   --k_reorder 0,5 --efSearch 1,1024`

The error is:
raise AttributeError("'%s' object has no attribute '%s'" % (class_type.__name__, name))
AttributeError: 'Clustering' object has no attribute 'obj'

![image](https://user-images.githubusercontent.com/33805236/98104480-77182000-1ed1-11eb-8586-3e35dff43b5b.png)

mdouze:
We forgot to move that script to the new Clustering interface. Will fix. 
In the meantime, you can patch it as in 

https://github.com/facebookresearch/faiss/blob/master/faiss/python/__init__.py#L905

Hap-Hugh:
I just patch this small bug. But I found there is still something wrong about 'get_neighbor_table'. Fortunately, it's easy to fix it.
![image](https://user-images.githubusercontent.com/33805236/98202912-007d3000-1f6e-11eb-8c24-16cecf57f74f.png)
To avoid this bug, I use `int(i)` to force the type of this variable to be 'int' instead of `'numpy.int64'`

mdouze:
Please submit a pull request with your changes when it works.

Hap-Hugh:
Signed on CLA and Submit the code. Please have a check on it.

Best

PR #1510, opened 2020-11-05
Fix Bugs in Link&Code

Hap-Hugh:
As the issue said, I patched these two bugs and the codes are working well now.

https://github.com/facebookresearch/faiss/issues/1503#issuecomment-722172257


mdouze:
Looks good.

PR #1600, opened 2020-12-30
make AVX2-detection platform-independent

h-vetinari:
In the context of https://github.com/conda-forge/faiss-split-feedstock/issues/23, I discussed with some of the conda-folks how we should support AVX2 (and potentially other builds) for faiss. In the meantime, we'd like to follow the model that faiss itself is using (i.e. build with AVX2 and without and then load the corresponding library at runtime depending on CPU capabilities).

Since windows support for this is missing (and the other stuff is also non-portable in `loader.py`), I chased down `numpy.distutils.cpuinfo`, which is pretty outdated, and opened: https://github.com/numpy/numpy/issues/18058

While the [private API](https://github.com/numpy/numpy/issues/18058#issuecomment-749852711) is obviously something that _could_ change at any time, I still think it's better than platform-dependent shenanigans.

Opening this here to ideally upstream this right away, rather than carrying patches in the conda-forge feedstock.

TODO:
* [ ] adapt conda recipe for windows in this repo to also build avx2 version

### DIFF
--- a/faiss/python/loader.py
+++ b/faiss/python/loader.py
@@ -11,42 +11,42 @@ import logging
 
 def instruction_set():
     """
-    Returns a dictionary for supported instruction sets, see
+    Returns the set of supported CPU features, see
     https://github.com/numpy/numpy/blob/master/numpy/core/src/common/npy_cpu_features.h
-    for the list of features that this dictionary contains per architecture.
+    for the list of features that this set may contain per architecture.
 
     Example:
     >>> instruction_set()  # for x86
-    {"SSE2": True, "AVX2": False, ...}
+    {"SSE2", "AVX2", ...}
     >>> instruction_set()  # for PPC
-    {"VSX": True, "VSX2": False, ...}
+    {"VSX", "VSX2", ...}
     >>> instruction_set()  # for ARM
-    {"NEON": True, "ASIMD": False, ...}
+    {"NEON", "ASIMD", ...}
     """
     import numpy
     if LooseVersion(numpy.__version__) >= "1.19":
         # use private API as next-best thing until numpy/numpy#18058 is solved
         from numpy.core._multiarray_umath import __cpu_features__
-        return __cpu_features__
+        # __cpu_features__ is a dictionary with CPU features
+        # as keys, and True / False as values
+        supported = {k for k, v in __cpu_features__.items() if v}
+        return supported
 
     # platform-dependent legacy fallback before numpy 1.19, no windows
     if platform.system() == "Darwin":
         if subprocess.check_output(["/usr/sbin/sysctl", "hw.optional.avx2_0"])[-1] == '1':
-            return {"AVX2": True}
+            return {"AVX2"}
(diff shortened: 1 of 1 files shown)


h-vetinari:
@wickedfoo @mdouze 
I'd like to build avx2 for windows too in conda-forge (see OP), and having your approval on this approach would make me more confident that I'm not going against your ideas somehow. PTAL

mdouze:
That would be great! Adding @beauby for the status of AVX2 on Windows (not sure if it is tested at all). 
NB that in the Faiss Circle CI we do not test AVX2 because not all CI machines support it.
This means that a lot of codepaths are not exercised. For AVX2 on Linux we rely on the FB internal tests.

h-vetinari:
> This means that a lot of codepaths are not exercised.

I was under the impression that it's just a compiler flag that's changing between building for avx2 or not...?

faiss/python/loader.py
```diff
@@ -3,30 +3,39 @@
...
+            return False
+    return False
 
 
 logger = logging.getLogger(__name__)
 
 try:
-    instr_set = instruction_set()
-    if instr_set == "AVX2":
+    has_AVX2 = supports_AVX2()
```
mdouze, line 37: I would like to keep the `instruction_set` name here generic since Faiss also compiles on ARM and PPC.

h-vetinari:
Responded to the feedback @mdouze, PTAL :)

faiss/python/loader.py
```diff
@@ -3,30 +3,48 @@
...
+    for the list of features that this dictionary contains per architecture.
+
+    Example:
+    >>> instruction_set()  # for x86
+    {"SSE2": True, "AVX2": False, ...}
+    >>> instruction_set()  # for PPC
+    {"VSX": True, "VSX2": False, ...}
+    >>> instruction_set()  # for ARM
+    {"NEON": True, "ASIMD": False, ...}
+    """
```
mdouze, line 25: I know this is constrained on the numpy side but what about returning a set with the flags set to True? 

For example, below the has_AVX2 check will fail on ARM with `KeyError`.

mdouze (commented):
small comment.

faiss/python/loader.py
h-vetinari: Good point about the key error. I think the cleanest fix would just be to use `.get("AVX2", False)` on the output, which is fine even if some keys are not there.

Alternatively, we could "pad" the dictionary with the keys from the other architectures with value `False`. I don't like this idea though, because we'd be duplicating the list of instruction sets compared to numpy

beauby: We could also return a set in every case (in the case of numpy, selecting the keys for which the value is `True`).

h-vetinari: Sure, that's a good idea as well.

mdouze: Thanks for using the set. Will get it through the FB internal checks.

PR #1603, opened 2021-01-05
Add sanity check and fix some typos

JinHai-CN:
Signed-off-by: JinHai-CN <hai.jin@zilliz.com>

### DIFF
--- a/faiss/IVFlib.cpp
+++ b/faiss/IVFlib.cpp
@@ -60,23 +60,21 @@ void check_compatible_for_merge (const Index * index0,
 
 const IndexIVF * try_extract_index_ivf (const Index * index)
 {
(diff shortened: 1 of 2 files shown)


faiss/IVFlib.cpp
```diff
@@ -60,23 +60,19 @@ void check_compatible_for_merge (const Index * index0,
 
 const IndexIVF * try_extract_index_ivf (const Index * index)
 {
+    if(index == nullptr) return nullptr;
     if (auto *pt =
         dynamic_cast<const IndexPreTransform *>(index)) {
         index = pt->index;
-    }
```
mdouze, line 66: This is on purpose so that a nested indexpretransform and an idmap can be extracted correctly.

Issue #1604, opened 2021-01-05
fail to run cmp_with_scann.py


cynricshu:
# Summary

I want to reproduce the benchmark results in https://github.com/facebookresearch/faiss/wiki/Indexing-1M-vectors

Running on:
- [X] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [X] Python

# Reproduction instructions

cd benchs/bench_all_ivf

python3 cmp_with_scann.py --db sift1M --base_dir /home/app/ann-data/results/cmp_ivf_scan_2

# log

- **just look at the last line**

args: Namespace(base_dir='/home/app/ann-data/results/cmp_ivf_scan_2', db='sift1M', download=False, k=10, lib='faiss', measure='1-recall', nprobe='1,2,5,10,20,50,100,200', nrun=5, pre_reorder_k='0,10,100,1000', thenscann=False)
nb processors 16
model name	: Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
path: /usr/local/lib/python3.6/dist-packages/faiss-1.6.5-py3.6.egg/faiss/contrib/datasets.py
load data sift1M
dataset in dimension 128, with metric L2, size: Q 10000 B 1000000 T 100000
use precomputed table= 1 by residual= False
adding a refine index
Traceback (most recent call last):
  File "cmp_with_scann.py", line 308, in <module>
    main()
  File "cmp_with_scann.py", line 114, in main
    nrun, args.measure
  File "cmp_with_scann.py", line 273, in faiss_eval_search
    index_refine = faiss.IndexRefineFlat(index, faiss.swig_ptr(xb))
  File "/usr/local/lib/python3.6/dist-packages/faiss-1.6.5-py3.6.egg/faiss/__init__.py", line 450, in replacement_init_multiple
    original_init(self, *args)
  File "/usr/local/lib/python3.6/dist-packages/faiss-1.6.5-py3.6.egg/faiss/swigfaiss.py", line 2633, in __init__
    this = _swigfaiss.new_IndexRefineFlat(*args)
RuntimeError: Error in faiss::IndexRefine::IndexRefine(faiss::Index*, faiss::Index*) at /home/app/lib-faiss/faiss/IndexRefine.cpp:35: Error: 'base_index->ntotal == refine_index->ntotal' failed

mdouze:
I can repro

PR #1607, opened 2021-01-06
Fix refine implementation

mdouze:
The IndexRefineFlat with pre-populated indexes could not be used because of the order of construction of the parent class. This diff fixes is. This addresses #1604.

### DIFF
--- a/benchs/bench_all_ivf/cmp_with_scann.py
+++ b/benchs/bench_all_ivf/cmp_with_scann.py
@@ -270,6 +270,8 @@ def faiss_eval_search(
           "by residual=", index.by_residual)
 
     print("adding a refine index")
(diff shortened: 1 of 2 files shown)


benchs/bench_all_ivf/cmp_with_scann.py
```diff
@@ -270,6 +270,8 @@ def faiss_eval_search(
           "by residual=", index.by_residual)
 
     print("adding a refine index")
+    flat_index = faiss.IndexFlat(index.d, index.metric_type)
+    flat_index.add(xb)
```
cynricshu, line 274: where did you use the variable flat_index ?

mdouze: ah right this should be removed

PR #1666, opened 2021-02-03
Add missing headers in faiss/[gpu/]CMakeLists.txt

h-vetinari:
While preparing https://github.com/conda-forge/faiss-split-feedstock/pull/26, I grepped for the expected headers based on the files in the repo, à la: 
```
>ls faiss/invlists/ | grep -E "h$"
BlockInvertedLists.h
DirectMap.h
InvertedLists.h
InvertedListsIOHook.h
OnDiskInvertedLists.h
```

Doing so uncovered that there were some headers missing (apparently) in `CMakeLists.txt`, namely:
```
faiss/impl/ResultHandler.h
faiss/gpu/impl/IVFInterleaved.cuh
faiss/gpu/impl/InterleavedCodes.h
faiss/gpu/utils/WarpPackedBits.cuh
```

It's possible that they were left out intentionally, but I didn't see something that would make me think so e.g. in [`ResultHandler.h`](https://github.com/facebookresearch/faiss/blob/master/faiss/impl/ResultHandler.h).

While I was at it, I decided to order the filenames consistently (alphabetically, except for the increasing bit-sizes for blockselect/warpselect, as is already the case for `impl/scan/IVFInterleaved<x>.cu`), but of course, those commits could easily be dropped.

By reviewing the commits separately, it should be clear (for the first two) from the equal number of deletions/insertions (and the simple diff) that this is just a reshuffle. The only additions are in the last commit.

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -110,6 +110,7 @@ set(FAISS_HEADERS
   impl/PolysemousTraining.h
   impl/ProductQuantizer-inl.h
   impl/ProductQuantizer.h
+  impl/ResultHandler.h
   impl/ScalarQuantizer.h
   impl/ThreadedIndex-inl.h
   impl/ThreadedIndex.h
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -21,18 +21,18 @@ add_library(faiss
   IndexIVF.cpp
   IndexIVFFlat.cpp
   IndexIVFPQ.cpp
-  IndexIVFPQR.cpp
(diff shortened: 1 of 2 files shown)


h-vetinari:
@beauby @mdouze @wickedfoo 
Would be glad to have a quick 👍 / 👎 on this, so I can move ahead with 1.7.0 one way or another (as it currently blocks 3 PRs).

Sidenote: the build times basically doubled across the board from 1.6.5 to 1.7.0 - not sure if you're seeing similar increases here, but just FYI.

mdouze:
LGTM, although I am not the cmake expert here. Will import it to the FB review system.

PR #1671, opened 2021-02-05
Parameterize CUDA_ARCHS in packaging jobs.

beauby:
This will allow us to support compute capabilities 8.0 and 8.6 (for
Ampere devices) with CUDA 11.


mdouze:
Please import to phabricator.

Issue #1010, opened 2019-10-31
Internal Library crash when searching with small numbers of vectors.


triclops200:
# Summary

# Platform

OS: Ubuntu Linux 19.04

Faiss version: 8bc09261f4f4cf66371e14f8f9cef0b4a904e84a
Faiss compilation options: no-cuda
Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [x] C++
- [ ] Python

# Reproduction instructions

I'm creating a IndexIVFFlat at 512 dimensions with 2048 probe points. I'm inserting 2k vectors then doing a search and the code crashes. This is the resulting error

```IndexPQ.cpp:861: void faiss::{anonymous}::MinSumK<T, SSA, use_seen>::run(const T*, int64_t, T*, int64_t*) [with T = float; SSA = faiss::{anonymous}::SemiSortedArray<float>; bool use_seen = false; int64_t = long int]: Assertion `heap_size > 0' failed.```

mdouze:
So that's an `IndexIVFFlat` with `MultiIndexQuantizer` quantizer. 
That's an interesting bug, which is probably due to the data distribution. 
Would it be possible for you to share the code + actual vectors this bug shows up?

triclops200:
Sure, take your demo_imi_flat.cpp and change nb to 1000 and you'll encounter the same error.

triclops200:
Any updates?

mdouze:
Yes I could repro the issue. Marking this as an enhancement.

mdouze:
Should be fixed by #1673

PR #1673, opened 2021-02-07
Handle the situation where nprobe > nlist in IndexIVF

KinglittleQ:
## Description

Fix the bug mentioned in https://github.com/facebookresearch/faiss/issues/1010. When `nprobe` is greater than `nlist` in `IndexIVF`, the program will crash because the index will ask the quantizer to return more centroids than it owns.

## Changes:
1. Set `nprobe` as `nlist` if it is greater than `nlist` during searching.
2. Add one test to detect this bug.
3. Fix typo in `IndexPQ.cpp`.

### DIFF
--- a/faiss/IndexIVF.cpp
+++ b/faiss/IndexIVF.cpp
@@ -277,7 +277,7 @@ void IndexIVF::set_direct_map_type (DirectMap::Type type)
 void IndexIVF::search (idx_t n, const float *x, idx_t k,
                          float *distances, idx_t *labels) const
 {
-    size_t nprobe = std::min(nlist, this->nprobe);
+    const size_t nprobe = std::min(nlist, this->nprobe);
 
     // search function for a subset of queries
     auto sub_search_func = [this, k, nprobe]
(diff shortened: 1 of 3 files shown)


faiss/IndexIVF.cpp
```diff
@@ -277,10 +277,10 @@ void IndexIVF::set_direct_map_type (DirectMap::Type type)
 void IndexIVF::search (idx_t n, const float *x, idx_t k,
                          float *distances, idx_t *labels) const
 {
-
+    size_t nprobe = std::min(nlist, this->nprobe);
 
     // search function for a subset of queries
-    auto sub_search_func = [this, k]
+    auto sub_search_func = [this, k, nprobe]
```
mdouze, line 283: this is useless

faiss/IndexIVF.cpp
```diff
@@ -352,6 +352,8 @@ void IndexIVF::search_preassigned (idx_t n, const float *x, idx_t k,
                                    IndexIVFStats *ivf_stats) const
 {
     long nprobe = params ? params->nprobe : this->nprobe;
+    nprobe = std::min((long)nlist, nprobe);
```
mdouze, line 355: looks good. While you are at it, could you replace occurrences of `long` with `idx_t` in this function ? We are trying to get rid of long, as its size is not portable.

mdouze (commented):
Please address comments.

KinglittleQ:
```
size_t nprobe = std::min(nlist, this->nprobe);
auto sub_search_func = [this, k, nprobe]
```
Here `nprobe` is a local variable instead of a member variable So it needs to be captured by the lambda function.

KinglittleQ:
I found that calling `range_search` method would lead to a crash when `nprobe > nlist`. So I apply the same modification to `range_search` method.

New Changes:
1. The `range_search` method is also fixed.
2. Replace `long` with `idx_t` in `IndexIVF.cpp`.
3. Add a test to test `range_search`.

PR #1390, opened 2020-09-02
Support C api compilation with cmake

mdouze:
This is an attempt to compile the C api with CMake. 

Current status is that the library (.a) compiles but the executable does not link.

```

make -C build VERBOSE=1

...

cd /private/home/matthijs/faiss_versions/faiss_c_interface/build/c_api && /private/home/matthijs/faiss_versions/faiss_20200817/current_cmake/bin/cmake -E cmake_link_script CMakeFiles/example_c.dir/link.txt --verbose=1
/usr/bin/c++ -O3 -DNDEBUG -o example_c  libfaiss_c.a
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/Scrt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
````

PR #1678, opened 2021-02-11
Added C API to CMake and CircleCI

prabhat00155:
This supersedes #1390 and #1141.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,10 @@ jobs:
             pytest --junitxml=test-results/pytest/results-torch.xml tests/torch_*.py
       - store_test_results:
           path: test-results
+      - run:
+          name: Build C API
+          command: |
+            make -C build -j faiss_c
 
   build_windows:
     executor:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       - run:
           name: Build faiss library
           command: |
-            cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=ON \
+            cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=OFF \
                   -DFAISS_ENABLE_C_API=ON \
                   -DCMAKE_BUILD_TYPE=Release -DBLA_VENDOR=Intel10_64_dyn .
             make -C build -j faiss
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,13 @@ jobs:
             pytest --junitxml=test-results/pytest/results-torch.xml tests/torch_*.py
       - store_test_results:
           path: test-results
+      - run:
+          name: Build C API
+          command: |
+            cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=OFF \
+                  -DFAISS_ENABLE_C_API=ON
+                  -DCMAKE_BUILD_TYPE=Release -DBLA_VENDOR=Intel10_64_dyn .
+            make -C build -j faiss
 
   build_osx:
     macos:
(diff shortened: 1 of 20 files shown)


.circleci/config.yml
```diff
@@ -39,6 +39,13 @@ jobs:
             pytest --junitxml=test-results/pytest/results-torch.xml tests/torch_*.py
       - store_test_results:
           path: test-results
+      - run:
+          name: Build C API
+          command: |
+            cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=OFF \
```
beauby, line 45: Let's set `FAISS_ENABLE_C_API=ON` on the first `cmake` call (which should be the only `cmake` call for this job). We can then do `make -C build -j faiss_c` in the last step.

c_api/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
+
+project(faiss_c_library LANGUAGES C CXX)
+
+set(CMAKE_C_STANDARD 11)
+
+file(GLOB SOURCES "${CMAKE_SOURCE_DIR}/c_api/*.cpp")
```
beauby, line 12: Please do not use `file(GLOB)` for sources as adding or removing a source file will not be caught by the generated build files. Instead, list the files manually in a variable (`set(FAISS_C_SRC ...)`).

c_api/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+add_library(faiss_c STATIC ${SOURCES})
```
beauby, line 13: Please add `EXCLUDE_FROM_ALL` to the library definition so that it is not build by default, but only when explicitly building the `faiss_c` target (cf. https://cmake.org/cmake/help/latest/command/add_library.html#normal-libraries).

c_api/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+target_link_libraries(faiss_c PRIVATE faiss)
+
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/c_api/
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/faiss/c_api/
+        FILES_MATCHING PATTERN "*.h")
+
+add_executable(example_c example_c.c)
```
beauby, line 20: Same here, please `EXCLUDE_FROM_ALL`.

c_api/CMakeLists.txt
beauby: Also, please replace `${CMAKE_SOURCE_DIR}/c_api/` with `${PROJECT_SOURCE_DIR}/`.

beauby (commented):
Looking good, thanks! A few remarks, but once those are addressed we can merge. This PR does not take care of C wrappers for GPU code, but this can be added later.

c_api/CMakeLists.txt
```diff
@@ -9,13 +9,27 @@ project(faiss_c_library LANGUAGES C CXX)
...
+  IndexPreTransform_c.cpp
+  IndexShards_c.cpp
+  Index_c.cpp
+  MetaIndexes_c.cpp
+  clone_index_c.cpp
+  error_impl.cpp
+  index_factory_c.cpp
+  index_io_c.cpp
+)
+add_library(faiss_c STATIC ${FAISS_C_SRC})
```
beauby, line 28: Please remove the `STATIC` here as it prevents building a shared library.

pauloue:
Should `impl/AuxIndexStructures_c.cpp` be built as well?

prabhat00155:
> Should `impl/AuxIndexStructures_c.cpp` be built as well?

I can create a separate PR for `impl` and `gpu`.

beauby:
@prabhat00155 The GPU part is fine in a subsequent PR, but the `impl/` files are actually needed for CPU-only builds.

PR #1680, opened 2021-02-12
add msvc-compatible AVX2 switch in CMakeLists.txt

h-vetinari:
Upstreaming patches from https://github.com/conda-forge/faiss-split-feedstock/pull/27, follow-up (sorta) to #1600.

Not sure if there are more CMake-native tricks to use here, but given that the flags don't have
an equivalent on the MSVC side, I think this approach is reasonable.

Without this patch, we would get:
```
cl : Command line warning D9002: ignoring unknown option '-mavx2'
cl : Command line warning D9002: ignoring unknown option '-mfma'
cl : Command line warning D9002: ignoring unknown option '-mf16c'
cl : Command line warning D9002: ignoring unknown option '-mpopcnt'
```

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -153,6 +153,9 @@ if(FAISS_OPT_LEVEL STREQUAL "avx2")
     target_compile_options(faiss PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mpopcnt>)
   else()
     # MSVC enables FMA with /arch:AVX2; no separate flags for F16C, POPCNT
+    # Ref. FMA (under /arch:AVX2): https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64
+    # Ref. F16C (2nd paragraph): https://walbourn.github.io/directxmath-avx2/
+    # Ref. POPCNT: https://docs.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64
     target_compile_options(faiss PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/arch:AVX2>)
   endif()
   set_target_properties(faiss PROPERTIES OUTPUT_NAME "faiss_avx2")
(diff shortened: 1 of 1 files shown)


faiss/CMakeLists.txt
```diff
@@ -149,7 +149,12 @@ if(NOT WIN32)
 endif()
 
 if(FAISS_OPT_LEVEL STREQUAL "avx2")
-  target_compile_options(faiss PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mpopcnt>)
+  if(NOT WIN32)
+    target_compile_options(faiss PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mpopcnt>)
+  else()
```
beauby, line 154: Nit: Maybe reverse the order (`if(WIN32) ... else()`).

faiss/CMakeLists.txt
```diff
@@ -149,7 +149,12 @@ if(NOT WIN32)
...
+    # MSVC enables FMA with /arch:AVX2; no separate flags for F16C, POPCNT
```
beauby, line 155: If you could add a link to MSVC's docs backing that, that would be awesome.

beauby (commented):
LGTM, two remarks but the changes are good.

faiss/CMakeLists.txt
h-vetinari: I can do it, but I think it's actually more consistent like that:
* most of the CMake-files deal with win/non-win like this (see e.g. a few lines above)
* the unix flags should come first, because they determine the goal, and the windows part argues why it doesn't have all the those flags (so reversing the order would make it harder to read logically, IMO).

faiss/CMakeLists.txt
h-vetinari: The problem is that windows doesn't particularly document the differences to GNU-land, but rather just what it does.

I've collected some references, but most of those insights ultimately came from somewhere on stackoverflow, like [here](https://stackoverflow.com/a/50829580/2965879).

h-vetinari:
Ping @beauby, not much more I can do in terms of documentation (though I raised a ticket for the MS docs).

PR #1681, opened 2021-02-12
Win+AVX2 compat

h-vetinari:
Upstreaming patches from conda-forge/faiss-split-feedstock#27, follow-up (sorta) to #1600.

All these should be fairly uncontroversial, I think (mostly just oversights or stuff that never got
triggered on windows due to not having #1680 so far).

Things work without `#include <faiss/impl/platform_macros.h>`, but I preferred to be explicit here,
because the `__SSEx__` macros (that are used in the affected files) are only defined there for windows.

### DIFF
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -13,6 +13,7 @@
 #include <algorithm>
 
 #include <omp.h>
+#include <faiss/impl/platform_macros.h>
 
 #ifdef __SSE__
 #include <immintrin.h>
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -41,7 +41,7 @@ namespace faiss {
  ********************************************************************/
 
 #ifdef __AVX2__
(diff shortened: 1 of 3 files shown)


faiss/impl/ScalarQuantizer.cpp
```diff
@@ -40,7 +41,7 @@ namespace faiss {
  ********************************************************************/
 
 #ifdef __AVX2__
-#ifdef __F16C__
+#if defined(__F16C__) || defined(_MSC_VER)
```
h-vetinari, line 44: Realizing now that `__F16__` could be set in `platform_macros.h` like `__SSEx__`...

h-vetinari:
Interesting that this manages to provoke a failure... 🤔 

The only change that is even visible on a non-windows system is the `ALIGNED` stuff - not sure what's going on, the macro is defined correctly as far as I can tell.

h-vetinari:
Ah, it seems that this is a flaky test... it also failed in https://github.com/facebookresearch/faiss/pull/1683, which did nothing but add log messages, but it passed in #1680

mdouze:
Looks good to me.

PR #1682, opened 2021-02-12
make setup.py win+avx2 compatible

h-vetinari:
While working on https://github.com/conda-forge/faiss-split-feedstock/pull/27, it turned out I needed
to patch `setup.py` anyway. In order to unify how the extension of the built lib is set, I fell back
to another patch that would/will become necessary if faiss ever wants to support PyPy
(see discussion in https://github.com/conda-forge/faiss-split-feedstock/pull/22).

It would be nice if this was done natively by CMake, but as far as I can tell from
https://gitlab.kitware.com/cmake/cmake/-/issues/21070, cmake is not likely to do that right away.

I didn't particularly expect this patch to be upstreamed (especially if there is no interest for PyPy support, for example),
but @beauby [invited](https://github.com/conda-forge/faiss-split-feedstock/pull/27#issuecomment-777291797)
me to post it so here goes (plus necessary adaptations to the conda recipes)

Related to #1600, #1680, #1681

PS. I thought about using `logger.INFO` in case of an import failure for AVX2, but since it's a setup file,
I thought `print` would actually be more useful. Happy to change or remove if desired.

### DIFF
--- a/.circleci/Dockerfile.faiss_gpu
+++ b/.circleci/Dockerfile.faiss_gpu
@@ -16,8 +16,7 @@ COPY . /faiss
 
 WORKDIR /faiss
 
-RUN export EXT_SUFFIX=$(python3 -c "import sysconfig; print(sysconfig.get_config_var('EXT_SUFFIX'))") && \
-    cmake -B build \
+RUN cmake -B build \
         -DFAISS_ENABLE_GPU=ON \
         -DFAISS_ENABLE_PYTHON=ON \
         -DBUILD_TESTING=ON \
--- a/.circleci/Dockerfile.faiss_gpu
+++ b/.circleci/Dockerfile.faiss_gpu
@@ -16,7 +16,8 @@ COPY . /faiss
 
 WORKDIR /faiss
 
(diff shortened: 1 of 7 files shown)


h-vetinari:
The failing test seems to [flaky](https://github.com/facebookresearch/faiss/pull/1681#issuecomment-778208995)

beauby:
For this, I'd be more inclined to keep the CMake config as is, and just deal with it in `setup.py`. It's not a perfect solution, but it's better than putting the burden of setting the correct `EXT_SUFFIX` env variable on the user. Thoughts?

h-vetinari:
> 
> 
> For this, I'd be more inclined to keep the CMake config as is, and just deal with it in `setup.py`. It's not a perfect solution, but it's better than putting the burden of setting the correct `EXT_SUFFIX` env variable on the user. Thoughts?

Understandable. I'd suggest to keep the changes in `setup.py`, but I've removed the rest for now.

h-vetinari:
@beauby 
Can this pared-down version still be considered? Without it, win+avx2 cannot be successfully installed.

PR #1695, opened 2021-02-17
Handle the situation where nprobe > nlist in IndexBinaryIVF

KinglittleQ:
## Description

It is the same as https://github.com/facebookresearch/faiss/pull/1673 but for `IndexBinaryIVF`. Ensure that `nprobe` is no more than `nlist`.

## Changes
1. Replace `nprobe` with `min(nprobe, nlist)`
2. Replace `long` with `idx_t` in `IndexBinaryIVF.cpp`
3. Add a unit test
4. Fix a small bug in https://github.com/facebookresearch/faiss/pull/1673, `index` should be replaced by `gt_index`


KinglittleQ:
@mdouze @beauby Any opinion about this PR?

tests/test_ivflib.py
```diff
@@ -179,7 +179,7 @@ def test_small_data(self):
         ref_lims, ref_D, ref_I = index.range_search(xq, thresh)
         gt_index = faiss.IndexFlat(d)
         gt_index.add(xb)
-        lims, D, I = index.range_search(xq, thresh)
+        lims, D, I = gt_index.range_search(xq, thresh)
```
mdouze, line 182: makes sense. NB that usually the ref_XXX is used for the ground truth (means reference)

tests/test_binary_ivf.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
```
mdouze, line 5: Why starting a new file? There is already test_index_binary.

mdouze (commented):
Looks mostly fine. Please address a few comments.

KinglittleQ:
@mdouze Updated. 

1. Rename variables related to ground truth as `ref_xx`.
2. The unit test is moved into `test_index_binary`.

mdouze:
LGTM.

PR #1707, opened 2021-02-24
Add NSG

KinglittleQ:
## Description:
This diff implemented Navigating Spreading-out Graph (NSG) which accepts a KNN graph as input.
Here is the interface of building an NSG graph:
``` c++
void IndexNSG::build(idx_t n, const float *x, idx_t *knn_graph, int GK);
```
where `GK` is the nb of neighbors per node and `knn_graph[i * GK + j]` is the j-th neighbor of node i. 

The `add` method is not implemented yet.

The unit tests could be found in `tests/test_nsg.cpp`.

@mdouze @beauby Maybe I need some advice on how to design the interface and support python.


faiss/impl/NSG.h
```diff
@@ -0,0 +1,191 @@
...
+
+namespace nsg {
+
+/**************************************************************
+ * Auxiliary structures
+ **************************************************************/
+
+template <class node_t>
+struct Graph {
+  node_t *data;
```
mdouze, line 57: please add comments to explain what these fields are.

faiss/impl/NSG.h
```diff
@@ -0,0 +1,191 @@
...
+      delete[] data;
+  }
+
+  inline node_t at(int i, int j) const { return data[i * K + j]; }
+
+  inline node_t &at(int i, int j) { return data[i * K + j]; }
+};
+
+/// set implementation optimized for fast access.
+struct VisitedTable {
```
mdouze, line 80: this should not be necessary once the NNDescent PR is merged.

faiss/impl/NSG.h
```diff
@@ -0,0 +1,191 @@
...
+
+  // It needs to be smaller than 0
+  static const int EMPTY_ID = -1;
+
+  int ntotal; // nb of nodes
+  int R;      // nb of neighbors per node
+  int L;      // expansion factor at construction time
+  int C;      // candidate pool size
+
+  int search_L;   // expansion factor at search time
```
mdouze, line 147: Are these the search-time parameters?

faiss/impl/NSG.cpp
```diff
@@ -0,0 +1,556 @@
...
+    return left;
+  }
+  if (addr[right].distance < nn.distance) {
+    addr[K] = nn;
+    return K;
+  }
+  while (left < right - 1) {
+    int mid = (left + right) / 2;
+    if (addr[mid].distance > nn.distance)
+      right = mid;
```
mdouze, line 78: I get a lot of warnings from the internal linter because if / while / for blocks should systematically be enclosed in {} 
Would you mind adding these in your code?

mdouze (commented):
If I understand correctly, the `build()` method should be wrapped in python so that the Python tests can be written. For this, please add an entry in __init__.py similar to 

https://github.com/facebookresearch/faiss/blob/master/faiss/python/__init__.py#L69

Basically: 

- the C++ `build` is renamed as `build_c`
- `build` points to a Python replacement_build function
- `replacement_build` checks the sizes of the inputs (should be an ndarray for x and a ndarray for knn_graph) and calls `build_c`
- get C++ pointers from ndarrays with swig_ptr

faiss/impl/NSG.h
KinglittleQ: `search_L` is a search-time parameter.

KinglittleQ: `search_L` is a search-time parameter.

faiss/impl/NSG.cpp
KinglittleQ: OK, I will solve this.

KinglittleQ:
@mdouze The PR is updated.

Changes:
1. Wrap the build method in Python
2. `VisitedTable` is moved into `AuxIndexStructures.h` as `VisitedTable` in HNSW and NSG will be conflicted in the compilation of Python wrappers.
3. Add unit tests in Python

KinglittleQ:
Updated. Fixed the conflicts and formated the code.

faiss/python/__init__.py
```diff
@@ -96,6 +96,21 @@ def replacement_decode(self, codes):
 handle_Quantizer(ScalarQuantizer)
 
 
+def handle_NSG(the_class):
+
+    def replacement_build(self, x, graph):
+        n, d = x.shape
+        graph = graph.reshape((-1,))
```
mdouze, line 103: You can assert that graph.shape[0] = n and set K = graph.shape[1]
It does not make sense to accept graphs in arbitrary shapes.

faiss/python/__init__.py
```diff
@@ -96,6 +96,21 @@ def replacement_decode(self, codes):
...
+        graph = np.ascontiguousarray(graph)
```
mdouze, line 104: It is the caller's responsibility to provide a contiguous array (this is checked in swig_ptr)

faiss/impl/NSG.cpp
```diff
@@ -0,0 +1,650 @@
...
+    final_graph.reset();
+    ntotal = 0;
+    is_built = false;
+}
+
+void NSG::init_graph(Index* storage, const nsg::Graph<idx_t>& knn_graph) {
+    int d = storage->d;
+    int n = storage->ntotal;
+
+    float* center = new float[d];
```
mdouze, line 228: For new code it is fine to use `unique_ptr` rather than `ScopeDeleter` (a legacy from when Faiss was in C++03).

faiss/impl/NSG.h
```diff
@@ -0,0 +1,198 @@
...
+    int enterpoint;
+
+    // the built graph structure
+    std::shared_ptr<nsg::Graph<int>> final_graph;
+
+    // NSG is built or not
+    bool is_built;
+
+    // random generator
+    mutable RandomGenerator rng;
```
mdouze, line 125: Is it feasible to make this a local variable? I would like to avoid having mutable fields.

faiss/impl/NSG.h
```diff
@@ -0,0 +1,198 @@
...
+    using idx_t = Index::idx_t;
+
+    // It needs to be smaller than 0
+    static const int EMPTY_ID = -1;
+
+    // nb of nodes
+    int ntotal;
+
+    // construction-time parameters
+    int R; // nb of neighbors per node
```
mdouze, line 108: Please use ///< to comment fields (doxygen can interpret those)

faiss/impl/NSG.h
```diff
@@ -0,0 +1,198 @@
...
+/***********************************************************
+ * Graph structure to store a graph.
+ *
+ * It is represented by an adjacency matrix `data`, where
+ * data[i, j] is the j-th neighbor of node i.
+ ***********************************************************/
+
+template <class node_t>
+struct Graph {
+    node_t* data;    // the flattened adjacency matrix
```
mdouze, line 57: please use ///< (doxygen)

faiss/IndexNSG.h
```diff
@@ -0,0 +1,70 @@
...
+#include <faiss/impl/NSG.h>
+#include <faiss/utils/utils.h>
+
+namespace faiss {
+
+/** The NSG index is a normal random-access index with a NSG
+ * link structure built on top */
+
+struct IndexNSG : Index {
+    using storage_idx_t = NSG::storage_idx_t;
```
mdouze, line 24: is storage_idx used somewhere?

faiss/IndexNSG.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+void IndexNSG::add(idx_t n, const float* x) {
+    // FAISS_THROW_IF_NOT_MSG(storage,
+    //    "Please use IndexNSGFlat (or variants) instead of IndexNSG directly");
+    // FAISS_THROW_IF_NOT(is_trained && is_built);
+    // int n0 = ntotal;
+    // storage->add(n, x);
+    // ntotal = storage->ntotal;
+
```
mdouze, line 141: Please implement a default add function with brute-force search. 
2nd step: support nndescent via a flag that switches between the two. 
Callers can still call `build()` themselves if they want to do it otherwise (eg. with GPU).

faiss/impl/NSG.cpp
```diff
@@ -0,0 +1,650 @@
...
+            vt.advance();
+        }
+    }
+
+    std::vector<std::mutex> locks(ntotal);
+#pragma omp parallel
+    {
+        DistanceComputer* dis = storage_distance_computer(storage);
+        ScopeDeleter1<DistanceComputer> del(dis);
+#pragma omp for schedule(dynamic, 100)
```
mdouze, line 386: Thanks! I did not know this construct. 
I think the main benefit is that it will not enter MT mode unless there are > 100 elements to process

mdouze (commented):
Thanks a lot for the update. 
Please address a few small comments, then I'll import in the internal build system and we'll see what additional issues need to be addressed.

faiss/impl/NSG.h
KinglittleQ: Get rid of the mutable variable now.

faiss/IndexNSG.h
KinglittleQ: No, it would be deleted.

faiss/impl/NSG.cpp
KinglittleQ: This means it will schedule the tasks dynamically with a chunk size of 100. The smaller the chunk size the more balance each thread works but more overhead of scheduling.

KinglittleQ:
@mdouze Updated. The GPU test failed but I have no idea of what caused it.

tests/test_index.py
```diff
@@ -592,6 +592,144 @@ def test_hnsw_IP(self):
...
+        _, self.xb, self.xq = get_dataset_2(d, nt, nb, nq)
+
+    def make_knn_graph(self, metric):
+        n = self.xb.shape[0]
+        d = self.xb.shape[1]
+        index = faiss.IndexFlat(d, metric)
+        index.add(self.xb)
+        _, I = index.search(self.xb, self.GK + 1)
+        knn_graph = np.zeros((n, self.GK), dtype=np.int64)
+        for i in range(n):
```
mdouze, line 614: I don't understand what this loop is for, what about knn_graph = I[:, 1:] ?

mdouze (commented):
Looks good. Importing to internal build sys.

The following steps are: 
1. benchmark to compare to other index types, especially HNSW, by adapting https://github.com/facebookresearch/faiss/blob/master/benchs/bench_hnsw.py
2. factory string support 
3. I/O support

These 3 steps can be done in a follow-up PR.

tests/test_index.py
KinglittleQ: It considers the inner product distance. The distance between a vector and itself may not be the smallest. For example, given a zero vector `a` and a unit vector `b`, we have `d(a, b) = 0 < d(a, a) = 1`.

h-vetinari:
This broke the windows builds on master:
```
  Building Custom Rule C:/tools/miniconda3/conda-bld/faiss-pkg_1615417741685/work/faiss/python/CMakeLists.txt
  swigfaissPYTHON_wrap.cxx
     Creating library C:/tools/miniconda3/conda-bld/faiss-pkg_1615417741685/work/_build_python_3.7/Release/swigfaiss.lib and object C:/tools/miniconda3/conda-bld/faiss-pkg_1615417741685/work/_build_python_3.7/Release/swigfaiss.exp
swigfaissPYTHON_wrap.obj : error LNK2019: unresolved external symbol "public: static int const faiss::NSG::EMPTY_ID" (?EMPTY_ID@NSG@faiss@@2HB) referenced in function Swig_var_NSG_EMPTY_ID_get [C:\tools\miniconda3\conda-bld\faiss-pkg_1615417741685\work\_build_python_3.7\swigfaiss.vcxproj]
C:\tools\miniconda3\conda-bld\faiss-pkg_1615417741685\work\_build_python_3.7\Release\_swigfaiss.pyd : fatal error LNK1120: 1 unresolved externals [C:\tools\miniconda3\conda-bld\faiss-pkg_1615417741685\work\_build_python_3.7\swigfaiss.vcxproj]
```

PR #1708, opened 2021-02-24
Remove duplicated advance()

KinglittleQ:
The `vt.advance()` statement is duplicated although it does not harm the performance too much...

This diff removed the duplicated statement.


mdouze:
Invalid because both visno and visno+1 are used, see https://github.com/facebookresearch/faiss/blob/908812266c4472b0ab07ac46a5a65635cdad8bb5/faiss/IndexHNSW.cpp#L963

PR #1710, opened 2021-02-25
Add CircleCI job to check format through clang-format.

beauby:


KinglittleQ:
It seems like the `.clang-format` file should be put under the main project folder since clang-format will find `.clang-format` automatically in the parent folders.

.circleci/config.yml
```diff
@@ -4,6 +4,29 @@ orbs:
...
+      - checkout
+      - run:
+          name: Install clang-format
+          command: |
+            apt-get update
+            apt-get install -y git-core clang-format-10
+      - run:
+          name: Verify clang-format
+          command: |
+             git ls-files | grep -E  '\.[cpp|h|cu|cuh]$' | xargs clang-format-10 -i
```
h-vetinari, line 20: This regex is likely doing something else than intended (TL;DR: `()` instead of `[]`). It sounds like you want to capture all files with extensions `.cpp`, `.h`,  `.cu`, and `.cuh`, but you're matching all extensions made of a single letter that's either `c`, `h`, `p` or `u`.

```suggestion
             git ls-files | grep -E  '\.(cpp|h|cu|cuh)$' | xargs clang-format-10 -i
```

beauby: Good catch, thanks for flagging!

Issue #1711, opened 2021-02-25
_swigfaiss_avx2.so may not be loaded properly in conda


matsui528:
# Summary

When I install faiss via conda, `IndexPQFastScan` is slower than `IndexPQ`. It seems that AVX2 is not activated properly because `_swigfaiss_avx2.so` is not loaded correctly.

# Platform

OS: Ubuntu 20.04 on AWS EC2. (ami-0e039c7d64008bd84, c5.large)

Faiss version: faiss-cpu 1.7.0 (pytorch/linux-64::faiss-cpu-1.7.0-py3.8_h2a577fa_0_cpu)

Installed from: conda install -c pytorch faiss-cpu

Faiss compilation options: 

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [x] Python

# Reproduction instructions

I found that `IndexPQFastScan` is slower than `IndexPQ` for faiss 1.7.0 installed from conda. Here is the benchmark code.

```python
import faiss
import numpy as np
import time

np.random.seed(123)
D = 128
N = 1000
X = np.random.random((N, D)).astype(np.float32)
M = 64
nbits = 4

pq = faiss.IndexPQ(D, M, nbits)
pq.train(X)
pq.add(X)

pq_fast = faiss.IndexPQFastScan(D, M, nbits)
pq_fast.train(X)
pq_fast.add(X)

t0 = time.time()
d1, ids1 = pq.search(x=X[:3], k=5)
t1 = time.time()
print(f"pq: {(t1 - t0) * 1000} msec")

t0 = time.time()
d2, ids2 = pq_fast.search(x=X[:3], k=5)
t1 = time.time()
print(f"pq_fast: {(t1 - t0) * 1000} msec")

assert np.allclose(ids1, ids2)
```

The result is:
```console
pq: 0.4680156707763672 msec
pq_fast: 1.6791820526123047 msec
```

After investigating, the cause seems that `_swigfaiss_avx2.so` is not loaded correctly. If I rename `_swigfaiss_avx2.so` to `_swigfaiss.so`, the above code works as expected:

```console
cd ~/anaconda/lib/python3.8/site-packages/faiss/
mv _swigfaiss.so _swigfaiss.so.bk
mv _swigfaiss_avx2.so _swigfaiss.so
```
Then the benchmark results in:
```console
pq: 0.8258819580078125 msec
pq_fast: 0.07104873657226562 msec
```
[...]

PR #1704, opened 2021-02-23
Make faiss compatible with Clang 11

h-vetinari:
Fixes #1696, removing a failing assertion that appeared in https://github.com/conda-forge/faiss-split-feedstock/pull/33
(and previously https://github.com/conda-forge/faiss-split-feedstock/pull/17) at the suggestion of @mdouze.

Also fixes a deprecation warning in the test suite.

PR #1717, opened 2021-02-28
Fix swig vs. AVX2

h-vetinari:
Towards #1711. There's still a test [error](https://github.com/facebookresearch/faiss/issues/1711#issuecomment-787509873) that won't be picked up unless an AVX2-specific CI run is added.

### DIFF
--- a/conda/faiss/build-pkg.sh
+++ b/conda/faiss/build-pkg.sh
@@ -10,6 +10,7 @@ set -e
 # Build avx2 version.
 cmake -B _build_python_${PY_VER}_avx2 \
       -Dfaiss_ROOT=_libfaiss_avx2_stage/ \
+      -DFAISS_OPT_LEVEL=avx2 \
       -DFAISS_ENABLE_GPU=OFF \
       -DCMAKE_BUILD_TYPE=Release \
       -DPython_EXECUTABLE=$PYTHON \
@@ -31,6 +32,6 @@ cmake --build _build_python_${PY_VER} -j $CPU_COUNT
 
 # Build actual python module.
 cp _build_python_${PY_VER}_avx2/swigfaiss.py _build_python_${PY_VER}/swigfaiss_avx2.py
-cp _build_python_${PY_VER}_avx2/_swigfaiss.so _build_python_${PY_VER}/_swigfaiss_avx2.so
+cp _build_python_${PY_VER}_avx2/_swigfaiss_avx2.so _build_python_${PY_VER}/_swigfaiss_avx2.so
 cd _build_python_${PY_VER}/
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt --prefix=$PREFIX
--- a/faiss/python/CMakeLists.txt
+++ b/faiss/python/CMakeLists.txt
@@ -61,6 +61,10 @@ swig_add_library(swigfaiss
   SOURCES swigfaiss.swig
 )
 
+if(FAISS_OPT_LEVEL STREQUAL "avx2")
+  set_target_properties(swigfaiss PROPERTIES OUTPUT_NAME "swigfaiss_avx2")
+endif()
+
 if(NOT WIN32)
   # NOTE: Python does not recognize the dylib extension.
   set_target_properties(swigfaiss PROPERTIES SUFFIX .so)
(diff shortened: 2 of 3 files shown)


mdouze:
@beauby  could you take a look?

conda/faiss/build-pkg.sh
```diff
@@ -21,6 +21,7 @@ cmake --build _build_python_${PY_VER}_avx2 -j $CPU_COUNT
 # Build vanilla version (no avx).
 cmake -B _build_python_${PY_VER} \
       -Dfaiss_ROOT=_libfaiss_stage/ \
+      -DFAISS_OPT_LEVEL=avx2 \
```
beauby, line 24: You might want to switch which build has avx2, to remain consistent with the comments.

faiss/python/CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ swig_add_library(swigfaiss
   SOURCES swigfaiss.swig
 )
 
+if(FAISS_OPT_LEVEL STREQUAL "avx2")
+  set_target_properties(swigfaiss PROPERTIES OUTPUT_NAME "swigfaiss_avx2")
```
beauby, line 65: This looks like a nice solution. Did you test it?

conda/faiss/build-pkg.sh
h-vetinari: My bad, I got the order wrong.

faiss/python/CMakeLists.txt
h-vetinari: I did indeed. :)
That's also how I found out that `OUTPUT_NAME` doesn't affect `swigfaiss.py` (not sure if this should be considered a bug in the swig-interface of CMake, but in any case, it doesn't matter much here, as we're renaming the file already).

You can download an artefact from https://github.com/conda-forge/faiss-split-feedstock/pull/34 and test it yourself.

h-vetinari:
I thinking that there should be an AVX2-specific CI run though (assuming the CI has AVX2-capable agents). There's an error in the test suite under AVX2 when building with this patch, see OP, resp. my comment in #1711.

beauby:
@h-vetinari In an ideal world I agree, but AFAIK there's no way to guarantee a CircleCI instance will have an AVX2-enabled CPU (and manual testing shows that at least some hosts do not have AVX2 CPUs).

By testing here I meant did you manually test that your patch leads to the proper .so being loaded when AVX2 is available?

h-vetinari:
Azure consistently has AVX2-enabled workers on linux + windows for example. The PRs in conda-forge run the test-suite with and without AVX2 enabled, and check that the right `.so`s have been loaded by checking the logging statements from `loader.py`. Though just the fact that the test suite now fails with this patch (where it didn't before) shows that it worked.

It's also possible to inspect the `swigfaiss_avx2.py` in the artefact, which reflects the `.so`-rename. In other words, with this patch, `swigfaiss.py` contains:
```python
# Import the low-level C/C++ module
if __package__ or "." in __name__:
    from . import _swigfaiss_avx2
else:
    import _swigfaiss_avx2
```
whereas before it was picking up the wrong `_swigfaiss.so`

beauby:
@h-vetinari Awesome, the last part is what I had in mind. As it seems that every CircleCI 2.0 host has AVX2 enabled, I'm adding a CI job for AVX2, and I'll add a test to our meta.yaml that basically checks that the correct .so is being loaded (something along the lines of `LD_DEBUG=libs python -c "import faiss" 2>&1 | grep libfaiss_avx2.so`. I'll also look into the failing test on your CI.

h-vetinari:
> something along the lines of `LD_DEBUG=libs python -c "import faiss" 2>&1 | grep libfaiss_avx2.so`

That's not bad a bad idea. To stay cross-platform, I used: 
```
pytest tests --log-file-level=INFO --log-file=log.txt
# ensure that expected logger-messages from loader.py is present;
# avoid final '\n', as well as the first 35 = len('INFO     faiss.loader:loader.py:51 ')
python -c "q = open('log.txt').readlines(); import sys; sys.exit(0 if 'Successfully loaded faiss with AVX2 support.' in [x[35:-1] for x in q] else 1)"
```
Which leverages the success messages from #1683. The downside is that this was just working on `swigfaiss.py`-level, so the problem with the wrong `.so` being loaded from withing that file didn't get flagged.

> I'll also look into the failing test on your CI.

Thanks! :)

h-vetinari:
I cherry-picked https://github.com/facebookresearch/faiss/pull/1717/commits/479abe0bfa514d845c0b5f0efabda801e7b3cadf from #1704, since that PR might take longer, but I'd still like to get rid of that annoying warning.

beauby:
@h-vetinari Would you mind reverting the test changes (and send out a separate PR) in order to keep the history clean? After that we can merge this and #1724.

PR #1723, opened 2021-03-02
Add CI job for AVX2 builds.

beauby:


h-vetinari:
I'm not sure exactly how the CI works here, since there's only one `libfaiss` and not two. However, seeing how I got a test error across all AVX2-enabled runs in https://github.com/conda-forge/faiss-split-feedstock/pull/34 with the patch from #1717, I'm not sure where that difference is coming from (except possibly by an issue in 1.7.0 that's already fixed on master).

beauby:
@h-vetinari So it seems the test was flaky (due to comparing IDs where there can be ties), and it was mitigated in https://github.com/facebookresearch/faiss/commit/27077c46274f87a50a0a28fb90433d6cfc029741#diff-e43d1c83e05a575f230512f6e4ea652bedf8df4e3fe03500bb4b943fb9f98031L249-R254.

h-vetinari:
Will give that a shot, thanks a lot! :)

PR #1724, opened 2021-03-02
Add test for CPU dispatch in conda packages.

beauby:


conda/faiss/build-pkg.sh
```diff
@@ -31,6 +32,6 @@ cmake --build _build_python_${PY_VER} -j $CPU_COUNT
 
 # Build actual python module.
 cp _build_python_${PY_VER}_avx2/swigfaiss.py _build_python_${PY_VER}/swigfaiss_avx2.py
-cp _build_python_${PY_VER}_avx2/_swigfaiss.so _build_python_${PY_VER}/_swigfaiss_avx2.so
+cp _build_python_${PY_VER}_avx2/_swigfaiss_avx2.so _build_python_${PY_VER}/_swigfaiss_avx2.so
```
h-vetinari, line 35: You need to adapt this also for the GPU build above.

h-vetinari (commented):
This could have been based on the commits form #1717...

conda/faiss/build-pkg.sh
beauby: Yeah I actually blindly copy-pasted the relevant parts from #1717 just to get to a point where the check works, I'll revert afterward, then merge #1717 and finally merge this one

PR #1738, opened 2021-03-05
fix warning about deprecate assertEquals

h-vetinari:
There's an annoying warning on every test run that I'd like to fix
```
=============================== warnings summary ===============================
tests/test_index_accuracy.py::TestRefine::test_IP
tests/test_index_accuracy.py::TestRefine::test_L2
  $SRC_DIR/tests/test_index_accuracy.py:726: DeprecationWarning: Please use assertEqual instead.
    self.assertEquals(recall1, recall2)
```

I've tried sneaking this into #1704 & #1717 already, but the first needs more time and
in the second, @beauby asked me to keep this separate, so here's a new PR. :)


mdouze:
LGTM

PR #1739, opened 2021-03-05
fix int32 vs int64 issue for GPU test on windows

h-vetinari:
Part of the work towards #1586; I already mentioned this diff as necessary for a passing
GPU test suite on windows [here](https://github.com/facebookresearch/faiss/issues/1586#issuecomment-782084250)


mdouze:
LGTM

PR #1740, opened 2021-03-05
add a subsection for installation from conda-forge

h-vetinari:
After existing for close to a year and considering the quality (IMO) of the collaboration since then,
I believe it would be reasonable to also mention the conda-forge packages in `INSTALL.md`.

WDYT @mdouze @beauby?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,9 +27,32 @@ Nightly pre-release packages can be installed as follows:
 # CPU-only version
 $ conda install -c pytorch/label/nightly faiss-cpu
 
-# GPU(+CPU) versoin
+# GPU(+CPU) version
 $ conda install -c pytorch/label/nightly faiss-gpu
 ```
+
+## Installing from conda-forge
+
+Faiss is also being packaged by [conda-forge](https://conda-forge.org/), the
+community-driven packaging ecosystem for conda. The packaging effort is
+collaborating with the Faiss team to ensure high-quality package builds.
+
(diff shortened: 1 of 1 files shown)


beauby:
@h-vetinari Yup, makes sense, feel free to open a PR.

mdouze:
Absolutely, thanks for the hard work @h-vetinari!

mdouze:
BTW how can we make that a bit more official? 
Add an entry in the CHANGELOG?
Add an entry in https://github.com/facebookresearch/faiss/wiki/Related-projects ?
Post to the Faiss users group on Facebook to say that we endorse conda-forge?

h-vetinari:
> 
> 
> Absolutely, thanks for the hard work @h-vetinari!

Happy to hear it! :)

> BTW how can we make that a bit more official?
> Add an entry in the CHANGELOG?
> Add an entry in https://github.com/facebookresearch/faiss/wiki/Related-projects ?
> Post to the Faiss users group on Facebook to say that we endorse conda-forge?

All of those sound fine to me - would be better if this comes from the maintainers though...?

h-vetinari:
> 
> 
> BTW how can we make that a bit more official?
> Add an entry in the CHANGELOG?
> Add an entry in https://github.com/facebookresearch/faiss/wiki/Related-projects ?
> Post to the Faiss users group on Facebook to say that we endorse conda-forge?

Now that the win+cuda builds are done, I think it would be a good time to announce that? What do you think?

PR #1742, opened 2021-03-05
Portable SWIG Vectors

### LINKED ISSUE #1741: RFC: replace {Long,LongLong,...}Vector with {Int32,Int64,...}Vector (with deprecation)
In trying to get `faiss-gpu` to work with windows (cf. #1586), I'm hitting a couple of points where the different architectures collide. One of the main points about this revolves about what types `int`, `long` and `long long` actually correspond to on the respective platforms.

Since these types are - sadly - fundamentally not cross-platform compatible, it is a bit unfortunate that the SWIG interface currently defines `IntVector`, `LongVector`, `LongLongVector`.

For reasons of portability, I therefore wanted to suggest to deprecate these names and replace them with something unambiguous, e.g:
* `CharVector` -> `Int8Vector`
* `IntVector` -> `Int32Vector`
* `LongVector` -> `Int32Vector if sizeof_long == 4 else Int64Vector`
* `LongLongVector` -> `Int64Vector`
* `ByteVector` -> `UInt8Vector`
* `FloatVector` -> `Float32Vector`
* `DoubleVector` -> `Float64Vector`

This is the maximal reasonable set of renames/deprecations (`Int16Vector`, `UInt16Vector`, `UInt32Vector`, `UInt64Vector` exist already), and arguably not all of the names on the left _need_ to be deprecated. However, the uniform naming structure and intrinsic match with the underlying bit-width (as well as the numpy-types) is something that would IMO also substantially reduce the cognitive burden of figuring out what object one is working with ("what was `long` again on this system?"; or not knowing that `char` or `byte` correspond to integer types).

I have a solution for doing exactly that which I believe is fairly elegant and unobtrusive - essentially generating wrapper classes around the equivalent non-deprecated versions in `__init__.py`, with the only difference that they raise a `DeprecationWarning` _once_ per session (if called). The advantage of this is that this can be done in pure python, and doesn't need any surgery on the SWIG interface.

If desired, I can post the patches directly, but wanted to first get a feel if this would be welcome.

WDYT @wickedfoo @beauby @mdouze?

h-vetinari:
After initial positive feedback to the idea in #1741 from @mdouze, here are the patches
I currently have as a basis for discussion.

Matthijs suggests to not bother with the deprecation warnings at all, which is fine for me
as well, though I would normally still try to provide users with _some_ advance notice
before removing parts of an interface.

Fixes #1741

PS. I haven't touched the `XXXVectorVector` classes yet, happy to add that here as well.


faiss/python/__init__.py
```diff
@@ -656,23 +657,61 @@ def knn_gpu(res, xq, xb, k, D=None, I=None, metric=METRIC_L2):
...
+    'Float': 'Float32',
+    'Double': 'Float64',
+    'Char': 'Int8',
+    'Int': 'Int32',
+    'Long': 'Int32' if sizeof_long == 4 else 'Int64',
+    'LongLong': 'Int64',
+    'Byte': 'UInt8',
+    # previously misspelled variant
+    'Uint64': 'UInt64',
 }
```
h-vetinari, line 672: _Not_ deprecating some of these (e.g. if it's considered desirable to keep `DoubleVector`) is easy: we would just need to move the line down from this dict to `vector_name_map` (and lower-case the value of the KV-pair)

faiss/python/swigfaiss.swig
```diff
@@ -204,23 +204,21 @@ namespace std {
 %include <std_shared_ptr.i>
```
beauby, line 205: While we're at it, we might as well ditch the above `std::vector` class declaration, and replace it with `%include <std_vector.i>`.

beauby:
Looks good, could you also add a line to the CHANGELOG.md (under `changed`)?

h-vetinari:
OSX error looks unfriendly:
```
/Users/distiller/project/build/faiss/python/CMakeFiles/swigfaiss.dir/swigfaissPYTHON_wrap.cxx:108209:12: error: assigning to 'std::vector<size_t, std::allocator<size_t> > *' (aka 'vector<unsigned long, allocator<unsigned long> > *') from incompatible type 'std::vector<unsigned long long, std::allocator<unsigned long long> > *'
    arg6 = ptr;
```

For whatever reason, it seems the custom interface didn't care about this...

faiss/python/swigfaiss.swig
beauby: Actually, it seems the SWIG wrappers for `std::vector` do not expose the `data()` method, so we might have to revert.

h-vetinari: Done

h-vetinari:
@beauby 
Any update on this? I reverted the removal of the custom `std::vector` class, as you asked.

h-vetinari:
Windows failure was caused by #1707, I believe

h-vetinari:
Ping @beauby 
Anything left for me to do here (rebase)?

beauby:
@h-vetinari Would you mind rebasing (now that CI is fixed)? @mdouze If you agree with those changes, please stamp the diff on Phabricator.

h-vetinari:
Thanks for picking this up again, it rebased cleanly 👍

h-vetinari:
Anything for me to do based on the failing internal builds & tests? I obviously cannot see them... 😅

Issue #1743, opened 2021-03-06
FileNotFoundError: [Errno 2] No such file or directory: '_swigfaiss.so'


KinglittleQ:
# Summary

When compiling from source following the instructions in INSTALL.md, the following error occurred:
``` shell
$ python setup.py install                                                    
Traceback (most recent call last):
  File "setup.py", line 22, in <module>                                      
    shutil.copyfile(f"{prefix}_swigfaiss{ext}", f"faiss/_swigfaiss{ext}")
  File "/opt/conda/lib/python3.7/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '_swigfaiss.so'
```
The CMake/make part passed but the Python setup failed.

# Platform

OS: 

Faiss version: 

Installed from: compiled from source

Faiss compilation options: 

cmake -B build -DCMAKE_BUILD_TYPE=Release -DFAISS_OPT_LEVEL=avx2 -DFAISS_ENABLE_GPU=OFF -DMKL_LIBRARIES=/opt/conda/lib -DPython_EXECUTABLE=$(which python)

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [x] Python

# Reproduction instructions

 

``` shell
cmake -B build -DCMAKE_BUILD_TYPE=Release -DFAISS_OPT_LEVEL=avx2 -DFAISS_ENABLE_GPU=OFF -DMKL_LIBRARIES=/opt/conda/lib -DPython_EXECUTABLE=$(which python)
make -C build -j faiss
make -C build -j swigfaiss
cd build/faiss/python
python setup.py install
```

KinglittleQ:
https://github.com/facebookresearch/faiss/blob/f2464141a7e15e9e201de84eeb360955aa229364/faiss/python/CMakeLists.txt#L65

This line of code makes the installation failed. Removing it could fix the problem.

mdouze:
Ah right, that's an artifact of the previous AVX2 fix... @beauby how could we handle this?

KinglittleQ:
Solved by https://github.com/facebookresearch/faiss/pull/1746.

PR #1744, opened 2021-03-06
Combine AVX with non-AVX in Python setup

KinglittleQ:
This diff fixed https://github.com/facebookresearch/faiss/issues/1743.
The python setup script is a little bit too complicated now. In my view, it is not necessary to separate AVX and non-AVX in the building.

### DIFF
--- a/conda/faiss-gpu/build-pkg.sh
+++ b/conda/faiss-gpu/build-pkg.sh
@@ -31,7 +31,7 @@ cmake --build _build_python_${PY_VER} -j $CPU_COUNT
 
 
 # Build actual python module.
-cp _build_python_${PY_VER}_avx2/swigfaiss.py _build_python_${PY_VER}/swigfaiss_avx2.py
-cp _build_python_${PY_VER}_avx2/_swigfaiss_avx2.so _build_python_${PY_VER}/_swigfaiss_avx2.so
+cp _build_python_${PY_VER}_avx2/swigfaiss.py _build_python_${PY_VER}/swigfaiss.py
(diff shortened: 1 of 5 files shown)


h-vetinari (commented):
This would force everyone to either build the package themselves, or will break all users that do not have an AVX2-enabled CPU. The whole point about compiling both is that the library can be used by everyone, but will use the advantages of AVX2 if they are available.

You can still build the package yourself with only one variant. The only thing to pay attention to then is _not_ to pass `-DFAISS_OPT_LEVEL=avx2` to `cmake ... faiss/python`.

h-vetinari:
Having now looked at #1743, this PR is IMO overkill to achieve what you want. The only thing that would be necessary to enable the "configure once, build everything" workflow is to _rename_ the switch used in `faiss/python/CMakeLists.txt` so that it doesn't coincide with `FAISS_OPT_LEVEL`.

Note that the following should already work as-is on master:
```
cmake -B build -DCMAKE_BUILD_TYPE=Release -DFAISS_OPT_LEVEL=avx2 -DFAISS_ENABLE_GPU=OFF -DMKL_LIBRARIES=/opt/conda/lib
make -C build -j faiss
cmake -B build -DCMAKE_BUILD_TYPE=Release -DFAISS_ENABLE_GPU=OFF -DMKL_LIBRARIES=/opt/conda/lib -DPython_EXECUTABLE=$(which python)
make -C build -j swigfaiss
cd build/faiss/python
python setup.py install
```

KinglittleQ:
Looks like the source code should be compiled twice to generate AVX and non-AVX prebuilt libraries while publishing to the conda channel. I missed this point. Closed.

PR #1746, opened 2021-03-06
Fix building with -DFAISS_OPT_LEVEL=avx2

KinglittleQ:
As stated in https://github.com/facebookresearch/faiss/issues/1743, `_swigfaiss.so` will not exist if compiled with `-DFAISS_OPT_LEVEL=avx2`. This diff added a conditional statement before copying it.

### DIFF
--- a/faiss/python/setup.py
+++ b/faiss/python/setup.py
@@ -29,7 +29,7 @@ found_swigfaiss_general = os.path.exists(swigfaiss_general_lib)
 found_swigfaiss_avx2 = os.path.exists(swigfaiss_avx2_lib)
 
 assert (found_swigfaiss_general or found_swigfaiss_avx2), \
-        f"Could not find {swigfaiss_general_lib} and " \
+        f"Could not find {swigfaiss_general_lib} or " \
         f"{swigfaiss_avx2_lib}. Faiss may not be compiled yet."
 
 if found_swigfaiss_general:
--- a/faiss/python/setup.py
+++ b/faiss/python/setup.py
@@ -16,8 +16,6 @@ shutil.copytree("contrib", "faiss/contrib")
 shutil.copyfile("__init__.py", "faiss/__init__.py")
 shutil.copyfile("loader.py", "faiss/loader.py")
 shutil.copyfile("swigfaiss.py", "faiss/swigfaiss.py")
-if os.path.exists("swigfaiss_avx2.py"):
-    shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")
 
 ext = ".pyd" if platform.system() == 'Windows' else ".so"
 prefix = "Release/" * (platform.system() == 'Windows')
--- a/faiss/python/setup.py
+++ b/faiss/python/setup.py
@@ -21,7 +21,6 @@ if os.path.exists("swigfaiss_avx2.py"):
 
 ext = ".pyd" if platform.system() == 'Windows' else ".so"
 prefix = "Release/" * (platform.system() == 'Windows')
-
 faiss_exists = os.path.exists(f"{prefix}_swigfaiss{ext}")
 faiss_avx2_exists = os.path.exists(f"{prefix}_swigfaiss_avx2{ext}")
 assert(faiss_exists or faiss_avx2_exists)
--- a/faiss/python/setup.py
+++ b/faiss/python/setup.py
@@ -21,17 +21,24 @@ if os.path.exists("swigfaiss_avx2.py"):
 
 ext = ".pyd" if platform.system() == 'Windows' else ".so"
 prefix = "Release/" * (platform.system() == 'Windows')
-faiss_exists = os.path.exists(f"{prefix}_swigfaiss{ext}")
-faiss_avx2_exists = os.path.exists(f"{prefix}_swigfaiss_avx2{ext}")
-assert(faiss_exists or faiss_avx2_exists)
 
-if faiss_exists:
-    print("Found swigfaiss")
-    shutil.copyfile(f"{prefix}_swigfaiss{ext}", f"faiss/_swigfaiss{ext}")
+swigfaiss_general_lib = f"{prefix}_swigfaiss{ext}"
+swigfaiss_avx2_lib = f"{prefix}_swigfaiss_avx2{ext}"
 
-if faiss_avx2_exists:
-    print("Found swigfaiss_avx2")
-    shutil.copyfile(f"{prefix}_swigfaiss_avx2{ext}", f"faiss/_swigfaiss_avx2{ext}")
+found_swigfaiss_general = os.path.exists(swigfaiss_general_lib)
+found_swigfaiss_avx2 = os.path.exists(swigfaiss_avx2_lib)
+
+assert (found_swigfaiss_general or found_swigfaiss_avx2), \
+        f"Could not find {swigfaiss_general_lib} and " \
+        f"{swigfaiss_avx2_lib}. Faiss may not be compiled yet."
+
+if found_swigfaiss_general:
+    print(f"Copying {swigfaiss_general_lib}")
+    shutil.copyfile(swigfaiss_general_lib, f"faiss/_swigfaiss{ext}")
+
+if found_swigfaiss_avx2:
+    print(f"Copying {swigfaiss_avx2_lib}")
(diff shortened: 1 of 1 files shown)


mdouze:
@beauby @h-vetinari any comment on this?

kj9lj (approved):

faiss/python/setup.py
```diff
@@ -16,17 +16,22 @@
...
 shutil.copyfile("swigfaiss.py", "faiss/swigfaiss.py")
+if os.path.exists("swigfaiss_avx2.py"):
+    shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")
 
 ext = ".pyd" if platform.system() == 'Windows' else ".so"
 prefix = "Release/" * (platform.system() == 'Windows')
-shutil.copyfile(f"{prefix}_swigfaiss{ext}", f"faiss/_swigfaiss{ext}")
+faiss_exists = os.path.exists(f"{prefix}_swigfaiss{ext}")
+faiss_avx2_exists = os.path.exists(f"{prefix}_swigfaiss_avx2{ext}")
+assert(faiss_exists or faiss_avx2_exists)
```
h-vetinari, line 26: This should be raised as an error with an appropriate error message, IMO.

faiss/python/setup.py
```diff
@@ -16,17 +16,22 @@
...
 
-try:
-    shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")
+if faiss_exists:
+    print("Found swigfaiss")
```
h-vetinari, line 29:
```suggestion
    print(f"Copying {prefix}_swigfaiss{ext}")
```

faiss/python/setup.py
```diff
@@ -16,17 +16,22 @@
...
+    shutil.copyfile(f"{prefix}_swigfaiss{ext}", f"faiss/_swigfaiss{ext}")
+
+if faiss_avx2_exists:
+    print("Found swigfaiss_avx2")
```
h-vetinari, line 33:
```suggestion
    print(f"Copying {prefix}_swigfaiss_avx2{ext}")
```

h-vetinari (commented):
Don't have a strong opinion on this, but left some comments.

Names of the booleans could also be improved, e.g.
`faiss_exists` -> `found_swigfaiss_general`
`faiss_avx2_exists` -> `found_swigfaiss_avx2`

KinglittleQ:
@beauby @h-vetinari Updated.

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
-    pass
+swigfaiss_general_lib = f"{prefix}_swigfaiss{ext}"
+swigfaiss_avx2_lib = f"{prefix}_swigfaiss_avx2{ext}"
+
+found_swigfaiss_general = os.path.exists(swigfaiss_general_lib)
+found_swigfaiss_avx2 = os.path.exists(swigfaiss_avx2_lib)
+
+assert (found_swigfaiss_general or found_swigfaiss_avx2), \
+        f"Could not find {swigfaiss_general_lib} or " \
+        f"{swigfaiss_avx2_lib}. Faiss may not be compiled yet."
```
h-vetinari, line 33:
```suggestion
if not (found_swigfaiss_general or found_swigfaiss_avx2):
    raise FileNotFoundError(f"Could find neither {swigfaiss_general_lib} nor "
                            f"{swigfaiss_avx2_lib}. Faiss may not be compiled yet.")
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
 prefix = "Release/" * (platform.system() == 'Windows')
-shutil.copyfile(f"{prefix}_swigfaiss{ext}", f"faiss/_swigfaiss{ext}")
 
-try:
-    shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")
-    shutil.copyfile(f"{prefix}_swigfaiss_avx2{ext}", f"faiss/_swigfaiss_avx2{ext}")
-except Exception as e:
-    print(f"Could not move swigfaiss_avx2.py / {prefix}_swigfaiss_avx2{ext} due to:\n{e!r}")
-    pass
+swigfaiss_general_lib = f"{prefix}_swigfaiss{ext}"
```
beauby, line 25:
```suggestion
swigfaiss_generic_lib = f"{prefix}_swigfaiss{ext}"
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
+swigfaiss_avx2_lib = f"{prefix}_swigfaiss_avx2{ext}"
+
+found_swigfaiss_general = os.path.exists(swigfaiss_general_lib)
```
beauby, line 28:
```suggestion
found_swigfaiss_generic = os.path.exists(swigfaiss_generic_lib)
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
+found_swigfaiss_avx2 = os.path.exists(swigfaiss_avx2_lib)
+
+assert (found_swigfaiss_general or found_swigfaiss_avx2), \
```
beauby, line 31:
```suggestion
assert (found_swigfaiss_generic or found_swigfaiss_avx2), \
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
+        f"Could not find {swigfaiss_general_lib} or " \
```
beauby, line 32:
```suggestion
        f"Could not find {swigfaiss_generic_lib} or " \
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
+        f"{swigfaiss_avx2_lib}. Faiss may not be compiled yet."
+
+if found_swigfaiss_general:
+    print(f"Copying {swigfaiss_general_lib}")
+    shutil.copyfile(swigfaiss_general_lib, f"faiss/_swigfaiss{ext}")
```
beauby, line 37:
```suggestion
if found_swigfaiss_generic:
    print(f"Copying {swigfaiss_generic_lib}")
    shutil.copyfile(swigfaiss_generic_lib, f"faiss/_swigfaiss{ext}")
```

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
 shutil.copyfile("__init__.py", "faiss/__init__.py")
 shutil.copyfile("loader.py", "faiss/loader.py")
 shutil.copyfile("swigfaiss.py", "faiss/swigfaiss.py")
```
beauby, line 18: Should this be guarded by an `if os.path.exists` as well?

beauby (commented):
Thanks for the PR, a few minor suggestions but we can merge it soon.

faiss/python/setup.py
KinglittleQ: Your suggestion is reasonable but I think a simple assertion is enough here.

faiss/python/setup.py
KinglittleQ: `swigfaiss.py` is always generated regardless of FAISS_OPT_LEVEL.

faiss/python/setup.py
```diff
@@ -16,17 +16,29 @@
...
+if os.path.exists("swigfaiss_avx2.py"):
+    shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")
```
beauby, line 20: Related to the above: it seems CMake's SWIG target generates `swigfaiss.py` regardless of whether it is an AVX2 build or not (and hence never generates anything called `swigfaiss_avx2.py`). Long term fix will be to introduce separate targets for `libfaiss`/`libfaiss_avx2` (and `swigfaiss`/`swigfaiss_avx2`), but for now I think we can just remove this `shutil.copyfile("swigfaiss_avx2.py", "faiss/swigfaiss_avx2.py")` as it is never executed.

faiss/python/setup.py
h-vetinari: Ultimately up to the maintainers here, but I think bare asserts (outside of tests) are bad style. You should return a specific error.

faiss/python/setup.py
KinglittleQ: That's good.

PR #1758, opened 2021-03-14
Support NSG in the index factory

KinglittleQ:
## Description
This PR added NSG into the index factory. Here are the supported index strings:
1. `NSG{0}` or `NSG{0},Flat`: Create an IndexNSGFlat with `R = {0}`.
2. `IVF{0}_NSG{1},{2}`: Create an IndexIVF using NSG as a coarse quantizer where `ncentroids = {0}`, `R = {1}` and `{2}` is the second level quantizer.

These two types of indexes may be the most useful ones. Other composite indexes could be supported in the future.


mdouze:
Very good, thanks !

PR #1766, opened 2021-03-17
Support I/O and clone for NSG

KinglittleQ:
This PR added IO and clone support to NSG.

### DIFF
--- a/faiss/IndexNSG.cpp
+++ b/faiss/IndexNSG.cpp
@@ -85,7 +85,8 @@ void IndexNSG::search(
             storage,
             "Please use IndexNSGFlat (or variants) instead of IndexNSG directly");
 
-    idx_t check_period = InterruptCallback::get_period_hint(d * nsg.search_L);
+    int L = std::max(nsg.search_L, (int)k); // in case of search L = -1
+    idx_t check_period = InterruptCallback::get_period_hint(d * L);
 
     for (idx_t i0 = 0; i0 < n; i0 += check_period) {
         idx_t i1 = std::min(i0 + check_period, n);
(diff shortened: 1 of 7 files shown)


faiss/impl/NSG.cpp
```diff
@@ -285,7 +285,7 @@ void NSG::search_on_graph(
         int pool_size,
         std::vector<Neighbor>& retset,
         std::vector<Node>& fullset) const {
-    RandomGenerator gen(rand());
+    RandomGenerator gen(0x1234);
```
h-vetinari, line 288: Not very random if you always initialize with the same seed, or am I missing something?

KinglittleQ: `rand()` is not thread-safe and I'm trying to make the process of searching deterministic. This change does not harm the performance according to the experiment I made.

KinglittleQ:
## Updated:
Fix a bug that will occur when search_L = -1.

Before:
```
search_L -1        0.133 ms per query, R@1 0.2889, missing rate 0.0000
```

After:
```
search_L -1        0.005 ms per query, R@1 0.2904, missing rate 0.0000
```

faiss/IndexNSG.cpp
```diff
@@ -85,7 +85,8 @@ void IndexNSG::search(
             storage,
             "Please use IndexNSGFlat (or variants) instead of IndexNSG directly");
 
-    idx_t check_period = InterruptCallback::get_period_hint(d * nsg.search_L);
```
KinglittleQ, line 88: When nsg.search_L = -1, it will cause unexpected behaviour.

mdouze:
Look good!

PR #1775, opened 2021-03-22
added doxygen flag + test

mdouze:
Added a flag to convert doxygen comments into docstrings.

### DIFF
--- a/faiss/impl/NSG.h
+++ b/faiss/impl/NSG.h
@@ -101,7 +101,10 @@ struct NSG {
     int ntotal; ///< nb of nodes
 
     /// construction-time parameters
-    int R; ///< nb of neighbors per node
+
+
+    /// number of neighbors per node
+    int R;
     int L; ///< length of the search path at construction time
     int C; ///< candidate pool size at construction time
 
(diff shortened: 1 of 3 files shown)


faiss/impl/NSG.h
```diff
@@ -101,7 +101,10 @@ struct NSG {
     int ntotal; ///< nb of nodes
 
     /// construction-time parameters
-    int R; ///< nb of neighbors per node
+
+
+    /// number of neighbors per node
+    int R;
```
mdouze, line 107: Cannot be parsed by -doxygen otherwise

faiss/python/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ set_source_files_properties(swigfaiss.swig PROPERTIES
 
 if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT WIN32)
   set_source_files_properties(swigfaiss.swig PROPERTIES
-    SWIG_FLAGS -DSWIGWORDSIZE64
+    SWIG_FLAGS "-DSWIGWORDSIZE64 -doxygen"
```
mdouze, line 29: should probably be made optional but I don't know how to do that in cmake

tests/test_documentation.py
```diff
@@ -15,5 +15,8 @@ def test_io_error(self):
 
         self.assertTrue("Adds vectors to the index" in index.add.__doc__)
 
-# At some point we will check here if SWIG imported the C++ doxygen comments
-# for the classes as docstrings
+    def test_class_doc(self):
+        index = faiss.IndexFlat(32)
```
mdouze, line 19: should run this test only if not in fbcode

PR #1790, opened 2021-03-29
Add get,set verbose for Index in c_api

ava57r:


c_api/Index_c.cpp
```diff
@@ -24,6 +24,9 @@ DEFINE_GETTER(Index, idx_t, ntotal)
 
 DEFINE_GETTER(Index, FaissMetricType, metric_type)
 
+DEFINE_GETTER(Index, bool, verbose);
+DEFINE_SETTER(Index, bool, verbose);
```
Enet4, line 28:
```suggestion
DEFINE_GETTER(Index, int, verbose);
DEFINE_SETTER(Index, int, verbose);
```

c_api/Index_c.h
```diff
@@ -55,6 +55,8 @@ FAISS_DECLARE_GETTER(Index, idx_t, ntotal)
 /// Getter for metric_type
 FAISS_DECLARE_GETTER(Index, FaissMetricType, metric_type)
 
+FAISS_DECLARE_GETTER_SETTER(Index, bool, verbose)
```
Enet4, line 58:
```suggestion
FAISS_DECLARE_GETTER_SETTER(Index, int, verbose)
```

Enet4 (changes requested):
`bool` exists as a built-in type in C++, but not in C (IIRC it's behind a stdbool header). I have instead used `int` throughout the C API, and let the compiler implicitly convert booleans back and forth.

On a side note, I was considering the use of Bool_

PR #1792, opened 2021-03-29
Add tests for avx2 building

KinglittleQ:
This PR added some tests to make sure the building with AVX2 works as we expected in Linux.


tests/test_fast_scan.py
```diff
@@ -33,6 +33,38 @@ def test_PQ4_accuracy(self):
...
+        index2 = faiss.index_factory(32, 'PQ16x4fs')
+        index2.train(xt)
+        index2.add(xb)
+
+        t0 = time.time()
+        D2, I2 = index2.search(xq, 10)
+        t1 = time.time()
+        pqfs_t = t1 - t0
+        print('PQ16x4fs search time:', pqfs_t)
+        self.assertLess(pqfs_t * 10, pq_t)
```
KinglittleQ, line 66: It would raise an error here If avx2 did not properly work.

.circleci/config.yml
```diff
@@ -70,6 +70,15 @@ jobs:
...
             python3 setup.py build
+      - when:
+          condition:
+            equal: [ "avx2", << parameters.opt_level >> ]
+          steps:
+            - run:
+                name: Test avx2 building
+                command: |
+                  export PYTHONPATH="$(ls -d ./build/faiss/python/build/lib*/)"
+                  python3 -c "import faiss; assert faiss.has_AVX2"
```
KinglittleQ, line 81: Make sure that faiss.has_AVX2 is set properly.

tests/test_fast_scan.py
```diff
@@ -33,6 +33,38 @@ def test_PQ4_accuracy(self):
         assert recall_at_1 > 0.6
         # print(f'recall@1 = {recall_at_1:.3f}')
 
+    def test_PQ4_speed(self):
+        # only test while building with avx2
+        print(faiss.has_AVX2)
+        if not faiss.has_AVX2:
+            return
```
beauby, line 40: Please use the `@unittest.skipIf()` decorator instead.

KinglittleQ: Thanks! I do not even know this usage.

mdouze:
Sorry for the delay, the internal tests fail on 
```
  File "/data/sandcastle/boxes/eden-trunk-hg-fbcode-fbsource/fbcode/buck-out/dev/gen/faiss/tests/test_fast_scan#binary,link-tree/test_fast_scan.py", line 36, in TestSearch
    @unittest.skipIf(not faiss.has_AVX2, "only test while building with avx2")
AttributeError: module 'faiss' has no attribute 'has_AVX2'
```
Where was the has_AVX2 added?

KinglittleQ:
It is set in `loader.py`:
https://github.com/facebookresearch/faiss/blob/267edb120b5d2b8d6abb884c30dfb94ec1a7331f/faiss/python/loader.py#L51
And then imported in `__init__.py`:
https://github.com/facebookresearch/faiss/blob/267edb120b5d2b8d6abb884c30dfb94ec1a7331f/faiss/python/__init__.py#L18

h-vetinari:
`has_AVX2` should probably be made private... It can still be used in the test suite, but I didn't add it to the API intentionally, it just seemed cleaner to re-use a variable for that (keeping the reloading code DRY by unifying the AVX2-fallback with the generic case).

KinglittleQ:
@h-vetinari Do you have any idea about how to detect which version (avx2 or generic) is loaded? `faiss.has_AVX2` works in the test suite but I'm not sure if it is always set after `import faiss`.

KinglittleQ:
@mdouze Updated. Could you take a look at it again? Now the version of faiss is detected by a global variable `simd_set` set in C++.

mdouze:
Thanks for the addition, I think that having C++ say how Faiss was compiled is the good way to go.
I think `has_AVX2` does not appear in the FB compile system because it is loaded differently.

However, could you make it a bit more generic? Could you add a function in utils.h/.cpp
```
std::string get_compile_options();
```
that returns a string containing "AVX2" if it is compiled with AVX2? 

The rationale is that other platforms have other implementations of simd, so we need to be more specific.

Thanks!

KinglittleQ:
@mdouze Updated.

mdouze:
Looks good to me. @beauby  @h-vetinari  any opinion on this?

tests/test_fast_scan.py
```diff
@@ -33,6 +33,35 @@ def test_PQ4_accuracy(self):
         assert recall_at_1 > 0.6
         # print(f'recall@1 = {recall_at_1:.3f}')
 
+    @unittest.skipUnless('AVX2' in faiss.get_compile_options(),
+            "only test while building with avx2")
```
h-vetinari, line 37: Could we have a test for the python-output of `faiss.get_compile_options()`?

I.e. `assert faiss.get_compile_options() in [<list of possible outputputs>]`?

Also, the indent here is off.

tests/test_fast_scan.py
h-vetinari: I see what the intention is, I'm just worried that it might be a flaky test. But fine IMO to try and see how it goes for a while.

h-vetinari (commented):
Can't verify the correctness of the time-based test (and left a nit), but otherwise looks fine to me. :)

h-vetinari:
> 
> 
> @h-vetinari Do you have any idea about how to detect which version (avx2 or generic) is loaded? `faiss.has_AVX2` works in the test suite but I'm not sure if it is always set after `import faiss`.

@beauby had set a test based on `LD_DEBUG` see [here](https://github.com/facebookresearch/faiss/blob/master/conda/faiss-gpu/test_cpu_dispatch.sh); that would be awkward to do from python though, and isn't portable to windows...

KinglittleQ:
Updated again. Add a test for `get_compile_options()`.

KinglittleQ:
@mdouze @beauby Updated.

I think a util function to tell users which version is used may be enough hence the added CI job is removed now.

This PR probably could help https://github.com/facebookresearch/faiss/issues/1711.

tests/test_fast_scan.py
mdouze: @h-vetinari I agree it is tricky to do timing tests. However, we've had so many timing regressions that were discovered months down the road by users that I am eager to try something to avoid them.

Issue #1812, opened 2021-04-08
Improve performance of IndexPQFastScan and IndexIVFPQFastScan on aarch64 and non-AVX2 devices


matsui528:
# Summary

Dear @mdouze and all,

Currently, `IndexPQFastScan` and `IndexIVFPQFastScan` are optimized only for AVX2. Thus, the algorithms are not much fast in non-AVX2 environments.

Our team (@vorj, @n-miyamoto-fixstars, @LWisteria, and @matsui528) has accelerated the above two algorithms for the following two environments:

- for non-AVX2 environment (e.g., old x86 computer w/o AVX2)
    - We optimized `simdlib_emulated.h`, enabling 4x speedup for SIFT1M.
- for aarch64 (e.g., Rasberry Pi)
    - We implemented `simdlib_neon.h`, which is a NEON counterpart of `simdlib_avx2.h`.
    - This achieved 60x faster performance on SIFT1M.

If you are likely to merge, we plan to submit a PR for the above changes. What do you think? The changes are mainly in the above header files and do not significantly change the faiss codebase.

Kind regards,

# Platform

OS: Ubuntu 20.04.2 LTS (GNU/Linux 5.4.0-1038-aws aarch64)

Faiss version: fe7b06197ee35c1f2677f677558239834133358b

Installed from: compiled by myself

Faiss compilation options: `cmake -DFAISS_ENABLE_GPU=OFF -DFAISS_OPT_LEVEL=aarch64 -DCMAKE_BUILD_TYPE=Release`

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [x] C++
- [ ] Python

# Reproduction instructions
We only post the results. If you need more detailed information, please let me know.
![1](https://user-images.githubusercontent.com/2842345/114064000-29ff0680-98d4-11eb-921e-ff8d19a77d95.png)
- Evaluated on an AWS EC2 ARM instance (c6g.4xlarge)
- `original` is the current code
- `improved-emulated` is the result of optimizing  `simdlib_emulated.h`, which is faster than `original`
- `neon` is the result by `simdlib_neon.h`, which is much faster than `original`

![2](https://user-images.githubusercontent.com/2842345/114064001-2b303380-98d4-11eb-85de-8855fd4cc3a3.png)
The above image illustrates the ratio of speedup.
- In the best case, `neon` is approx. 60x faster than `original` (M=32, nbits=4, nprobe=16)
    - `original`: 4.5 ms
[...]

PR #1814, opened 2021-04-12
Make simdlib_emulated.h faster

vorj:
related: #1812 

This PR improves the performance of contents in `simdlib_emulated.h` .
`IndexPQFastScan` and `IndexIVFPQFastScan` will become faster on non-AVX2 environments, e.g., 4x faster on SIFT1M.
This PR contains below changes:

- Use `template` instead of `std::function` on argument of `unary_func` and `binary_func`
    - Because `std::function` hinders some optimizations like function inlining
- Use `const T&` instead of `T` for vector classes like `simd16uint16` on argument of functions
    - Vector classes on `simdlib_emulated.h` has the data member as array, so the runtime cost for copying is not so low.
    - Passing by const lvalue-ref prevents copy.


mdouze:
Looks good! thanks for the careful performance evaluation and fixes.

PR #1816, opened 2021-04-12
Impl IndexPreTransform for c_api

ava57r:
This PR extends c_api for IndexPreTransform


ava57r:

Enet4 (approved):
Seems good to go.

The matter of immutable reference guarantee may be handled at `faiss-rs`.

PR #1821, opened 2021-04-14
Improve impl IndexRefineFlat for c_api

ava57r:
This PR extends c_api for IndexRefineFlat


ava57r:

mdouze:
I did not notice there are so many typos in the Faiss source ;-)

PR #1827, opened 2021-04-15
Add setters for IndexIVF* indexes

ava57r:
This PR gives users control over resources

### DIFF
--- a/c_api/IndexIVFFlat_c.cpp
+++ b/c_api/IndexIVFFlat_c.cpp
@@ -25,6 +25,8 @@ DEFINE_INDEX_DOWNCAST(IndexIVFFlat)
 DEFINE_GETTER(IndexIVFFlat, size_t, nlist)
 /// number of probes at query time
 DEFINE_GETTER(IndexIVFFlat, size_t, nprobe)
+DEFINE_SETTER(IndexIVFFlat, size_t, nprobe)
+
 /// quantizer that maps vectors to inverted lists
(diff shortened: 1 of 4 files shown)


c_api/IndexIVFFlat_c.cpp
```diff
@@ -25,6 +25,8 @@ DEFINE_INDEX_DOWNCAST(IndexIVFFlat)
 DEFINE_GETTER(IndexIVFFlat, size_t, nlist)
 /// number of probes at query time
 DEFINE_GETTER(IndexIVFFlat, size_t, nprobe)
+DEFINE_SETTER(IndexIVFFlat, size_t, nprobe)
```
mdouze, line 28: Not sure this setter is very safe to use, but for consistency with C++ it can be added.

ava57r: I can change this setter

Enet4:
It comes down to whether the given setters unlock practical use cases. Not all of them will be safe to use, but if they are not made available, it may also prevent legitimate uses from this API.

I don't have a strong opinion about setting `nprobe`, although it appears that this makes sense in the [Python API](https://github.com/facebookresearch/faiss/blob/master/tutorial/python/3-IVFPQ.py#L30). `own_fields` might seem an easy way for foot shooting, but may also play an important role in designing safer abstractions.

With this reasoning, these changes look good to me.

Enet4 (approved):

PR #1829, opened 2021-04-16
Add IndexIVFScalarQuantizer for c_api

ava57r:
This PR add impl IndexIVFScalarQuantizer in c_api interface

### DIFF
--- a/c_api/IndexScalarQuantizer_c.cpp
+++ b/c_api/IndexScalarQuantizer_c.cpp
@@ -15,6 +15,7 @@
 
 using faiss::Index;
 using faiss::IndexScalarQuantizer;
+using faiss::IndexIVFScalarQuantizer;
 
 DEFINE_DESTRUCTOR(IndexScalarQuantizer)
 DEFINE_INDEX_DOWNCAST(IndexScalarQuantizer)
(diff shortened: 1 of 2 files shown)


c_api/IndexScalarQuantizer_c.cpp
```diff
@@ -42,3 +43,80 @@ int faiss_IndexScalarQuantizer_new_with(
...
+        FaissIndex* quantizer,
+        size_t d,
+        size_t nlist,
+        FaissQuantizerType qt,
+        FaissMetricType metric,
+        int encode_residual) {
+    try {
+        auto q = reinterpret_cast<Index*>(quantizer);
+        auto mt = static_cast<faiss::MetricType>(metric);
+        auto er = (bool)encode_residual;
```
Enet4, line 91: Prefer more conservative casts constructs from C++ over C-style casts.
Not that it would be necessary here, as `int` converts implicitly to a `bool` once passed to the constructor.
```suggestion
        auto er = static_cast<bool>(encode_residual);
```

Enet4 (commented):
Other than the suggested change, this seems OK.

Enet4 (approved):

ava57r:

Issue #1831, opened 2021-04-19
Different filename for c_api lib when opt level = avx2


ava57r:
# Summary

Hello.
When opt level == avx2
c_api lib builds in `libfaiss_c.so` with link `libfaiss_avx2.so`

When opt level == general
c_api lib builds in `libfaiss_c.so` with link `libfaiss.so`

Add different target for` c_api lib`
opt level == `avx2`
`libfaiss_avx2_c.so` with link `libfaiss_avx2.so`

# Platform

OS: Linux 

Faiss version: v1.7.0

Running on:
- CPU

Interface: 
-  C++

References:
#1783 
https://github.com/Enet4/faiss-rs/issues/19

Enet4:
As I stated in Enet4/faiss-rs#19, I would prefer the building process of the C API to result in a library always named `faiss_c` (libfaiss_c.so on Linux), regardless of the considered instruction set extensions or features.

ava57r:
> As I stated in [Enet4/faiss-rs#19](https://github.com/Enet4/faiss-rs/issues/19), I would prefer the building process of the C API to result in a library always named `faiss_c` (libfaiss_c.so on Linux), regardless of the considered instruction set extensions or features.

I agree. 
Thanks.

PR #1838, opened 2021-04-22
Fixed building faiss_c when OPT_LEVEL=avx2

### LINKED ISSUE #1836: C_API not supports avx2 OPT_LEVEL
# Summary

C_API builds without support `avx2` OPT_LEVEL.

# Platform

OS: Linux

Faiss version: bde7c0027191f29c9dadafe4f6e68ca0ee31fb30

Installed from: source

Faiss compilation options:

```
cmake -B $BUILD_BUILD \
-DBUILD_TESTING=OFF \
-DFAISS_ENABLE_GPU=OFF \
-DFAISS_OPT_LEVEL=avx2 \
-DFAISS_ENABLE_C_API=ON \
-DFAISS_ENABLE_PYTHON=OFF \
-DBUILD_SHARED_LIBS=ON \
-DCMAKE_BUILD_TYPE=Release .
```

Running on:
CPU

Interface: 
C++

# Reproduction instructions

1. I use this commit bde7c0027191f29c9dadafe4f6e68ca0ee31fb30
2. build from source `FAISS_OPT_LEVEL=avx2`
3. I have 

`libfaiss_avx2.so`
`libfaiss.so`
`libfaiss_c.so`

and `libfaiss_c.so` link with `libfaiss.so`

```console
ldd ./build/c_api/libfaiss_c.so 
	...
	libfaiss.so =>.../libfaiss.so
	...
```

I builded `v1.7.0` with the same options.

```console
ldd ./libfaiss_c-avx2-v1.7.0.so 
	...
	libfaiss_avx2.so => .../libfaiss_avx2.so
	...
```

I think it happened after this commit c65f6705236570c9aac365c68e3cc5da2343c888

ava57r:
Fixes #1836


Enet4:
I am confused about the recent decision to create different library build targets for the instruction set characteristics. Assuming that those `faiss_c` and `faiss_avx2_c` have the same set of exported symbols when built by the same machine, it is strange that we're contemplating them as different targets resulting in library files with different names as if the difference was more than implementation details.

I also expressed this concern in #1831 and Enet4/faiss-rs#19, and I would prefer the C API not to be affected by this replication, which so far seems only necessary for delivering "fat" binaries for conda packages. Rather, we would have a single `faiss_c` that would link against a valid native `faiss` library (regardless of which CPU capabilities are used).

ava57r:
Oh, Yes. You are right.
I copied from the parent cmake configuration.

mdouze:
@beauby could you take a look?

ava57r:
Could you include this fix in `v1.7.1`? issue #1915.

Enet4:
As v1.7.1 is already tagged in this repository, I suspect this may have to be delayed to a later version. This can be mitigated by renaming libfaiss_avx2.so to libfaiss.so in the delivery/installation phase.

Still, it might be better to coordinate with #1877, as some of the decisions there might invalidate the need for these changes. In particular, if under a certain building configuration we can always ensure outputs named libfaiss.so or libfaiss.a, then building faiss_c would just work.

ava57r:
@beauby could you take a look? :-)

ava57r:
#1877 closed.

the library still does not support `avx2` instructions for `c_api` main branch.

Issue #1762, opened 2021-03-15
IndexFlatL2 performance degradation


Alexsandruss:
# Summary

IndexFlatL2 has performance degradation on large number of threads since e1adde0d84ece584f3b4d86db0b1532329f8cdb8 commit.
This degradation can be observed with [benchs/bench_index_flat.py](https://github.com/facebookresearch/faiss/blob/master/benchs/bench_index_flat.py).

## bench_index_flat benchmark aggregated results:
q=100
`56 threads` | `100 query size` | before degradation (698a4592e87920f036aa7a2d8a3a56e12387a8f0) | after degradation (e1adde0d84ece584f3b4d86db0b1532329f8cdb8) | master (6977d72b1d6fde2085505b99199c6f95a3b2f10a)
-- | -- | -- | -- | --
dimension | k | time, ms | time, ms | time, ms
16 | 1 | 0.767 | 2.631 | 2.894
16 | 10 | 0.834 | 3.004 | 3.194
16 | 100 | 0.949 | 7.237 | 7.108
32 | 1 | 0.297 | 2.642 | 2.744
32 | 10 | 0.369 | 3.009 | 3.05
32 | 100 | 0.491 | 7.218 | 7.002
64 | 1 | 0.317 | 2.668 | 2.772
64 | 10 | 0.372 | 3.044 | 3.069
64 | 100 | 0.5 | 7.341 | 7.089
128 | 1 | 0.466 | 2.734 | 2.837
128 | 10 | 0.529 | 3.104 | 3.14
128 | 100 | 0.649 | 7.431 | 7.178

q=10000
`56 threads` | `10000 query size` | before degradation (698a4592e87920f036aa7a2d8a3a56e12387a8f0) | after degradation (e1adde0d84ece584f3b4d86db0b1532329f8cdb8) | master (6977d72b1d6fde2085505b99199c6f95a3b2f10a)
-- | -- | -- | -- | --
dimension | k | time, s | time, s | time, s
16 | 1 | 0.008 | 0.181 | 0.19
16 | 10 | 0.009 | 0.212 | 0.214
16 | 100 | 0.019 | 0.529 | 0.52
32 | 1 | 0.009 | 0.167 | 0.135
32 | 10 | 0.01 | 0.189 | 0.157
32 | 100 | 0.02 | 0.508 | 0.513
64 | 1 | 0.01 | 0.174 | 0.146
64 | 10 | 0.011 | 0.197 | 0.166
[...]

PR #1840, opened 2021-04-22
Fix performance regression in ResultHandler

mdouze:
Summary:
This diff is related to

https://github.com/facebookresearch/faiss/issues/1762

The ResultHandler introduced for FlatL2 and FlatIP was not multithreaded. This diff attempts to fix that. To be verified if it is indeed faster.

Differential Revision: D27939173


mdouze:
lldb backtrace: 
https://gist.github.com/mdouze/d4f3bf7e2edec4ff207d4c869eacfd0c

Could be because an OMP parallel loop is called from several threads.

PR #1805, opened 2021-04-06
Parallelize add_with_id of IndexIVFFlat and IndexIVFFlatDedup

KinglittleQ:
This PR parallelized the `add_with_ids` methods of `IndexIVFFlat` and `IndexIVFFlatDedup`. Related to https://github.com/facebookresearch/faiss/issues/1617.

PR #1842, opened 2021-04-25
Using DirectMapAdd to fix IVFFLAT parallel adding

shengjun1985:
Signed-off-by: shengjun.li <shengjun.li@zilliz.com>

When `direct_map.type` is `DirectMap::Type::Array`, IVFFLAT parallel adding may fail here.
```C++
void DirectMap::add_single_id(idx_t id, idx_t list_no, size_t offset) {
    ...
    if (type == Array) {
        assert(id == array.size());
```

However, DirectMapAdd has solved it.

### DIFF
--- a/faiss/IndexIVFFlat.cpp
+++ b/faiss/IndexIVFFlat.cpp
@@ -50,6 +50,8 @@ void IndexIVFFlat::add_core(
 
     int64_t n_add = 0;
 
+    DirectMapAdd dm_adder(direct_map, n, xids);
+
 #pragma omp parallel reduction(+ : n_add)
     {
         int nt = omp_get_num_threads();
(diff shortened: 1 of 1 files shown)


KinglittleQ:
This is a bug I introduced in https://github.com/facebookresearch/faiss/pull/1805. I didn't aware that the inserted ids may not be in ascending order due to OMP parallel 😓 .

mdouze:
OK, what was wrong here? Is it when list_no = -1 or something more complex? 
It does not matter that the inserted ids are not in ascending order.

KinglittleQ:
@mdouze `DirectMap::add_single_id` always appends `list_no|offset` into `array` and requires `id` should be the same as the array size. `id` may exceed the array size if it come out of order.

https://github.com/facebookresearch/faiss/blob/31bd194e2e589baf29a2aef2810cf28b40610a1d/faiss/invlists/DirectMap.cpp#L85-L101

mdouze:
Thanks! this sounds good.

Issue #185, opened 2017-08-25
Interest in non-orthogonal codebooks?


una-dinosauria:
I'm wondering if there is interest from the maintainers (and the community) on implementing some non-orthogonal variants of product quantization into faiss?

I have observed that these methods tend to perform very well on features extracted from deep neural networks, and are also pretty well suited for dot product similarity search.

Some baselines that come to my mind are

* Residual Vector Quantization (RVQ) [[paper](http://www.mdpi.com/1424-8220/10/12/11259/htm)]
* Enhanced RVQ (RVQ + a few rounds of ICM) [[paper](https://link.springer.com/article/10.1007/s00530-015-0470-9)]
* Optimized Tree Quantization (OTQ) [paper] -- I'm actually thinking of a variant that does not use Gurobi, similar to [this](https://github.com/una-dinosauria/local-search-quantization/blob/master/src/encodings/encode_chain.jl). [[paper](http://www.cv-foundation.org/openaccess/content_cvpr_2015/papers/Babenko_Tree_Quantization_for_2015_CVPR_paper.pdf)]
* Local Search Quantization (LSQ) [paper] -- Iterated local search with ICM. [[paper](https://www.cs.ubc.ca/~julm/papers/eccv16.pdf)]
* Some other improvements to LSQ that are currently under review (will post later)

I have implemented some of these on Julia before, and ported LSQ to the GPU (e.g. [here](https://github.com/una-dinosauria/local-search-quantization)). Please do let me know if there is interest for this; since the licence has been relaxed I'd be happy to contribute some of my own research to the library.

mdouze:
Hi, 
We have no plans of adding non-orthogonal quantization to Faiss, but we would be very grateful if you could implement an option of your choosing. 
If you plan to do this, I would suggest that you look around in the library's structure and propose an index structure (ie a .h file) with the required methods and fields, so that we can comment on it before you spend a lot of time in the implementation.
Thanks!

una-dinosauria:
I think we can close this one since #1862 has landed.

PR #1862, opened 2021-05-04
Add LocalSearchQuantizer

KinglittleQ:
This PR implemented LSQ/LSQ++, a vector quantization technique described in the following two papers:

1. Revisiting additive quantization
2. LSQ++: Lower running time and higher recall in multi-codebook quantization

Here is a benchmark running on SIFT1M for 64 bits encoding:
```
===== lsq:
        mean square error = 17335.390208
        training time: 312.729779958725 s
        encoding time: 244.6277096271515 s
===== pq:
        mean square error = 23743.004672
        training time: 1.1610801219940186 s
        encoding time: 2.636141061782837 s
===== rq:
        mean square error = 20999.737344
        training time: 31.813055515289307 s
        encoding time: 307.51959800720215 s
```

Changes:

1. Add LocalSearchQuantizer object
2. Fix an out of memory bug in ResidualQuantizer
3. Add a benchmark for evaluating quantizers
4. Add tests for LocalSearchQuantizer


faiss/impl/ResidualQuantizer.cpp
```diff
@@ -328,22 +329,44 @@ void ResidualQuantizer::compute_codes(
...
-            x,
-            max_beam_size,
-            codes.data(),
-            residuals.data(),
-            distances.data());
-
-    // pack only the first code of the beam (hence the ld_codes=M *
-    // cur_beam_size)
-    pack_codes(n, codes.data(), codes_out, M * max_beam_size);
+    // split into chunks to avoid allocating too large memory
```
mdouze, line 332: could you please remove this chunking logic? 
I have a PR in the making that does exactly that.

faiss/impl/ResidualQuantizer.h
```diff
@@ -119,7 +121,8 @@ struct ResidualQuantizer {
             int new_beam_size,
             int32_t* new_codes,
             float* new_residuals = nullptr,
-            float* new_distances = nullptr) const;
+            float* new_distances = nullptr,
+            bool verb = false) const;
```
mdouze, line 125: please call it verbose for consistency with other uses of the flag

faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -0,0 +1,726 @@
...
+void fvec_add(size_t d, const float* a, const float* b, float* c) {
+    for (size_t i = 0; i < d; i++) {
+        c[i] = a[i] + b[i];
+    }
+}
+
+void fmat_inverse(float* a, int n) {
+    int* ipiv = new int[n];
+    int lwork = n * n;
+    float* workspace = new float[lwork];
```
mdouze, line 93: please use std::vector or unique_ptr<float []> so that there is no mem leak if an exception is raised.

mdouze (commented):
Thank you very much for this contribution. It is extremely valuable and I never took the time to work on it. 
Please address the small comments below.

mdouze:
Flagging @una-dinosauria in case she is still following this project ;-) 

The discussions around this date back to issue #185

una-dinosauria:
> Flagging @una-dinosauria in case she is still following this project ;-)

Thanks for tagging me! I'll try to have a look this weekend 😄

faiss/impl/ResidualQuantizer.cpp
KinglittleQ: As long as https://github.com/facebookresearch/faiss/pull/1865 is merged, I will remove it and update this branch to make sure no conflicts.

KinglittleQ:
@una-dinosauria By the way, while contributing to this PR, I found that setting the regularization factor lambda to 1e-4 as stated in the paper probably would lead to some unstable results. So I set it to 0.01 instead. Have you encountered the same problem?

KinglittleQ:
@mdouze Updated.

An abstract class named AdditiveQuantizer is added as a base class of ResidualQuantizer and LocalSearchQuantizer.

KinglittleQ:
@una-dinosauria I did not develop in docker. Maybe you could try `docker pull beauby/faiss-circleci:cpu`. It is a container used in CircleCI jobs.

una-dinosauria:
Thank you @KinglittleQ! I was able to build a container following the example under the `.circleci` folder and the ci commands. I am having a look at the code now and running some tests locally. Will get back to you in the next few days.

faiss/impl/AdditiveQuantizer.cpp
```diff
@@ -0,0 +1,103 @@
...
+        tot_bits += nbit;
+        if (nbit % 8 != 0) {
+            is_byte_aligned = false;
+        }
+    }
+    // convert bits to bytes
+    code_size = (tot_bits + 7) / 8;
+}
+
+void AdditiveQuantizer::train(size_t n, const float* x) {
```
mdouze, line 55: Please declare this as =0 in the .h file (then the compiler will make sure people don't instantiate the class).

faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -0,0 +1,690 @@
...
+    std::vector<float> workspace(lwork);
+
+    sgetrf_(&n, &n, a, &n, ipiv.data(), &info);
+    FAISS_THROW_IF_NOT(info == 0);
+    sgetri_(&n, a, &n, ipiv.data(), workspace.data(), &lwork, &info);
+    FAISS_THROW_IF_NOT(info == 0);
+}
+
+void random_int32(
+        std::vector<int32_t>* x,
```
mdouze, line 102: Please use 
`std::vector<int32_t> & x`

faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    // add a regularization term to B'B
+#pragma omp parallel for
+    for (int64_t i = 0; i < M * K; i++) {
+        bb[i * (M * K) + i] += lambd;
+    }
+
+    // compute (B'B)^(-1)
+    fmat_inverse(bb.data(), M * K); // [M*K, M*K]
```
mdouze, line 348: It is usually faster to solve a least squares system with SGELS than to compute the inverse and multiply with the right-hand term 

https://www.netlib.org/lapack/lug/node27.html

faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -0,0 +1,690 @@
...
+                int32_t code2 = codes[i * M + m2];
+                int32_t idx2 = m2 * K + code2;
+                bb[idx1 * M * K + idx2] += 1;
+                bb[idx2 * M * K + idx1] += 1;
+            }
+        }
+    }
+
+    // add a regularization term to B'B
+#pragma omp parallel for
```
mdouze, line 342: parallel for not necessary here. In general it is better to use parallel for only when we are sure that the loop takes > 1M operations, otherwise the overhead makes it slower (not sure about the exact break-down point)

faiss/impl/LocalSearchQuantizer.cpp
mdouze: @una-dinosauria thanks for the explanation!

KinglittleQ:
@mdouze Updated. Could you take a look again?

mdouze:
LGTM, importing into FB build repo.

PR #1878, opened 2021-05-12
Implement peel loop for fvec_L2sqr, fvec_inner_product, and fvec_norm_L2sqr on aarch64

vorj:
In the current `faiss` implementation for x86, `fvec_L2sqr` , `fvec_inner_product` , and `fvec_norm_L2sqr` are [optimized for any dimensionality](https://github.com/facebookresearch/faiss/blob/e86bf8cae1a0ecdaee1503121421ed262ecee98c/faiss/utils/distances_simd.cpp#L404-L432).

On the other hand, the functions for aarch64 are optimized [**only** if `d` is multiple for 4](https://github.com/facebookresearch/faiss/blob/e86bf8cae1a0ecdaee1503121421ed262ecee98c/faiss/utils/distances_simd.cpp#L583-L584); thus, they are not much fast for vectors with `d % 4 != 0` . 
This PR has accelerated the above three functions for any input size on aarch64.

Kind regards,

peel-loop
- Evaluated on an AWS EC2 ARM instance (c6g.4xlarge)
- sift1m127 is the dataset with dropped trailing elements of sift1m
    - Therefore, the vector length of sift1m127 is 127 that is not multiple of 4
    - "optimized" runs 2.45-2.77 times faster than "original" with sift1m127
- Two methods, "original" and "optimized", are expected to achieve the same level of performance for sift1m
    - And actually there is almost no significant difference


faiss/utils/distances_simd.cpp
```diff
@@ -580,42 +580,60 @@ float fvec_inner_product(const float* x, const float* y, size_t d) {
...
 float fvec_L2sqr(const float* x, const float* y, size_t d) {
-    if (d & 3)
-        return fvec_L2sqr_ref(x, y, d);
-    float32x4_t accu = vdupq_n_f32(0);
-    for (size_t i = 0; i < d; i += 4) {
+    float32x4_t accux4 = vdupq_n_f32(0);
+    const size_t d_simd = d - (d & 3);
+    size_t i;
+    for (i = 0; i < d_simd; i += 4) {
         float32x4_t xi = vld1q_f32(x + i);
```
mdouze, line 587: Looks good. TBH I thought that ARM SIMD accesses needed to be aligned.

mdouze (approved):
The diff looks good. It's unfortunate that we don't have ARM CI (yet).

Issue #1880, opened 2021-05-12
Problems with SIMD when compiling for ARM (Apple M1)


vsolovyov:
When I try to compile Faiss, it complains in this way:

```
In file included from /Users/murkt/dev/faiss/faiss/impl/pq4_fast_scan_search_1.cpp:11:
In file included from /Users/murkt/dev/faiss/faiss/impl/simd_result_handlers.h:15:
In file included from /Users/murkt/dev/faiss/faiss/utils/simdlib.h:23:
/Users/murkt/dev/faiss/faiss/utils/simdlib_neon.h:256:54: error: argument to '__builtin_neon_vshrq_n_v' must be a constant integer
                data, [shift](uint16x8_t a) { return vshrq_n_u16(a, shift); })};
                                                     ^              ~~~~~
/Library/Developer/CommandLineTools/usr/lib/clang/12.0.5/include/arm_neon.h:25132:24: note: expanded from macro 'vshrq_n_u16'
  __ret = (uint16x8_t) __builtin_neon_vshrq_n_v((int8x16_t)__s0, __p1, 49); \
                       ^                                         ~~~~
```

There are a bunch of them, each time `simdlib_neon.h` is included.

# Platform

OS: macOS 11.3
Faiss version: b4c320a671e58252b229f12459f2d9c285e8b535
Clang: 12.0.5
SWIG: 4.0.2b
cmake: 3.20.1

Compiling Faiss with these:

```
cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=OFF -DCMAKE_BUILD_TYPE=Release -DFAISS_ENABLE_C_API=ON -DPython_EXECUTABLE=`which python3`
make -C build -j faiss
```

Code in `faiss/utils/simdlib_neon.h` seems to already be aware of the issue that `shift` argument must be a constant integer:

```
    // shift must be known at compile time
    simd16uint16 operator>>(const int shift) const {
        return simd16uint16{detail::simdlib::unary_func(
                data, [shift](uint16x8_t a) { return vshrq_n_u16(a, shift); })};
    }

    // shift must be known at compile time
    simd16uint16 operator<<(const int shift) const {
        return simd16uint16{detail::simdlib::unary_func(
                data, [shift](uint16x8_t a) { return vshlq_n_u16(a, shift); })};
    }
```

[...]

PR #1882, opened 2021-05-13
Workaround for vshl/vshr on aarch64 GCC

vorj:
related: #1815,  #1880

`vshl` / `vshr` of ARM NEON requires immediate (compiletime constant) value as shift parameter.
However, the implementations of those intrinsics on GCC can receive runtime value.
Current faiss implementation depends on this, so some correct-behavioring compilers like Clang can't build faiss for aarch64.
This PR fix this issue; thus faiss applied this PR can be built with Clang for aarch64 machines like M1 Mac.

### DIFF
--- a/faiss/utils/simdlib_neon.h
+++ b/faiss/utils/simdlib_neon.h
@@ -181,6 +181,16 @@ static inline uint32_t cmp_xe32(
     return d0_mask | static_cast<uint32_t>(d1_mask) << 16;
 }
 
+template <std::uint8_t Shift>
+static inline uint16x8_t vshlq(uint16x8_t vec) {
+    return vshlq_n_u16(vec, Shift);
+}
+
+template <std::uint8_t Shift>
+static inline uint16x8_t vshrq(uint16x8_t vec) {
+    return vshrq_n_u16(vec, Shift);
+}
+
 } // namespace simdlib
 
 } // namespace detail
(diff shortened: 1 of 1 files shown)


vsolovyov:
Wow, you're fast :) With this fix it does compile for me, and tests pass 100%.
Clang 12.0.5, macOS 11.3, M1 Mac

mdouze:
Cool if it works. Hopefully the optimizer will be able to remove the switch at compile time.

faiss/utils/simdlib_neon.h
```diff
@@ -252,14 +262,108 @@ struct simd16uint16 {
...
+                return simd16uint16{detail::simdlib::unary_func(
+                        data, detail::simdlib::vshrq<13>)};
+            case 14:
+                return simd16uint16{detail::simdlib::unary_func(
+                        data, detail::simdlib::vshrq<14>)};
+            case 15:
+                return simd16uint16{detail::simdlib::unary_func(
+                        data, detail::simdlib::vshrq<15>)};
+            default:
+                return *this;
```
beauby, line 312: This assumes `(shift <= 1 || shift > 15)` means shift == 0, right? If so, we should probably have:
```
case 0:
  return *this;
default:
  FAISS_THROW_FMT("Invalid shift %d", shift);
```

vorj: I've fixed them.

Issue #1876, opened 2021-05-11
Merge d2c04d093576d1d5fe26de28352b857d7236a447 before next release.


h-vetinari:
Not sure what the timeline is for 1.7.1 or 1.8.0, but it's been a while since 1.7.0, and I'd just like to make sure that https://github.com/facebookresearch/faiss/pull/1610/commits/d2c04d093576d1d5fe26de28352b857d7236a447 gets merged before the next release (doesn't need the whole PR, just as a single commit is fine), otherwise windows GPU support is not possible - or rather, continues to require carrying patches...

I can open a PR with that commit if that's helpful.

mdouze:
We plan to do 1.7.1 ASAP, the current blocker is issues with anaconda.

PR #1884, opened 2021-05-14
Fix Conda CI

vorj:
Currently CI jobs using conda are failed due to conflict packages.
This PR fixes this.

- use newer `numpy` to build `faiss-cpu`
- install `pytorch` when testing `faiss-cpu`
    - to find correct `pytorch` package, `pytorch` channel is set at `conda build`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,9 +93,8 @@ jobs:
           name: Conda build
           command: |
             conda install -y -q conda-build
-            conda install -c pytorch pytorch
             cd conda
-            conda build faiss --python 3.7
+            conda build faiss --python 3.7 -c pytorch
 
   build_osx:
     macos:
@@ -158,9 +157,8 @@ jobs:
           name: Build/test
           command: |
             conda install conda-build
-            conda install -c pytorch pytorch
             cd conda
-            conda build faiss --python 3.7
+            conda build faiss --python 3.7 -c pytorch
 
   build_linux_gpu:
     machine:
--- a/conda/faiss/meta.yaml
+++ b/conda/faiss/meta.yaml
@@ -65,7 +65,7 @@ outputs:
         - make  # [not win]
       host:
         - python {{ python }}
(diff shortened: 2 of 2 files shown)


vorj:
The job for Windows will be fixed by #1876 .

conda/faiss/meta.yaml
```diff
@@ -65,7 +65,7 @@ outputs:
         - make  # [not win]
       host:
         - python {{ python }}
-        - numpy 1.11.*
+        - numpy >=1.11.*
```
beauby, line 68: We need to build against numpy `1.11`, as this is the earliest version of numpy we support, and we use the C API. If we relax the constraint here, packages generated will only be compatible with numpy versions newer than whichever conda-build ends up selecting at build time.

.circleci/config.yml
```diff
@@ -93,9 +93,8 @@ jobs:
           name: Conda build
           command: |
             conda install -y -q conda-build
-            conda install -c pytorch pytorch
             cd conda
-            conda build faiss --python 3.7
+            conda build faiss --python 3.7 -c pytorch
```
beauby, line 97: These changes are still desirable. Would you mind rebasing your PR?

conda/faiss/meta.yaml
vorj: I see. I confirmed that you fixed this part on #1894, so I've reverted this change.

conda/faiss/meta.yaml
```diff
@@ -40,7 +40,7 @@ outputs:
       host:
         - mkl =2018
       run:
-        - mkl >=2018
+        - mkl >=2018,<2021
```
vorj, line 43: `mkl-2021.2.0-haa95532_296` doesn't contain `mkl_rt.dll` (it contains `mkl_rt.1.dll` instead). We need to constrain `mkl` as `<2021` if `faiss` should be built with `mkl=2018` (linked to `mkl_rt.dll`) .

Issue #1889, opened 2021-05-17
Range search on mac gives weird results


mdouze:
# Summary

# Platform

OS: mac os 

Faiss version: 1.7.0

Installed from: anaconda

Running on:
- [x ] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [x ] Python

# Reproduction instructions
```
import numpy as np
import faiss

d = 256
nq = 16
nb = 1000000

index = faiss.IndexFlatL2(d)

xb = np.zeros((nb, d), dtype="float32")
index.add(xb)

xq = np.zeros((nq, d), dtype="float32")
lims, D, I = index.range_search(xq, 1.0)

assert len(D) == len(xb) * len(xq)
```
In this case len(D) = 36M
Tests: 
- with OMP_NUM_THREADS=1 , len(D) is 1M (also bad)
- when nb is smaller, the bug disappears
- with varying number of threads, d and nq, the bug appears and disappears...

mdouze:
Fixed in 1.7.1

PR #1890, opened 2021-05-17
Range search bug

mdouze:
Just added a test to see if CI catches the bug.
See #1889

Test plan: PYTHONPATH=../build/faiss/python/build/lib python3 -m unittest test_index.TestLargeRangeSearch


mdouze:
During range search in a flat index, (exhaustive_L2sqr_seq and exhaustive_inner_product_seq) when running in multiple threads, the per-thread results are collected into RangeSearchPartialResult structures. 

When the computation is finished, they are aggregated into a RangeSearchResult. In the previous version of the code, this loop was nested into a second loop that is used to check for KeyboardInterrupts. Thus, at each iteration, the results were overwritten. 

The fix removes the outer loop. It is most likely useless anyways because the sequential code is called only for a small number of queries, for a larger number the BLAS version is used.

mdouze:
Unclear why this is still open: the fix landed.

PR #1894, opened 2021-05-18
Fix conda recipes.

beauby:


h-vetinari:
This seems like a step back... Did you have issues with resolvability, or is an old mkl build somehow faster for your usecase...? 🤔

beauby:
This only pins the host version of MKL, not the one used at runtime.

h-vetinari:
OK, I overlooked that the mkl-changes are only for host; but than also... numpy 1.11? 😄
I'm surprised if you still get that for relevant python versions? Conda-forge recently moved from 1.16 to 1.17 as the version used at build time for all packages.

Anyway, this was just a drive-by comment... 🙃

PR #1900, opened 2021-05-19
Avoid OOM in Linux CPU CI jobs.

beauby:


mdouze:
3 seems a bit small but I assume the CPU part compiles fast anyways.

PR #1914, opened 2021-05-27
Add ARM to CI

KinglittleQ:
This PR added a CI job for ARM.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,20 +172,27 @@ jobs:
         default: generic
     environment:
       OMP_NUM_THREADS: 10
+      CONDA_HOME: /home/circleci/miniconda3
+      PYTHON: /home/circleci/miniconda3/bin/python
     steps:
       - checkout
       - run:
           name: Install dependencies
           command: |
-            sudo apt-get update && sudo apt-get install -y swig python3
-            pip3 install numpy==1.19 scipy torch pytest cmake
+            sudo apt-get update && sudo apt-get install -y swig
+            wget https://repo.anaconda.com/miniconda/Miniconda3-py39_4.9.2-Linux-aarch64.sh
+            bash Miniconda3-py39_4.9.2-Linux-aarch64.sh -b -p $CONDA_HOME
+            pip3 install cmake
+            $CONDA_HOME/bin/conda install -y numpy scipy
+            $CONDA_HOME/bin/pip install pytest torch
       - run:
           name: Build faiss library
           command: |
             cmake -B build -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=OFF \
                   -DFAISS_OPT_LEVEL=<< parameters.opt_level >> \
                   -DFAISS_ENABLE_C_API=ON \
-                  -DCMAKE_BUILD_TYPE=Release .
+                  -DCMAKE_BUILD_TYPE=Release \
+                  -DPython_EXECUTABLE=$PYTHON .
             make -C build -j3 faiss
       - run:
           name: Test faiss library
@@ -198,14 +205,14 @@ jobs:
           command: |
             make -C build -j3 swigfaiss
             cd build/faiss/python
-            python3 setup.py build
+            $PYTHON setup.py build
       - run:
           name: Test python extension
           command: |
             export PYTHONPATH="$(ls -d ./build/faiss/python/build/lib*/)"
-            python3 -c "import faiss; assert 'NEON' in faiss.get_compile_options()"
-            pytest --junitxml=test-results/pytest/results.xml tests/test_*.py
-            pytest --junitxml=test-results/pytest/results-torch.xml tests/torch_*.py
+            $PYTHON -c "import faiss; assert 'NEON' in faiss.get_compile_options()"
+            $PYTHON -m pytest --junitxml=test-results/pytest/results.xml tests/test_*.py
+            $PYTHON -m pytest --junitxml=test-results/pytest/results-torch.xml tests/torch_*.py
       - store_test_results:
           path: test-results
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,7 +178,7 @@ jobs:
           name: Install dependencies
           command: |
             sudo apt-get update && sudo apt-get install -y swig python3
-            pip3 install numpy scipy torch pytest cmake
+            pip3 install numpy==1.19 scipy torch pytest cmake
       - run:
           name: Build faiss library
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,6 +162,57 @@ jobs:
             cd conda
             conda build faiss --python 3.7
 
+  build_arm:
+    machine:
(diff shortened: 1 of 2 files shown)


KinglittleQ:
@mdouze @beauby @vorj Could you 
There are two failed tests in ARM:

```
=================================== FAILURES ===================================
_______________ TestOrthognalReconstruct.test_recons_orthonormal _______________

self = <test_build_blocks.TestOrthognalReconstruct testMethod=test_recons_orthonormal>

    def test_recons_orthonormal(self):
        lt = faiss.LinearTransform(20, 10, True)
        rs = np.random.RandomState(10)
        A, _ = np.linalg.qr(rs.randn(20, 20))
        A = A[:10].astype('float32')
        faiss.copy_array_to_vector(A.ravel(), lt.A)
        faiss.copy_array_to_vector(rs.randn(10).astype('float32'), lt.b)
    
        lt.set_is_orthonormal()
        lt.is_trained = True
        assert lt.is_orthonormal
    
        x = rs.rand(30, 20).astype('float32')
        xt = lt.apply_py(x)
        xtt = lt.reverse_transform(xt)
        xttt = lt.apply_py(xtt)
    
        err = ((xt - xttt)**2).sum()
    
>       self.assertGreater(1e-5, err)
E       AssertionError: 1e-05 not greater than nan

tests/test_build_blocks.py:112: AssertionError
___________________________ TestPCAWhite.test_white ____________________________

self = <test_index_composite.TestPCAWhite testMethod=test_white>

    def test_white(self):
    
        # generate data
        d = 4
        nt = 1000
        nb = 200
        nq = 200
    
        # normal distribition
        x = faiss.randn((nt + nb + nq) * d, 1234).reshape(nt + nb + nq, d)
    
        index = faiss.index_factory(d, 'Flat')
    
        xt = x[:nt]
        xb = x[nt:-nq]
        xq = x[-nq:]
    
        # NN search on normal distribution
        index.add(xb)
        Do, Io = index.search(xq, 5)
    
        # make distribution very skewed
        x *= [10, 4, 1, 0.5]
        rr, _ = np.linalg.qr(faiss.randn(d * d).reshape(d, d))
        x = np.dot(x, rr).astype('float32')
    
        xt = x[:nt]
        xb = x[nt:-nq]
        xq = x[-nq:]
    
        # L2 search on skewed distribution
        index = faiss.index_factory(d, 'Flat')
    
        index.add(xb)
        Dl2, Il2 = index.search(xq, 5)
    
        # whiten + L2 search on L2 distribution
        index = faiss.index_factory(d, 'PCAW%d,Flat' % d)
    
        index.train(xt)
        index.add(xb)
        Dw, Iw = index.search(xq, 5)
    
        # make sure correlation of whitened results with original
        # results is much better than simple L2 distances
        # should be 961 vs. 264
>       assert (faiss.eval_intersection(Io, Iw) >
                2 * faiss.eval_intersection(Io, Il2))
E       AssertionError: assert 0 > (2 * 0)
E        +  where 0 = <function eval_intersection at 0xffffb7231550>(array([[ 42,  44,  76,   6, 169],\n       [ 58, 157,   1,  37,   8],\n       [128,  85,  71, 120, 141],\n       [181,  72... 106,  14, 165],\n       [168,  50, 196,  36, 123],\n       [ 61,  53,  59,  39,  40],\n       [ 81,  77, 185, 142, 129]]), array([[-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n  ...],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1]]))
E        +    where <function eval_intersection at 0xffffb7231550> = faiss.eval_intersection
E        +  and   0 = <function eval_intersection at 0xffffb7231550>(array([[ 42,  44,  76,   6, 169],\n       [ 58, 157,   1,  37,   8],\n       [128,  85,  71, 120, 141],\n       [181,  72... 106,  14, 165],\n       [168,  50, 196,  36, 123],\n       [ 61,  53,  59,  39,  40],\n       [ 81,  77, 185, 142, 129]]), array([[-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n  ...],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1],\n       [-1, -1, -1, -1, -1]]))
E        +    where <function eval_intersection at 0xffffb7231550> = faiss.eval_intersection

tests/test_index_composite.py:260: AssertionError
```

KinglittleQ:
Looks like NEON works well in faiss.

KinglittleQ:
There are two failed tests in ARM.

KinglittleQ:
@mdouze @beauby @vorj Do you have any ideas of what caused the failed tests?

beauby (approved):

vorj:
@KinglittleQ I confirmed that the two errors reproduced on an AWS c6g.4xlarge instance.
However, as I confirmed, those tests passes on a Raspberry Pi 4 Model B and an NVIDIA Jetson AGX Xavier.
Currently I haven't found the cause of the errors, but:

- It seems that the Circle CI uses AWS as backend
- The cause may be something like below (just expectation):
    - The Neoverse N1 micro architecture or the Graviton 2 ARM processor have some critical bug
        - If so, we can't fix this right now...
    - The BLAS or LAPACK libraries for Neonverse N1 have some critical bug
        - For example, the compilers output execution binaries for Neoverse N1 with incorrect optimization, and so on.
        - I think that this pattern may be not. I installed `libblas-dev` and `liblapack-dev` using `apt` , so the binaries should not be specialized to Neoverse N1
    - Compilers optimize faiss incorrect
    - Faiss contains some undefined/unspecified behavior codes, then the execution binary behaves different on AWS and other
    - Those tests expect some implicit assumption about the behavior of processors, and Graviton 2 doesn't satisfied that (but the behavior is also well-formed)
        - In this pattern, the tests need to be fixed to stop expecting the behavior.

mdouze:
Thanks for this diff! 

These tests are reliable and the last time I tested them on an AWS ARM instance they were working. 
The two tests call some Lapack functions (QR decomposition and eigenvalues), so the error probably comes from Lapack. 
IIRC I was using the BLAS and Lapack from a miniconda distribution. 
So probably we need to force compilation with a proper Lapack implem for ARM.

KinglittleQ:
There is a bug in NumPy >= 1.20 installed by pip. But the conda version works well.

To reproduce the bug in NumPy >= 1.20:
```
import numpy as np
rs = np.random.RandomState(10)
x = rs.randn(20, 20)
q, _ = np.linalg.qr(x)

assert not np.isnan(np.min(q))
```

KinglittleQ:
Switching to NumPy 1.19 could solve the above problem but Pytorch does not support NumPy 1.19. 

The current solution is using an Anaconda Python interpreter. This should fix the problem.

PR #1945, opened 2021-06-14
C_API: Improve PreTransformIndex

ava57r:


mdouze:
@Enet4 does this work?

Enet4 (approved):
This appears to be an accurate extension to the C API around the `PreTransformIndex` construct in the native library. LGMT

PR #1961, opened 2021-06-25
Support for additive quantizer search

mdouze:
Summary:
This diff implements LUT-based search for additive quantizers.
It also further merges code for LSQ and the RedisualQuantizer.

This is WIP. What's missing:
- IVF indexes
- i/o
- index factory
- lots of evaluation...

Differential Revision: D29395079


KinglittleQ:
Looks pretty good! Would you mind if I fork this and add IVF to it?

mdouze:
I’m working on IVF. I would appreciate a critical look on the PR when ready ;-)

Matthijs

On 3 Jul 2021, at 04:20, Chengqi Deng ***@***.***> wrote:

﻿

Looks pretty good! Would you mind if I fork this and add IVF to it?

KinglittleQ:
Cool, looking forward to your update😁

faiss/impl/ResidualQuantizer.cpp
```diff
@@ -354,9 +377,16 @@ void ResidualQuantizer::compute_codes(
             residuals.data(),
             distances.data());
 
+    std::vector<float> norms;
+    if (search_type == ST_norm_float || search_type == ST_norm_qint8 ||
+        search_type == ST_norm_qint4) {
+        norms.resize(n);
+        fvec_norms_L2sqr(norms.data(), x, d, n);
```
KinglittleQ, line 384: This may lead to a high quantization error of vectors although the norm is accurate. A more general way is to store the norm of reconstructed vectors.

faiss/impl/AdditiveQuantizer.cpp
```diff
@@ -75,25 +96,94 @@ void AdditiveQuantizer::set_derived_values() {
...
+    return xi < 0 ? 0 : xi > 15 ? 15 : xi;
+}
+
+float decode_qint8(uint8_t i, float amin, float amax) {
+    return (i + 0.5) / 256 * (amax - amin) + amin;
+}
+
+float decode_qint4(uint8_t i, float amin, float amax) {
+    return (i + 0.5) / 16 * (amax - amin) + amin;
+}
```
KinglittleQ, line 144: We could replace these with a non-uniform quantization using 1-D K-means. But I think it would not improve the performance too much for 8 bits code. I can do it in a follow-up PR.

KinglittleQ (commented):
Some small suggestions.

faiss/impl/ResidualQuantizer.cpp
mdouze: I think that storing the original norms gives better accuracy. Will redo experiments to confirm.

mdouze: OK you're right, encoding the compressed norms is more accurate.

PR #1969, opened 2021-06-29
Generate python docstrings from doxygen comments.

beauby:


mdouze:
Could you add a test in 

https://github.com/facebookresearch/faiss/blob/master/tests/test_documentation.py

that checks if the doxygen comments are there?

beauby:
The docker image for GPU tests has a stale version of swig, which does not support the `-doxygen` flag. I'll update it before we merge this.

PR #1975, opened 2021-07-03
Update codebooks with double type

KinglittleQ:
## Description

The process of updating the codebook in LSQ may be unstable if the data is not zero-centering. This diff fixed it by using `double` instead of `float` during codebook updating. This would not affect the performance since the update process is quite fast.

Users could switch back to `float` mode by setting `update_codebooks_with_double = False`

## Changes

1. Support `double` during codebook updating. 
2. Add a unit test.
3. Add `__init__.py` under `contrib/` to avoid warnings.

### DIFF
--- a/benchs/bench_quantizer.py
+++ b/benchs/bench_quantizer.py
@@ -50,7 +50,7 @@ nbits = 8
 
 if 'lsq' in todo:
     lsq = faiss.LocalSearchQuantizer(d, M, nbits)
-    lsq.log_level = 2  # show detailed training progress
+    lsq.verbose = True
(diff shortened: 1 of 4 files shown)


mdouze:
Can you say a bit more about what datasets this affects? eg. is this a problem for SIFT1M that is not 0-centrered?

KinglittleQ:
@mdouze  Yes, this affects SIFT1M. A simple way to reproduce the problem is to set `M = 16` and `nbits = 4` in `benchs/bench_quantizer.py`. Another workaround is to increase `lsq.lambd` to 0.1 but this needs tuning.

PR #1982, opened 2021-07-13
[C_API] Improve meta indexes

ava57r:


mdouze:
LGTM, @Enet4 any objection?

Enet4 (approved):
Looks good to me as well. It is basically providing an API to index replicas by an instantiation of `IndexReplicas<Index*>`.

Issue #1835, opened 2021-04-21
Faiss assertion 'err__ == cudaSuccess' failed in void faiss::gpu::runTransposeAny - CUDA error 9 invalid configuration argument


cmacdonald:
# Summary

Faiss assertion 'err__ == cudaSuccess' failed in void faiss::gpu::runTransposeAny(faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, int, int, faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, cudaStream_t) [with T = float; int Dim = 3; cudaStream_t = CUstream_st*] at /root/miniconda3/conda-bld/faiss-pkg_1613235005464/work/faiss/gpu/utils/Transpose.cuh:207; details: CUDA error 9 invalid configuration argument

# Platform

OS: linux, kernel 3.10.0-1062.12.1.el7.x86_64; TITAN RTX GPU

Faiss version: 1.7.0

Installed from: Anaconda

Faiss compilation options: Anaconda build py3.7_h080d439_0_cuda10.2 and py3.7_hf4b0e32_0_cuda10.1

Running on:
- [ ] CPU
- [X] GPU

Interface: 
- [ ] C++
- [X] Python

# Reproduction instructions

 

We are indexing using [ColBERT](https://github.com/stanford-futuredata/ColBERT/), c.f. https://github.com/stanford-futuredata/ColBERT/blob/master/colbert/indexing/faiss_index.py#L41

The index has been trained, the error comes when adding the first bit of data to the index. The index type is IndexIVFPQ

I have tried two different Conda environments with different CUDA versions.

Faiss assertion 'err__ == cudaSuccess' failed in void faiss::gpu::runTransposeAny(faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, int, int, faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, cudaStream_t) [with T = float; int Dim = 3; cudaStream_t = CUstream_st*] at /root/miniconda3/conda-bld/faiss-pkg_1613235005464/work/faiss/gpu/utils/Transpose.cuh:207; details: CUDA error 9 invalid configuration argument

It is not the case that the GPU runs out of memory.

cmacdonald:
[...]

Issue #1771, opened 2021-03-20
CUDA error 9 invalid configuration argument when running demo_ivfpq_indexing_gpu and more


FjodBorg:
# Summary

I get some weird errors and segmentation faults when running the demos/tests. I've written all the steps i've made in the reproduction instructions. I'm using a 1650 GTX, cuda-11.2, python3.7 and openblas. The memory usage never goes above 1700MiB/3908MiB when running the tests.

# Platform

OS: Ubuntu 18.04

Faiss version: v1.7.0 with 5efe1a97323a3e327b9058d57a54d4469ef6baad applied

Installed from: source with cmake 3.18.6

Faiss compilation options: 
cmake -B build . -DFAISS_ENABLE_GPU=ON \
                -DFAISS_ENABLE_PYTHON=ON \
                -DCMAKE_BUILD_TYPE=Release \
                -DCUDAToolkit_ROOT=/usr/local/cuda-11.2 \
                -DCMAKE_CUDA_ARCHITECTURES="75;72" \
                -DPython_EXECUTABLE=/usr/bin/python3.7 \
Running on:
- [ ] CPU
- [X] GPU

Interface: 
- [x] C++
- [x] Python

# Reproduction instructions
This is the exact commands i used to install faiss. I had to rename ```/usr/bin/python``` to something else otherwise it would use python3.7 as executable and python2.7 as pythonInterp.
```
cd $HOME/repos
git clone https://github.com/facebookresearch/faiss
cd faiss/faiss
git checkout tags/v1.7.0
wget https://github.com/facebookresearch/faiss/pull/1245/commits/5efe1a97323a3e327b9058d57a54d4469ef6baad.diff --output-document=gpu_fix.patch
patch -p1 < gpu_fix.patch --force
cd $HOME/repos/faiss
mkdir build
PYTHON_WRONG=$(which python)
[ -f $PYTHON_WRONG ] && sudo mv $PYTHON_WRONG "${PYTHON_WRONG}.back"
PYTHON_BIN=$(which python3.7)
cmake -B build . -DFAISS_ENABLE_GPU=ON \
                -DFAISS_ENABLE_PYTHON=ON \
                -DCMAKE_BUILD_TYPE=Release \
                -DCUDAToolkit_ROOT=/usr/local/cuda \
                -DCMAKE_CUDA_ARCHITECTURES="75;72" \
                -DPython_EXECUTABLE=$PYTHON_BIN \
[...]

PR #1996, opened 2021-07-23
Possible fix for the GPU transpose utility

viclafargue:
I identified a bug in the GPU transpose utility and could create an attempt to fix it. The bug is visible when using the GPU approximate nearest neighbors method `IVFPQ` and is reproducible with the following code (requires `cuML`) :
```
import cupy as cp
from cuml.neighbors import NearestNeighbors

n_dims = 2
n_neighbors = 10
n_samples = 65_536
n_unknown = 10_000

index = cp.random.random((n_samples, n_dims))
query = cp.random.random((n_unknown, n_dims))

knn_model = NearestNeighbors(n_neighbors=n_neighbors,  algorithm='ivfpq')
knn_model.fit(index)
nearest_neighbors = knn_model.kneighbors(query)
```
This code throws the following error:
`Faiss assertion 'err__ == cudaSuccess' failed in void faiss::gpu::runTransposeAny(faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, int, int, faiss::gpu::Tensor<OtherT, OtherDim, true, int, faiss::gpu::traits::DefaultPtrTraits>&, cudaStream_t) [with T = float; int Dim = 3; cudaStream_t = CUstream_st*] at <...>/faiss/faiss/gpu/utils/Transpose.cuh:207; details: CUDA error 9 invalid configuration argument
Aborted (core dumped)`

The problem appears when the number of samples in the index is above 65535. The issue seems to stem from the fact that in preparation of the launch of the `transposeOuter` CUDA kernel, the y dimension of the grid of thread blocks is parametrized with the number of rows/samples. Indeed, this causes a problem as the maximum y-, or z-dimension of a grid of thread blocks is 65535 (see [CUDA compute compatibility technical specifications](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications)). In my understanding, this is what throws a `CUDA error 9 invalid configuration argument` during the launch of the CUDA kernel that follows.

Here is the part of the code that is concerned by the problem:
https://github.com/facebookresearch/faiss/blob/7c2d2388a492d65fdda934c7e74ae87acaeed066/faiss/gpu/utils/Transpose.cuh#L160-L175

Tagging issues:
https://github.com/rapidsai/cuml/issues/4020
https://github.com/facebookresearch/faiss/issues/1771
https://github.com/facebookresearch/faiss/issues/1835

### DIFF
--- a/faiss/gpu/utils/Transpose.cuh
+++ b/faiss/gpu/utils/Transpose.cuh
@@ -95,8 +95,8 @@ template <typename T, typename IndexT>
 __global__ void transposeOuter(const T* in,
                                T* out,
                                IndexT t1, IndexT t2, IndexT i1) {
-  IndexT gt1 = blockIdx.y;
-  IndexT gt2 = blockIdx.x;
+  IndexT gt1 = blockIdx.x;
+  IndexT gt2 = blockIdx.y;
 
   in += i1 * (gt1 * t2 + gt2);
   out += i1 * (gt2 * t1 + gt1);
@@ -157,7 +157,7 @@ void runTransposeAny(Tensor<T, Dim, true>& in,
       innerSize *= in.getSize(i);
     }
 
-    auto grid = dim3(in.getSize(1), in.getSize(0));
+    auto grid = dim3(in.getSize(0), in.getSize(1));
     int block = (innerSize < maxThreads) ? innerSize : maxThreads;
 
     if (totalSize <= (size_t) std::numeric_limits<int>::max()) {


wickedfoo:
Thanks for finding this!

Flipping the grid indices doesn't really fix the issue though, it would come up in other circumstances. We'd need to cap the y block dimension and add a loop to the kernel I think.

viclafargue:
> Thanks for finding this!
> 
> Flipping the grid indices doesn't really fix the issue though, it would come up in other circumstances. We'd need to cap the y block dimension and add a loop to the kernel I think.

Yes indeed, it looks like the issue could come back if somehow `in.getSize(1) > 65535` among others. This is just a little attempt to find a fix, but it's probably better that the people who really master this part of the code write good one.

PR #2036, opened 2021-09-03
MOD: add some cpp feature, like distance/norm/inner prod computations

intech07:
I want to invoke norm  computations by using CGO, but I find some functions which have been implemented in cpp are not exported in c api, so I commit the PR to solve the problem.

### DIFF
--- a/c_api/utils/distances_c.cpp
+++ b/c_api/utils/distances_c.cpp
@@ -10,6 +10,54 @@
 
 #include "distances_c.h"
 #include <faiss/utils/distances.h>
+#include <cstdio>
+
+void faiss_pairwise_L2sqr(
+        int64_t d,
+        int64_t nq,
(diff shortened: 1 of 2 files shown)


Enet4 (approved):
To me, this seems like an accurate translation to the C API.

intech07:
> To me, this seems like an accurate translation of those functions to the C API.

Thanks, I just need these functions, because I would invoke them to normalize vector in CGO.

mdouze:
LGTM.

intech07:
> Test(s) failing
> Test(s) failing internally on Facebook.
> 
> DETAILS
> If you are a Facebook employee, you can view these tests on Phabricator.

@mdouze Sorry, CI indicates that something failed, but I can't see more detail, please help checking it, then I can fix it ASAP.

PR #2037, opened 2021-09-05
Non-uniform quantization of vector norms

KinglittleQ:
This diff implemented non-uniform quantization of vector norms in additive quantizers. index_factory and I/O are supported.

index_factory:  `XXX_Ncqint{nbits}` where `nbits` is the number of bits to quantize vector norm.

For 8 bits code, it is almost the same as 8-bit uniform quantization. It will slightly improve the accuracy if the code size is less than 8 bits.
```
RQ4x8_Nqint8:  R@1 0.1116
RQ4x8_Ncqint8: R@1 0.1117

RQ4x8_Nqint4:  R@1 0.0901
RQ4x8_Ncqint4: R@1 0.0989
```

### DIFF
--- a/faiss/impl/LocalSearchQuantizer.cpp
+++ b/faiss/impl/LocalSearchQuantizer.cpp
@@ -148,13 +148,8 @@ LocalSearchQuantizer::LocalSearchQuantizer(
         size_t d,
         size_t M,
         size_t nbits,
-        Search_type_t search_type,
-        size_t nbits_norm)
-        : AdditiveQuantizer(
-                  d,
-                  std::vector<size_t>(M, nbits),
-                  search_type,
-                  nbits_norm) {
+        Search_type_t search_type)
+        : AdditiveQuantizer(d, std::vector<size_t>(M, nbits), search_type) {
     is_trained = false;
     verbose = false;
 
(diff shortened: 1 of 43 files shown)


faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -266,6 +272,11 @@ void LocalSearchQuantizer::train(size_t n, const float* x) {
                 norm_max = norms[i];
             }
         }
+
+        if (search_type == ST_norm_cqint) {
+            Clustering clus(1, (1 << nbits_norm));
+            clus.train(n, norms.data(), qnorm);
```
mdouze, line 278: apparently there is a dynamic programming way to solve the 1D k-means.

faiss/impl/index_read.cpp
```diff
@@ -263,6 +263,8 @@ static void read_AdditiveQuantizer(AdditiveQuantizer* aq, IOReader* f) {
     READ1(aq->search_type);
     READ1(aq->norm_min);
     READ1(aq->norm_max);
+    READ1(aq->nbits_norm);
```
mdouze, line 266: please make it backwards compatible by storing the qnorm only for qcint variants

mdouze (changes requested):
It is on purpose that I did not add a field to set the number of bits for the quantized norm. Could you just support cqint4 and cqint8 and maybe a few others instead of adding a new field for the number of bits in the norm?

KinglittleQ:
@mdouze  Sure, I'm going to fix them.

mdouze:
Very interesting paper about 1D k-means: https://arxiv.org/abs/1701.07204
@KinglittleQ if you want to give a stab at extracting a practical algorithm from it... Otherwise I'll do it at some point.

KinglittleQ:
@mdouze  I've added 1D K-means as described in sec 2.2 of https://arxiv.org/abs/1701.07204. It is O(kn) in time and O(kn) in space.

KinglittleQ:
There is a segmentation fault bug in OSX. Maybe I need some time to find out the problem...

KinglittleQ:
@mdouze PTAL.

mdouze:
Excellent, will import into faiss' FB version.

mdouze (approved):

Issue #2054, opened 2021-09-17
Faiss index memory leak


dmitrijeuseew:
When searching vectors in IndexIVFFlat, RAM usage increases with every new iteration (I suppose that the index has memory leak). I tried to use reset() function, but memory leak remained the same. @mdouze could you tell me please how to cope with increase of RAM usage?

My code which I used for testing the index:

```python
for i in range(10000):
    ent_substr_tfidfs = tfidf_vectorizer.transform([word_list[i]]).toarray().astype(np.float32)
    D_all, I_all = tfidf_faiss_index.search(ent_substr_tfidfs, 10)
    if i%100 == 0:
        print(i, faiss.get_mem_usage_kb())

'''
i = 0, faiss.get_mem_usage_kb() returns 9353148, in htop 9142M
i = 1000, faiss.get_mem_usage_kb() returns 9400856, in htop 9185M
i = 2000, faiss.get_mem_usage_kb() returns 9470916, in htop 9252M
i = 3000, faiss.get_mem_usage_kb() returns 9552472, in htop 9334M
i = 4000, faiss.get_mem_usage_kb() returns 9633428, in htop 9400M
'''
```

mdouze:
In the first loop, could you move the line 
```
   ent_substr_tfidfs = tfidf_vectorizer.transform([word_list[i]]).toarray().astype(np.float32)
```
out of the loop (just repeatedly search the same vector). Then do loops until the memory usage explodes, eg. reaches above 2*9142M, the initial mem usage?

dmitrijeuseew:
@mdouze I suppose that if different vectors are searched, cache is saved in RAM and it causes memory explosion.

dmitrijeuseew:
@mdouze I modified the code so that list of tfidf vectors is calculated before searching in Faiss index and memory explosion was the same.

```python
import copy
import gc
import pickle
import random
import sys
import time
import faiss
import numpy as np
from sklearn.feature_extraction.text import TfidfVectorizer

# file can be downloaded from http://files.deeppavlov.ai/kbqa/wikidata/tfidf_vectorizer_vx.pk
with open("/archive/evseev/.deeppavlov/downloads/wikidata_rus/tfidf_vectorizer_vx.pk", 'rb') as fl:
    tfidf_vectorizer = pickle.load(fl)

# file can be downloaded from http://files.deeppavlov.ai/kbqa/wikidata/word_to_idlist_vx.pickle
[...]

PR #2059, opened 2021-09-22
Reduce mem usage  + improve performance for sequential search imlementation

mdouze:
Following up on issue #2054 it seems that this code crashes Faiss (instead of just leaking memory). 
This PR is just to run it in CI to see if this is because I did not compile Faiss properly.

### DIFF
--- a/faiss/utils/distances.cpp
+++ b/faiss/utils/distances.cpp
@@ -105,8 +105,9 @@ void exhaustive_inner_product_seq(
         size_t ny,
         ResultHandler& res) {
     using SingleResultHandler = typename ResultHandler::SingleResultHandler;
+    int nt = std::min(int(nx), omp_get_max_threads());
 
-#pragma omp parallel
+#pragma omp parallel num_threads(nt)
     {
         SingleResultHandler resi(res);
 #pragma omp for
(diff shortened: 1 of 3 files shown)


mdouze:
To repro crash: 
```
LD_LIBRARY_PATH=/private/home/matthijs/.conda/envs/faiss_1.7.1/lib  $cmake -B build . -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON -DFAISS_OPT_LEVEL=avx2 -DFAISS_ENABLE_GPU=OFF -DBLA_VENDOR=Intel10_64_dyn

make -C build -j 10 test_mem_leak

OMP_NUM_THREADS=1 gdb build/demos/test_mem_leak

```

mdouze:
Findings: 
- when running in MT mode, each search in an indexflat used as coarse quantizer consumes some memory
- this mem consumption does not appear in single-thread mode or with few threads
- in gdb it appears that even when the nb of queries is 1, each search spawns max_threads threads (80 on the test machine)
This diff: 
- adds a C++ test that checks how much mem is used when repeatedly searching a vector
- adjusts the number of search threads to the number of query vectors. This is especially useful for single-vector queries.

mdouze:
unclear why not auto closed

PR #2085, opened 2021-10-19
Add an epsilon to avoid numerical instability in PCA whiteneing

mdouze:
Summary:
PCA whitening implies to multiply eigenvectors with 1/sqrt(singular values of convariance matrix). The singular values are sometimes 0 (because the vector subspace is not full-rank) or negative (because of numerical issues).
Therefore, this diff adds an epsilon to the denominator above (default 0).

Reviewed By: edpizzi

Differential Revision: D31725075


mdouze:
was merged. Unclear why the PR is not closed automatically.

Issue #2081, opened 2021-10-16
IndexNSG reorder error


guoxiangzhou:
# Summary

# Platform

OS: 
macOS 11.5.2

Faiss version: 

Installed from:  

Faiss compilation options: 

Running on:
-  CPU

Interface: 
- C++

# Reproduction instructions

```
    int d = 64;      // dimension
    int nb = 1000; // database size
    int nq = 1;  // nb of queries

    std::mt19937 rng;
    std::uniform_real_distribution<> distrib;

    auto* xb = new float[d * nb];
    auto* xq = new float[d * nq];

    for (int i = 0; i < nb; i++) {
        for (int j = 0; j < d; j++)
            xb[d * i + j] = distrib(rng);
        xb[d * i] += i / 1000.;
    }

    for (int i = 0; i < nq; i++) {
        for (int j = 0; j < d; j++)
            xq[d * i + j] = xb[d * i + j];
    }

    faiss::IndexNSGFlat index(d, 32);
    index.train(nb, xb);
    index.add(nb, xb);

    {
        int k = 4;

        idx_t* I = new idx_t[k * nq];
        float* D = new float[k * nq];

        index.search(nq, xq, k, D, I);

        printf("I=\n");
        for (int i = 0; i < nq; i++) {
            for (int j = 0; j < k; j++)
                printf("%5lld:%f ", I[i * k + j], D[i * k + j]);
            printf("\n");
        }
    }
```
result :

723:7.600873   254:7.687237   152:7.827541     0:0.000000 

IndexNSG.cpp

```
157                maxheap_heapify(k, simi, idxi);
158                nsg.search(*dis, k, idxi, simi, vt, bitset);
159            //  maxheap_reorder(k, simi, idxi);
```

I can get result:

0:0.000000   723:7.600873   254:7.687237   152:7.827541

PR #2086, opened 2021-10-19
Fix unorder bug in NSG

KinglittleQ:
The results returned by `NSG::search` are already sorted. Calling `maxheap_reorder` will make the results unorder.

Fixed https://github.com/facebookresearch/faiss/issues/2081.

### DIFF
--- a/faiss/IndexNNDescent.cpp
+++ b/faiss/IndexNNDescent.cpp
@@ -167,9 +167,7 @@ void IndexNNDescent::search(
                 float* simi = distances + i * k;
                 dis->set_query(x + i * d);
 
-                maxheap_heapify(k, simi, idxi);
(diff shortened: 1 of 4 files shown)


faiss/IndexNNDescent.cpp
```diff
@@ -167,9 +167,7 @@ void IndexNNDescent::search(
                 float* simi = distances + i * k;
                 dis->set_query(x + i * d);
 
-                maxheap_heapify(k, simi, idxi);
                 nndescent.search(*dis, k, idxi, simi, vt);
-                maxheap_reorder(k, simi, idxi);
```
KinglittleQ, line 172: The same problem as NSG.

SuoSiFire:
This seems to be a more important issue, can it be merged?

PR #2101, opened 2021-11-05
Speedup write_InvertedLists

wxingda:
Use reserve() before push_back() to avoid memory multiple reallocation, it will improve write_InvertedLists's performance.


mdouze:
unless proven otherwise I think the impact is negligible.

mdouze:
was merged already, unclear why not closed.

PR #2166, opened 2021-12-24
Post-training refinement of residual quantizer codebooks

mdouze:
Summary:
RQ training is done progressively from one quantizer to the next, maintaining a current set of codes and quantization centroids.
However, for RQ as for any additive quantizer, there is a closed form solution for the centroids that minimizes the quantization error for fixed codes.
This diff offers the option to estimate that codebook at the end of the optimization. It performs this estimation iteratively, ie. several rounds of code computation - codebook refinement are performed.

A pure python implementation + results is here:
https://github.com/fairinternal/faiss_improvements/blob/dbcc746/decoder/refine_aq_codebook.ipynb

Differential Revision: D33309409

### DIFF
--- a/faiss/impl/ResidualQuantizer.cpp
+++ b/faiss/impl/ResidualQuantizer.cpp
@@ -10,6 +10,7 @@
 #include <faiss/impl/ResidualQuantizer.h>
 
 #include <algorithm>
+#include <cmath>
 #include <cstddef>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


faiss/impl/ResidualQuantizer.cpp
```diff
@@ -381,13 +409,128 @@ void ResidualQuantizer::train(size_t n, const float* x) {
...
     }
 
-    is_trained = true;
-
     if (!(train_type & Skip_codebook_tables)) {
         compute_codebook_tables();
     }
 }
 
+float ResidualQuantizer::retrain_AQ_codebook(size_t n, const float* x) {
```
KinglittleQ, line 417: That looks great! If I understand correctly, it is similar to `LocalSearchQuantizer::update_codebooks`. Could we move this function to AdditiveQuantizer?

mdouze: Ah right, there is some redundancy with LSQ. LMC which function is better to keep. Probably needs to be merged because yours has a double version but mine supports multiple nb bits per quantizer.

mdouze: Sorry, I am going to integrate the PR as it is now because I need it quickly. 
For reference the differences are: 
- both float + double support for the LocalSearchQuantizer version
- variable nb bits support for the ResidualQuantizer version
- degeneracy (presumably due to missing example codes in the training data): 
* LSQ --> add lamdba * diagonal to linear system matrix 
* RQ --> explicit estimate of rank of the matrix + clip SVD decomposition to this rank 
The merits of both solutions should be compared when merging.

PR #2168, opened 2021-12-26
Fix omp & memory leak

KinglittleQ:
Fix an OMP bug and a memory leakage bug. The first one would lead to non-deterministic results and even worse.


KinglittleQ:
@mdouze Could you take a look? I think it is quite important.

mdouze:
Ough, indeed a beginner's error in OMP.

PR #2169, opened 2021-12-26
AQ fastscan

KinglittleQ:
Work in progress.


mdouze:
I am really sorry I did not notice this PR earlier. 
I am also working on a fast-scan version of AQ but it is not functional. 
The issue with your version is that there is way too much code duplication between the PQ and AQ code. 
Let's see how we can simplify that.

mdouze:
My version with AQ fast scan is here:
 https://github.com/facebookresearch/faiss/pull/2213
It is much less advanced than yours, so probably it is better to keep yours. 

However, could you take the following comments into account: 

- please explain in comments or a doc somewhere what the normalization options lsq2x4 and rq2x4 refer to and how they are stored along with the AQ codes (is it worthy of a paper?)

- the difference in code between PQ and AQ is not that big, the accumulation is essentially the same. In my version I made a common ancestor IndexFastScan for the IndexPQFastScan and IndexResidualQuantizerFastScan, please do the same here + also the IVF version

- I think the specific handling of norms should be hidden in a class that can be templatized so that the code in pq_fast_scan does not need to be duplicated.

- we need benchmarks to show that the AQ fasrt-scan version brings accuracy gains compared to PQ fast-scan. I currently have that only for IP search and without IVF, see 
image

KinglittleQ:
Oh... That's a pity. I thought you'd already seen this. Maybe I should create an issue first. This implementation works now and has passed the tests. 

Here are what my implementation includes:

1. Use a 2x4 bits additive quantizer to encode the norm (ST_norm_lsq2x4 and ST_norm_rq2x4).
    This could achieve almost the same accuracy as ST_norm_qint8 and benefit from 4-bit AQ fastscan search.

2. Rescale the norm look-up tables before searching in IVFAQFastScan. This could slightly improve the accuracy
    (0.3% in SIFT1M, IVF1024,LSQ8x4fs, nprobe = 256).
    This trick does not slow down the search speed but involves some modifications in the search code
    compared with IVFPQFastScan. If you found that it is hard to merge this trick, we could remove it.


A base class named VectorQuantizer which includes M, nbits, compute_codes, decode and compute_LUT may
help to merge the duplicated code.

I'm sorry that we're both working on the same thing. If you have any ideas about how to deal with this, please tell me.

KinglittleQ:
> My version with AQ fast scan is here: #2213 It is much less advanced than yours, so probably it is better to keep yours.
> 
> However, could you take the following comments into account:
> 
> * please explain in comments or a doc somewhere what the normalization options lsq2x4 and rq2x4 refer to and how they are stored along with the AQ codes (is it worthy of a paper?)
> * the difference in code between PQ and AQ is not that big, the accumulation is essentially the same. In my version I made a common ancestor IndexFastScan for the IndexPQFastScan and IndexResidualQuantizerFastScan, please do the same here + also the IVF version
> * I think the specific handling of norms should be hidden in a class that can be templatized so that the code in pq_fast_scan does not need to be duplicated.
> * we need benchmarks to show that the AQ fasrt-scan version brings accuracy gains compared to PQ fast-scan. I currently have that only for IP search and without IVF, see
>   image

Sure, I'm going to address them.

mdouze:
ST_norm_lsq2x4 / rq2x4 is actually a very good idea, that makes it more uniform w.r.t. the other components. 
IIUC the motivation of the scale factor is because the magnitude of the norm component is much larger than that of the other components, correct?

faiss/IndexIVFAQFastScan.cpp
```diff
@@ -0,0 +1,1320 @@
...
+    for (idx_t i = 0; i < n; i++) {
+        const float* lut = dis_tables.get() + i * M * ksub;
+        scale += quantize_lut::aq_estimate_norm_scale(M, ksub, 2, lut);
+    }
+    scale /= n;
+
+    scale = std::max(scale, 1.0f);
+    scale = std::roundf(scale);
+    norm_scale = (int)scale;
+
```
mdouze, line 251: if(verbose)

mdouze:
> ST_norm_lsq2x4 / rq2x4 is actually a very good idea, that makes it more uniform w.r.t. the other components. IIUC the motivation of the scale factor is because the magnitude of the norm LUT entries is much larger than that of the other components, correct?

Seems quite obvious here: https://gist.github.com/mdouze/69fad2b26fdaca526f3db14a8f815d7e

faiss/impl/aq4_fast_scan_search_qbs.cpp
```diff
@@ -0,0 +1,267 @@
...
+                // 16 base elements
+                accu[q][0] += simd16uint16(res0);
+                accu[q][1] += simd16uint16(res0) >> 8;
+
+                // the other 16 base elements
+                accu[q][2] += simd16uint16(res1);
+                accu[q][3] += simd16uint16(res1) >> 8;
+            } else {
+                // 16 base elements
+                accu[q][0] += simd16uint16(res0) * scale_factor;
```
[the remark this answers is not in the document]
KinglittleQ: Could we put this in a follow-up PR? In most cases, I didn't observe too much improvement with scaling the norm. I need some time to confirm the effects.

KinglittleQ:
Yes, the norm LUT has a wider range than the inner-product LUTs.  ST_norm_lsq2x4 / rq2x4 could reduce the discrepancy because the norm would be splitted into two tables. In a IVF1024,LSQ8x4fs index, the norm scale factor is 3.

faiss/impl/AdditiveQuantizer.h
```diff
@@ -37,7 +37,9 @@ struct AdditiveQuantizer {
     bool verbose;    ///< verbose during training?
     bool is_trained; ///< is trained or not
 
-    IndexFlat1D qnorm; ///< store and search norms
+    IndexFlat1D qnorm;            ///< store and search norms
+    std::vector<float> norm_tabs; ///< store norm tables for 4-bit fastscan
```
mdouze, line 41: norms of codebook entries?

faiss/impl/AdditiveQuantizer.h
```diff
@@ -57,6 +59,9 @@ struct AdditiveQuantizer {
         ST_norm_qint4,
         ST_norm_cqint8, ///< use a LUT, and store non-uniform quantized norm
         ST_norm_cqint4,
+
+        ST_norm_lsq2x4, ///< use a 2x4 bits lsq as norm quantizer
```
mdouze, line 63: (for fast scan)

mdouze (commented):
Thanks a lot for the hard work. 

I would prefer to call the classes 

IndexAdditiveQuantizerFastScan 
IndexResdiualQuantizerFastScan 

ie, not using RQ and AQ in the names,  because that is how the original classes are called as well. 

I understand that this makes very long names.

mdouze:
Did you remove the norm scale support in the current version? A boost of 0.3 recall points is not negligible. 

In fact, if we have a special handling of the norms we may as well directly encoding a qint8 norm. 

I realized that I forgot how the fast-scan code is organized. I am going to make a wiki page about it, just to keep that in mind.

mdouze:
You mention a norm / codebook scale of 3, but that is for AQ8x4, typical use cases for Fast scan are more in the order of AQ32x4, so the scale will be even larger.

KinglittleQ:
I've added the norm scale back but I am still validating the results. The improvement of it may not be obvious as we expected and it does not slow down the speed according to my observation.

Here is a benchmark result by `benchs/bench_ivf_fastscan.py`:

bench_ivf_fastscan

I will give a more detailed analysis in a few days.

faiss/impl/pq4_fast_scan.h
```diff
@@ -89,15 +89,16 @@ void pq4_pack_LUT(int nq, int nsq, const uint8_t* src, uint8_t* dest);
  * @param codes   packed codes array
  * @param LUT     packed look-up table
```
mdouze, line 90: please document the scaler parameter

faiss/impl/pq4_fast_scan_search_qbs.cpp
```diff
@@ -27,12 +28,13 @@ namespace {
...
 void kernel_accumulate_block(
         int nsq,
         const uint8_t* codes,
         const uint8_t* LUT,
-        ResultHandler& res) {
+        ResultHandler& res,
+        const Scaler& scaler) {
     // dummy alloc to keep the windows compiler happy
     constexpr int NQA = NQ > 0 ? NQ : 1;
     // distance accumulators
```
mdouze, line 40: Would you mind adding the following comment: 
```
// layout: accu[q][b]: distance accumulator for vectors 8*b..8*b+7
```

faiss/impl/pq4_fast_scan_search_qbs.cpp
```diff
@@ -63,11 +65,11 @@ void kernel_accumulate_block(
...
-            accu[q][1] += simd16uint16(res0) >> 8;
+            accu[q][0] += scaler.scale(sq, simd16uint16(res0));
+            accu[q][1] += scaler.scale(sq, simd16uint16(res0) >> 8);
 
-            accu[q][2] += simd16uint16(res1);
-            accu[q][3] += simd16uint16(res1) >> 8;
+            accu[q][2] += scaler.scale(sq, simd16uint16(res1));
+            accu[q][3] += scaler.scale(sq, simd16uint16(res1) >> 8);
         }
     }
```
mdouze, line 75: There are two improvements that I would like to have here: 
- make sure that the internal loop does not include tests (on sq) 
- make the scaler object more generic, so that it can handle 8-bit encoded norms as well. 

For this I suggest to 
```c++
// where static const int nscale = 0 or 2 for the time being 
for (int sq = 0; sq < nsq - scaler.nscale; sq += 2) {
  /// keep default loop the same 
} 

// specific handling for norms encoding 
for (int sq = 0; sq < scaler.nscale; sq += 2) {
....
    for (int q = 0; q < NQ; q++) {
            // load LUTs for 2 quantizers
            simd32uint8 lut(LUT);
            LUT += 32;

            simd32uint8 res0 = scaler.lookup(lut, clo);
            accu[q][0] += scaler.scale_lo(res0); // handle vectors 0..7
            accu[q][1] += scaler.scale_hi(res0); // handle vectors 8..15 
            
            simd32uint8 res1 = scaler.lookup(lut, chi);
            accu[q][2] += scaler.scale_lo(res1);  // handle vectors 16..23
            accu[q][3] += scaler.scale_hi(res1); //  handle vectors 24..31

   } 
}

```

scale_lo and scale_hi could choose to combine the indices for nsq - 1 and nsq-2 to reconstruct a 8-bit int. 
Then the scaler.lookup function would be a NOP, since the lookup is not used.

mdouze:
Thanks again for the changes. This result plot looks very good! 

If I understand what you mean: introducing the scaling does not slow down the search but does not improve the results by that much. If you redo the experiments, could you: 
- include results with / without scaling 
- run experiments single-threaded to reduce jitter in timings
- add the script that generates the results in faiss/benchs/ so that is is easy to repro

Are the results with or without residual wrt. the IVF centroids?

KinglittleQ:
The above results are with residuals (by_residual=True).

I made some changes to the index factory. Now `IVF1024,RQ32x4fs` means `by_residual=False` and `IVF1024,RQ32x4fsr` means `by_residual=True`. 

Here is a benchmark result from `benchs/bench_ivf_fastscan.py`, running with a single thread:
bench_ivf_fastscan

mdouze:
Thanks for the changes. It could be a bit more documented (eg the Scaler object), but I'll import it to the FB build sys.

PR #2185, opened 2022-01-10
fix c api example

qieqieplus:


Enet4 (approved):
Yup, probably a mistake from code copy-pasting.

PR #2192, opened 2022-01-13
Update binary index default metric type to METRIC_Hamming

cydrain:


mdouze:
METRIC_Hamming does not exist because it is equivalent to METRIC_L2 for binary values (and METRIC_L1).

Issue #2227, opened 2022-02-20
TypeError: in method 'GpuIndex_add_with_ids', argument 4 of type 'faiss::Index::idx_t const *'


poorya2021:
# Summary

Hi guys
I have a problem 
This is my code:

```
from pyterrier_colbert.indexing import ColBERTIndexer
indexer = ColBERTIndexer(addr_checkpoint, addr + "Indexing/", "index_doc",  chunksize=3)
doc_26k_collection = pd.read_csv(addr+'Sajad_ds_small/doc_26k.tsv',sep='\t')
doc_collection = doc_26k_collection.rename({'doc_id':'docno'}, axis='columns')
doc_collection['docno'] = doc_collection['docno'].astype(str)
indexer.index(doc_26k_collection_new.to_dict("records"))

```
Outputs:

```
TypeError                                 Traceback (most recent call last)
C:\Users\ADMINI~1\AppData\Local\Temp/ipykernel_7936/1940468918.py in <module>
     13 
     14 # ###############
---> 15 indexer2.index(doc_26k_collection_new.to_dict("records"))

C:\ProgramData\Anaconda3\lib\site-packages\pyterrier_colbert\indexing.py in index(self, iterator)
    351             warn("Default computation chooses", self.args.partitions,
    352                         "partitions (for {} embeddings)".format(num_embeddings))
--> 353         index_faiss(self.args)
    354         print("#> Faiss encoding complete")
    355         endtime = timer()

C:\ProgramData\Anaconda3\lib\site-packages\colbert\indexing\faiss.py in index_faiss(args)
    106 
    107             print_message("#> Processing a sub_collection with shape", sub_collection.shape)
--> 108             index.add(sub_collection)
    109 
    110         print_message("Done indexing!")

C:\ProgramData\Anaconda3\lib\site-packages\colbert\indexing\faiss_index.py in add(self, data)
     46 
     47         if self.gpu.ngpu > 0:
---> 48             self.gpu.add(self.index, data, self.offset)
     49         else:
     50             self.index.add(data)
C:\ProgramData\Anaconda3\lib\site-packages\colbert\indexing\faiss_index_gpu.py in add(self, index, data, offset)
    107             xs = data[i0:i1]
    108 
--> 109             self.gpu_index.add_with_ids(xs, np.arange(offset+i0, offset+i1))
    110 
    111             if self.max_add > 0 and self.gpu_index.ntotal > self.max_add:

[...]

PR #2250, opened 2022-03-11
Typechecking of numpy arrays

cmacdonald:
Please consider this pull request to add extra type checking of the numpy arrays and useful error messages, as discussed in #2227.

The enclosed PR passes the Python tests on my Intel imac machine, but I get a segfault in tests/test_index.py (even without this patch).

One doubt I have - I made sure to run tests/torch_test_contrib.py (which passed), but I think it is necessary to check for torch types as well as numpy types?

### DIFF
--- a/faiss/python/__init__.py
+++ b/faiss/python/__init__.py
@@ -596,24 +596,16 @@ def handle_IndexBinary(the_class):
     def replacement_add(self, x):
         n, d = x.shape
         assert d * 8 == self.d
-        if x.dtype != np.float32:
-            raise TypeError("Input argument %s must be ndarray of dtype float32, but found %s" % ("x", str(x.dtype)))
         self.add_c(n, swig_ptr(x))
 
     def replacement_add_with_ids(self, x, ids):
         n, d = x.shape
-        if x.dtype != np.float32:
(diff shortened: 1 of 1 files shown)


cmacdonald:
I have signed the CLA - not sure when the bot will notice.

faiss/python/__init__.py
```diff
@@ -211,6 +211,10 @@ def replacement_add(self, x):
         """
 
         n, d = x.shape
+
+        if x.dtype != np.float32:
+            raise TypeError("Input argument %s must be ndarray of dtype float32, but found %s" % ("x", str(x.dtype)))
```
mdouze, line 216: Please make sure that the line is < 80 characters.
Nit: %s already applies str() on its argument, so str() is not necessary.

mdouze (commented):
Thanks for clarifying the error messages!
Please address the comment and I'll accept the PR.

faiss/python/__init__.py
cmacdonald: now addressed

cmacdonald:
Feedback has been addressed

mdouze:
In fact, this diff highlights a recurrent problem with Faiss input arrays, so I started a PR that does automatic conversion + a few errors when the conversion is not obvious (that uses a bit of this PR's code), see: 

https://github.com/facebookresearch/faiss/pull/2274

PR #2233, opened 2022-02-28
Summary: Add Binary Index i.o to c_api

CarterMcClellan:
Summary:
Add Binary Index i.o to c_api

In reference to isse: https://github.com/facebookresearch/faiss/issues/2225

PR #2261, opened 2022-03-15
linted Add Binary Index i.o to c_api

mdouze:
Differential Revision: D34894263


mdouze:
This was landed as https://github.com/facebookresearch/faiss/pull/2233

PR #2263, opened 2022-03-15
Fix GPU IVF serialization bug; fix stream issues (#2258)

wickedfoo:
Summary:
Pull Request resolved: https://github.com/facebookresearch/faiss/pull/2258

GPU IVF indices could not be properly deserialized or copied from CPU then added to after the fact, this resulted in the following C++ assertion:

```
Faiss assertion 'indices->numVecs == oldNumVecs' failed in int faiss::gpu::IVFBase::addVectors(faiss::gpu::Tensor<float, 2, true>&, faiss::gpu::Tensor<long int, 1, true>&) at faiss/gpu/impl/IVFBase.cu:581
```

as the count of vectors present was not updated properly everywhere, as discovered internally by vtantia.

This diff fixes this issue by properly updating the count, as well as cleaning up stream usage in the IVF code. The problem is that the code was previously using `thrust::device_vector` which does not have a means to control on which stream copies or other work is performed. This is fixed by replacing all usage of `thrust::device_vector` with our own `DeviceVector` which was already used to store IVF data but not metadata. `DeviceVector` provides sufficient control over the proper CUDA stream usage.

Reviewed By: vtantia, mdouze

Differential Revision: D34886859

### DIFF
--- a/faiss/gpu/CMakeLists.txt
+++ b/faiss/gpu/CMakeLists.txt
@@ -152,7 +152,7 @@ set(FAISS_GPU_HEADERS
   utils/StaticUtils.h
   utils/Tensor-inl.cuh
   utils/Tensor.cuh
-  utils/ThrustAllocator.cuh
+  utils/ThrustUtils.cuh
   utils/Timer.h
   utils/Transpose.cuh
   utils/WarpPackedBits.cuh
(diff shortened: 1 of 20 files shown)


JeanBaptiste-dlb:
I face the same issue, thanks for the fix. When can it be made available in global release? 
Thanks in advance for your reply,

PR #2274, opened 2022-03-21
Automatic type conversions for Python API

mdouze:
Summary:
All input matrices needed to be of the correct type and to be C-contiguous. This diff passes the main entry points of the api through `np.tocontiguousarray` so that the function parameters are transparently converted to the suitable format if needed.

We did not have this before because users need to be made aware of the performance impact, but it seems that maybe usability is more useful.

This diff is an alternative to

D35007365
https://github.com/facebookresearch/faiss/pull/2250

Differential Revision: D35009612


cmacdonald:
Numpy.ascontiguousarray() also requires numpy 1.20 - worth setting in setup.py?

PR #2286, opened 2022-04-05
Add ProductAdditiveQuantizer

KinglittleQ:
This diff added ProductAdditiveQuantizer.

A Simple Algo description:

1. Divide the vector space into several orthogonal sub-spaces, just like PQ does.
2. Quantize each sub-space by an independent additive quantizer.

Usage:

Construct a ProductAdditiveQuantizer object:
- `d`: dimensionality of the input vectors
- `nsplits`: number of sub-spaces divided into
- `Msub`: `M` of each additive quantizer
- `nbits`: `nbits` of each additive quantizer

```python
d = 128
nsplits = 2
Msub = 4
nbits = 8
plsq = faiss.ProductLocalSearchQuantizer(d, nsplits, Msub, nbits)
prq = faiss.ProductResidualQuantizer(d, nsplits, Msub, nbits)
```


benchs/bench_quantizer.py
```diff
@@ -106,6 +106,12 @@ def eval_quantizer(q, xq, xb, gt, xt, variants=None):
     print("===== PQ")
     eval_quantizer(pq, xq2, xb2, gt, xt2)
 
+if 'plsq' in todo:
+    print("===== Product LSQ")
+    plsq = faiss.ProductLocalSearchQuantizer(d, 2, M // 2, nbits)
```
mdouze, line 111: is 2 sub-quantizers a sweet spot for the ProductAddtiveQuantizer?

faiss/impl/ProductAdditiveQuantizer.h
```diff
@@ -0,0 +1,128 @@
...
+#include <faiss/impl/AdditiveQuantizer.h>
+#include <faiss/impl/LocalSearchQuantizer.h>
+#include <faiss/impl/ResidualQuantizer.h>
+
+namespace faiss {
+
+/** Product Additive Quantizers
+ * TODO: add some docs
+ */
+struct ProductAdditiveQuantizer : AdditiveQuantizer {
```
mdouze, line 23: why is this a sub-class of additivequantizer ? If you override the decoding method, then there is little left from AQ no?

faiss/impl/ProductAdditiveQuantizer.h
```diff
@@ -0,0 +1,128 @@
...
+    size_t nsplits; ///< number of sub-vectors we split a vector into
+
+    std::vector<AdditiveQuantizer*> quantizers;
+
+    ProductAdditiveQuantizer(
+            size_t d,
+            const std::vector<AdditiveQuantizer*>& aqs,
```
mdouze, line 30: please clarify ownership of the AddtitiveQuantizer's. Is it transferred to the ProductAdditiveQuantizer?

mdouze (commented):
Do you have comparative search results wrt. PQ and/or LSQ to show where this quantizer stands?

faiss/impl/ProductAdditiveQuantizer.h
KinglittleQ: I make it a sub-class of AQ because the search code of AQs could be re-used in the future (e.g. IndexAdditiveQuantizer and IndexIVFAdditiveQuantizer ).

faiss/impl/ProductAdditiveQuantizer.h
KinglittleQ: The ownership is not transferred to the ProductAdditiveQuantizer. I will clarify it in the docs.

benchs/bench_quantizer.py
KinglittleQ: No, this is just a simple demo configuration.

KinglittleQ:
Thank you very much for your comments. This PR is still in progress. I will give a detailed analysis in a few days.

faiss/impl/ProductAdditiveQuantizer.cpp
```diff
@@ -0,0 +1,358 @@
...
+        size_t nsplits,
+        size_t Msub,
+        size_t nbits,
+        Search_type_t search_type) {
+    FAISS_THROW_IF_NOT(d % nsplits == 0);
+    size_t dsub = d / nsplits;
+    std::vector<AdditiveQuantizer*> aqs;
+    for (size_t i = 0; i < nsplits; i++) {
+        rqs.emplace_back(dsub, Msub, nbits, ST_decompress);
+        aqs.push_back(&rqs[i]);
```
KinglittleQ, line 346: BUG: the underlying memory of `rqs` could be changed during the loop. Use `reserve()` to fix it.

KinglittleQ:
Here are some benchmark results. For more details and analysis, please see [here ](https://gist.github.com/KinglittleQ/af4b78c71c8509e89ebcad8df6efb60e.).

image

image

faiss/impl/ProductAdditiveQuantizer.h
```diff
@@ -0,0 +1,151 @@
...
+struct ProductAdditiveQuantizer : AdditiveQuantizer {
+    size_t nsplits; ///< number of sub-vectors we split a vector into
+
+    std::vector<AdditiveQuantizer*> quantizers;
+
+    /** Construct a product additive quantizer.
+     *
+     * The ownership of the additive quantizers passed in is not
+     * transferred to the ProductAdditiveQuantizer object. Users
+     * are responsible to take care of it.
```
mdouze, line 37: This is problematic because the object cannot be self-contained in a C++ context (for example in an index_factory). 
To copy the object, we need a clone_quantizer function and a common ancestor class for the quantizer objects. 
IMO as long as we don't have the Index object implementation we don't need it, but to keep in mind for later.

mdouze (commented):
The results look convincing!

mdouze:
See here for a PR that makes a common ancestor for quantizer classes: https://github.com/facebookresearch/faiss/pull/2295

faiss/impl/ProductAdditiveQuantizer.h
KinglittleQ: We have `ProductLocalSearchQuantizer` and `ProductResidualQuantizer` which are able to be self-contained. Should we need to make `ProductAdditiveQuantizer` self-contained too?

KinglittleQ:
> See here for a PR that makes a common ancestor for quantizer classes: #2295

That looks good! I will solve the conflicts after #2295 is merged.

KinglittleQ:
> See here for a PR that makes a common ancestor for quantizer classes: #2295

That looks good! I will solve the conflicts after #2295 is merged.

faiss/impl/ProductAdditiveQuantizer.cpp
```diff
@@ -0,0 +1,347 @@
...
+        FINTEGER* k,
+        const float* alpha,
+        const float* a,
+        FINTEGER* lda,
+        const float* b,
+        FINTEGER* ldb,
+        float* beta,
+        float* c,
+        FINTEGER* ldc);
+}
```
KinglittleQ, line 41: By the way, the matrix multiplication code appears in many places. Could we make it a utility function that accepts matrices in row-major order as input? If so, I could do it in the future.

mdouze: In fact I would like to move to the cblas interface (http://www.netlib.org/blas/cblas.h). This would also be good because some platforms don't support the fortran calling conventions (iOS). 
However, it's a lot of tedious work to convert all the calls + make sure it works on all platforms. https://github.com/facebookresearch/faiss/projects/3

KinglittleQ:
@mdouze PTAL

faiss/impl/ProductAdditiveQuantizer.cpp
```diff
@@ -0,0 +1,347 @@
...
+    norm_max = norm_min = NAN;
+    code_size = 0;
+    tot_bits = 0;
+    total_codebook_size = 0;
+    only_8bit = false;
+    set_derived_values();
+
+    // ProductAdditiveQuantizer
+    nsplits = aqs.size();
+    quantizers = aqs;
```
mdouze, line 83: please use the clone_Quantizer function here to transfer ownership

mdouze (commented):
I agree that the PR is already useful in its current state. 
What should be added, in a later PR: 
- transfer ownership of passed-in additive quantizers (or own_fields flag)
- Index support + index_factory
- I/O 
- doc in the wiki + benchmarks

faiss/python/__init__.py
```diff
@@ -194,6 +194,48 @@ def replacement_decode(self, codes):
     replace_method(the_class, 'decode', replacement_decode)
 
 
+def handle_ProductAdditiveQuantizer(the_class):
```
mdouze, line 197: This is more magic than we want for the Python interface. Please let the caller do the loop to get the subquantizer parameters.
Besides, this fails on the internal FB system with 
```

... __init__.py", line 149, in replacement_train
    assert d == self.d

AttributeError: 'ProductResidualQuantizer' object has no attribute 'd'
```

KinglittleQ:
Update:

1. `handle_ProductAdditiveQuantizer` has been removed
2. Now the additive quantizers passed in will be cloned into the ProductAdditiveQuantizer object.

PR #2287, opened 2022-04-05
Support reconstruct() in IndexFastScan

KinglittleQ:
This diff implemented `reconstruct()` interface in `IndexFastScan`.

unit test: `python -m unittest test_fast_scan_ivf.TestReconstruct`


faiss/python/__init__.py
```diff
@@ -883,6 +883,8 @@ def replacement_function(*args):
 add_ref_in_constructor(IndexIVFPQFastScan, 0)
 add_ref_in_constructor(IndexIVFResidualQuantizer, 0)
 add_ref_in_constructor(IndexIVFLocalSearchQuantizer, 0)
+add_ref_in_constructor(IndexIVFResidualQuantizerFastScan, 0)
```
mdouze, line 886: thanks for noticing...

faiss/IndexAdditiveQuantizerFastScan.h
```diff
@@ -69,6 +69,8 @@ struct IndexAdditiveQuantizerFastScan : IndexFastScan {
             idx_t k,
             float* distances,
             idx_t* labels) const override;
+
+    void sa_decode(idx_t n, const uint8_t* bytes, float* x) const override;
```
mdouze, line 73: This deserves a bit of documentation because the bytes are non-contiguous in the index memory. Could you say that this takes a packed representation?

mdouze, line 73: This deserves a bit of documentation because the bytes are non-contiguous in the index memory. Could you say that this takes a packed representation?

mdouze (commented):
LGTM, please address small comments and I'll import.

mdouze (approved):

PR #2292, opened 2022-04-09
Cond dev

lilothar:


mdouze:
This is an interesting addition to Faiss. 
The hardest thing here is to avoid code duplication + not incur a runtime penalty for the non-conditional case. 
The way to obtain this is by having a templatized inner loop that resolves to a NOP when the conditional check is not used.

lilothar:
You have a good suggestion, I understand what you mean and the technical points of concern, template generalization is an interesting idea, I think I can explore it

PR #2298, opened 2022-04-14
Fix reconstruct bug when by_residual is false

spectaclehong:
When I reconstruct with by_residual turned off, the distance was greatly increased. 
This is because the reconstruct_from_offset function did not check if the by_residual option was off. 
I fix this bug with simple if statement.
(like this https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVFPQ.cpp#L365)


mdouze:
Thanks for noticing!
Would you mind adding this test: https://gist.github.com/mdouze/09a4f371ba95d2268fdb81efcd4f70be

Thanks!

PR #2321, opened 2022-05-07
Remove duplicated and in README.md

Uzay-G:
Very small PR, just noticed there was a duplicated and in the README.


mdouze:
There is a big overhead to import PRs to the internal build system, that I will not do for just this change.

PR #2364, opened 2022-06-22
Add sa_code_size, sa_encode, and sa_decode to the C_API

gregsinclair42:
Also link the C API against the FAISS AVX2 build.


gregsinclair42:
Need to fix the CMAKE setting

PR #2401, opened 2022-07-30
Modify installation for Macbook M1 chip.

Ray0907:
It's seems that conda-forge is not support for the macbook m1 chip .So that I add the notes for the m1 chip user


mdouze:
I don't feel confortable advising (and supporting) pip as an install method, especially not for Mac OS, as it produces random crashes. 
I would prefer supporting it via anaconda.

algoriddle:
M1 is supported in conda since v1.7.4

PR #2419, opened 2022-08-21
Update example_c.c

zeyuyun1:
This line should be searching for xq, but the current line is searching the first 5 line in xb. It should be a bug.


Enet4 (approved):
The change seems sound to me.

PR #2423, opened 2022-08-24
Fix potential integer overflow

hhy3:
Signed-off-by: zh Wang <rekind133@outlook.com>


faiss/invlists/InvertedLists.cpp
```diff
@@ -106,7 +106,7 @@ void InvertedLists::print_stats() const {
     }
     for (size_t i = 0; i < sizes.size(); i++) {
         if (sizes[i]) {
-            printf("list size in < %d: %d instances\n", 1 << i, sizes[i]);
+            printf("list size in < %lld: %d instances\n", 1LL << i, sizes[i]);
```
Enet4, line 109: I believe one might as well depend on the same type as `i` and/or that we know that it is unsigned.

So either this:
```suggestion
            printf("list size in < %zu: %d instances\n", static_cast<size_t>(1) << i, sizes[i]);
```

Or this:
```suggestion
            printf("list size in < %llu: %d instances\n", 1ull << i, sizes[i]);
```

Enet4 (approved):

PR #2424, opened 2022-08-24
T120894305 add remove feature to Faiss library 

AbdelrahmanElmeniawy:
add remove method to indexPQ 
test plan:
cd build
cmake .. -DFAISS_ENABLE_GPU=OFF -DBUILD_TESTING=ON
make -j
make test

### DIFF
--- a/faiss/IndexFlatCodes.cpp
+++ b/faiss/IndexFlatCodes.cpp
@@ -25,6 +25,7 @@ void IndexFlatCodes::add(idx_t n, const float* x) {
     }
     codes.resize((ntotal + n) * code_size);
     sa_encode(n, x, codes.data() + (ntotal * code_size));
+    is_removed.resize(ntotal + n);
     ntotal += n;
 }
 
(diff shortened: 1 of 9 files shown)


mdouze:
Thanks for the contribution, however, I don't understand the point of this diff. 
The remove functionality is already in IndexFlatCodes. 
Besides, this adds an auxiliary array that must be checked at search time, an overhead. 
Therefore, I close the diff.

AbdelrahmanElmeniawy:
@mdouze  Yes, you are right in indexIVFPQ is supported and implemented by using Direct map so I will not add any thing for it 
but for indexPQ the implemented way shift the vectors after removing that will cause a problem for any remove or search operations as all ids after removed vectors would be changed, consequently, if the user wants to remove another vector with specific id, we will remove another one. in addition to search issue as if he searches, he will get totally different Ids than expected ones.

e.g: if user used remove_ids(index 2), data in indices 3 and 4 would be sifted. So if user wanted to remove index 3 (as he knew, it should be 20), we would remove 25 instead of 20. the same situation in search
remove with shift

So, I and @beauby  decided to use removed array to flag removed vectors
consequently, when we use removed array, we leave ids not changed and save a lot of shifting time.
yes, there will be overhead in searching but it is only checking a boolean in for every vector (we already iterate over all vectors in search)
remove by flags

PR #2433, opened 2022-08-29
Fix RPC lib logging

francois-travais:
I couldn't run the client-server implementation because of the logging. Indeed `LOG.info('Connected by', addr, end=' ')` raised an exception (`end` is not recognised as a valid argument).

Other warnings are also showing up. This PR clean things up a bit and fixes the client-server.


mdouze:
LGTM

PR #2442, opened 2022-09-01
Update macOS CI Image Tag

wx257osn2:
It seems that [`xcode:12.4.0` has been retired from CircleCI](https://circleci.com/docs/en/using-macos#supported-xcode-versions), so this PR updates the image version and fixes stopping the pipeline at the spin-up stage.

This PR changes only `.circleci/config.yml` , and doesn't affect the software behavior.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ jobs:
 
   build_osx:
     macos:
-      xcode: 12.4.0
+      xcode: 12.5.1
     environment:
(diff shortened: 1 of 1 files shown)


wx257osn2:
`test_standalone_codec.py::TestEncodeDecode::test_RQ6x8` has been failed on macOS CI, but I'd like to separate the investigation and fixing that from this PR (Actually I don't have macOS environment, so I can't investigate it).

PR #2449, opened 2022-09-02
[document] Add vcpkg instruction document

JackBoosY:
`Faiss` available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `Faiss` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `Faiss`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/faiss/portfile.cmake). We try to keep the library maintained as close as possible to the original library.


junjieqi:
We will evaluate it based on the number of questions through github issues and see if there is many similar ask for it

asadoughi:
Closing as we haven't received other requests for this installation method

PR #2451, opened 2022-09-02
remove feature for GpuIndexIVFPQ, but not support interleaved layout

lailifeng-st:
Remove feature for GpuIndexIVFPQ, but not support interleaved layout.
Unit tests are written in the upper-level application, so PR is not included.


mdouze:
Please add a test to see if the functionality works, in the lines of 

https://github.com/facebookresearch/faiss/blob/c0052c15336a57f7068a7d098d5ce5b6234a2d70/tests/test_index_composite.py#L72

(preferably in Python to test the whole stack)

lailifeng-st:
> Please add a test to see if the functionality works, in the lines of
> 
> https://github.com/facebookresearch/faiss/blob/c0052c15336a57f7068a7d098d5ce5b6234a2d70/tests/test_index_composite.py#L72
> 
> (preferably in Python to test the whole stack)



> Please add a test to see if the functionality works, in the lines of
> 
> https://github.com/facebookresearch/faiss/blob/c0052c15336a57f7068a7d098d5ce5b6234a2d70/tests/test_index_composite.py#L72
> 
> (preferably in Python to test the whole stack)

ok

lailifeng-st:
> Please add a test to see if the functionality works, in the lines of
> 
> https://github.com/facebookresearch/faiss/blob/c0052c15336a57f7068a7d098d5ce5b6234a2d70/tests/test_index_composite.py#L72
> 
> (preferably in Python to test the whole stack)




Last week, the python environment was set up, but the " recipe for target' swigfaiss' failed " problem could not be solved, so I wrote the unit test to gpu/test/TestGpuIndexIVFPQ.cpp, and fixed the bugs found in the test. thanks

mdouze:
To compile the python version, please follow the steps lined out in 

https://github.com/facebookresearch/faiss/blob/main/.circleci/config.yml#L222

faiss/gpu/GpuIndexIVFPQ.cu
```diff
@@ -453,5 +457,26 @@ void GpuIndexIVFPQ::verifySettings_() const {
             requiredSmemSize);
 }
 
+size_t GpuIndexIVFPQ::remove_ids(const IDSelector& sel) {
+    const IDSelectorArray* sela = dynamic_cast<const IDSelectorArray*>(&sel);
+    if (sela->n <= 0 || !sela->ids) {
```
wickedfoo, line 462: throw an error if `sela` is NULL (i.e., if an IDSelectorArray instance wasn't passed)

faiss/gpu/impl/IVFBase.cu
```diff
@@ -389,6 +389,9 @@ void IVFBase::addEncodedVectorsToList_(
     // Handle the indices as well
     addIndicesFromCpu_(listId, indices, numVecs);
 
+    // save index mapping
+    indexMapping(listId, indices, numVecs);
```
wickedfoo, line 393: whether or not the index mapping is maintained on the host should be an optional feature for a `GpuIndexIVFPQ` index, which is initialized upon index construction (e.g., in `GpuIndexIVFPQConfig`)

faiss/gpu/impl/IVFBase.cu
```diff
@@ -677,9 +680,209 @@ int IVFBase::addVectors(
             listOffsetDevice,
             stream);
 
+    // save index mapping
+    HostTensor<Index::idx_t, 1, true> indicesHost(indices, stream);
+    indexMapping(listIdsHost, listOffsetHost, indicesHost);
```
wickedfoo, line 685: same thing here, should be gated based on whether or not this feature is enabled upon index construction

faiss/gpu/impl/IVFBase.cu
```diff
@@ -677,9 +680,209 @@ int IVFBase::addVectors(
...
+void IVFBase::removeVectors_(
+        Tensor<int, 1, true>& listIds,
+        Tensor<int, 1, true>& listOffset,
+        Tensor<int, 1, true>& listReplaceOffset,
+        Tensor<long, 1, true>& listIndicesReplaceOffset,
+        cudaStream_t stream) {
+    FAISS_ASSERT_MSG(false, "need to be implemented in subclasses");
+}
+
+size_t IVFBase::removeVectors(size_t n, const Index::idx_t* indices) {
```
wickedfoo, line 700: can you indicate in the name that `indices` is a host pointer (e.g., `indicesHost`)?

Also, looks like the case INDICES_CPU is not handled, can this case be dealt with as well?

faiss/gpu/impl/IVFBase.cuh
```diff
@@ -212,6 +243,9 @@ class IVFBase {
     /// INDICES_CPU), then this maintains a CPU-side map of what
     /// (inverted list id, offset) maps to which user index
     std::vector<std::vector<Index::idx_t>> listOffsetToUserIndex_;
+
+    /// use this map to store the mapping of index to storage location
+    std::map<Index::idx_t, uint64_t> indexMap;
```
wickedfoo, line 248: change this to `unordered_map`, and can you make a struct for the index -> storage information instead of packing it into a `uint64_t` (something like `struct IVFLocation { uint32_t listId; uint32_t offset; };` or some such)

faiss/gpu/impl/IVFPQ.cu
```diff
@@ -256,6 +256,35 @@ void IVFPQ::appendVectors_(
...
+    runIVFIndicesRemove(
+            listIds,
+            listOffset,
+            listReplaceOffset,
+            listIndicesReplaceOffset,
+            indicesOptions_,
+            deviceListIndexPointers_,
+            stream);
+
+    if (interleavedLayout_) {
```
wickedfoo, line 274: A substantial amount of work has already been done up until this point (e.g., `IVFBase::removeVectors`). This check should be the first thing that happens if you call remove, and change the assert to a throw instead

faiss/gpu/test/TestGpuIndexIVFPQ.cpp
```diff
@@ -735,6 +738,98 @@ TEST(TestGpuIndexIVFPQ, UnifiedMemory) {
...
+            faiss::METRIC_L2,
+            config);
+
+    gpuIndex.setNumProbes(opt.nprobe);
+
+    gpuIndex.train(opt.numTrain, trainVecs.data());
+
+    std::vector<faiss::Index::idx_t> ids(opt.numAdd);
+    for (size_t i = 0; i < opt.numAdd; i++) {
+        ids[i] = i;
```
wickedfoo, line 771: can you try using arbitrary, non-sequential/non-monotonic IDs in this test?

faiss/gpu/impl/IVFAppend.cu
```diff
@@ -125,6 +125,70 @@ void runIVFIndicesAppend(
...
+void runIVFIndicesRemove(
+        Tensor<int, 1, true>& listIds,
+        Tensor<int, 1, true>& listOffset,
+        Tensor<int, 1, true>& listReplaceOffset,
+        Tensor<long, 1, true>& listIndicesReplaceOffset,
+        IndicesOptions opt,
+        DeviceVector<void*>& listIndices,
+        cudaStream_t stream) {
+    FAISS_ASSERT(
+            opt == INDICES_CPU || opt == INDICES_IVF || opt == INDICES_32_BIT ||
```
wickedfoo, line 172: this function should only allow `INDICES_32_BIT` and `INDICES_64_BIT`, as remove on `INDICES_IVF` doesn't make sense, and `INDICES_CPU` should be handled elsewhere (this file is just for actual GPU kernels)

wickedfoo (commented):
Awesome work! Take a look at the comments, main thing is that I think `INDICES_CPU` should be handled as well (to remove the indices as needed on the CPU side), and that whether or not removal is supported at all (the host-side map) should be an option passed upon index construction time.

lailifeng-st:
> To compile the python version, please follow the steps lined out in
> 
> https://github.com/facebookresearch/faiss/blob/main/.circleci/config.yml#L222

thanks, maybe I will try later

asadoughi:
Closing this stale PR as we expect the removal pattern to be satisfied by filtering followed by compaction in most ANN applications.

Issue #1864, opened 2021-05-04
Question: Should index_factory use IndexIDMap2 instead?


michellemay:
# Summary

`index_factory` currently use `IndexIDMap` by default.   Should it use the `IndexIDMap2` that allows reconstruction?

Interface: 
- [x] C++
- [ ] Python

michellemay:
Maybe not the default one, but should be configurable in the factory string ?

mdouze:
you can get an index wrapped in IDMap or IDMap2 with the corresponding index_factory string suffixes. What makes you think that IDMap is used "by default" ?

michellemay:
> you can get an index wrapped in IDMap or IDMap2 with the corresponding index_factory string suffixes. What makes you think that IDMap is used "by default" ?

The code in current master branch explicitely use `IndexIDMap`:  https://github.com/facebookresearch/faiss/blob/bb3c52a057fb5353e043d1f7a41a895879b5c77a/faiss/index_factory.cpp#L463

On the same subject, I saw that IndexIVF have a DirectMap built in.  Again, the index_factory does not allow to specify that setting. (or does it through a different mechanism?)

Maybe I misunderstood something about the factory.. If so, please indulge my ignorance and just point me in the right direction and I'll do my homework properly.

mdouze:
Right, IDMap2 is not supported in the factory (only IDMap is). 
Will mark this as an enhancement.

AbdelrahmanElmeniawy:
Now, Index_factory supports IDMap2 just by adding "IDMap2" to prefix or suffix of the input String!
e.g. "Flat,IDMap2" or "IDMap2,Flat"

PR #2478, opened 2022-09-16
T131797600 make index_factory support IDMap2

AbdelrahmanElmeniawy:
makes index_factory support IDMap2 not only IDMap and add required tests
adding IDMap2 to index_factory would help users to take advantage of its extra features more than IDMap such as reconstruct the indices.

solves [issue 1864](https://github.com/facebookresearch/faiss/issues/1864) 

+fix downcast_index IDMap / IDMap2 order

Test plan:
cd build
make -j
cd faiss/python && python setup.py build
cd ../../..
PYTHONPATH="$(ls -d ./build/faiss/python/build/lib*/)" pytest tests/test_*.py

### DIFF
--- a/faiss/index_factory.cpp
+++ b/faiss/index_factory.cpp
@@ -738,6 +738,14 @@ std::unique_ptr<Index> index_factory_sub(
 
     // IndexIDMap -- it turns out is was used both as a prefix and a suffix, so
     // support both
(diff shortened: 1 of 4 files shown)


mdouze:
Please fix the downcast_index IDMap / IDMap2 order as well.

PR #2489, opened 2022-09-22
Add faiss.IO_FLAG_SKIP_IVFPQ_PRECOMPUTE_TABLE flag for faiss.read_index()

alexanderguzhva:
Summary: Currently, Faiss does not save the precomputed_table for IVFPQ when saving an index, and reconstructs one on the fly upon loading. The introduced flag allows to skip the reconstruction during faiss.read_index()

Differential Revision: D39747978


shiwanghua:
not save to disk, and not compute it when load ? why ? When to compute it ?

PR #2497, opened 2022-09-29
add remove and merge features for IndexFastScan

AbdelrahmanElmeniawy:
* Modify pq4_get_paked_element to make it not depend on an auxiliary table 
* Create pq4_set_packed_element which sets a single element in codes in packed format
(These methods would be used in merge and remove for IndexFastScan
get method is also used in FastScan indices for reconstruction)
* Add remove feature for IndexFastScan
* Add merge feature for indexFast Scan

test plan:
cd build && make -j
make test
cd faiss/python && python setup.py build && cd ../../..
PYTHONPATH="$(ls -d ./build/faiss/python/build/lib*/)" pytest tests/test_*.py

### DIFF
--- a/faiss/impl/pq4_fast_scan.cpp
+++ b/faiss/impl/pq4_fast_scan.cpp
@@ -155,6 +155,42 @@ uint8_t pq4_get_packed_element(
     }
 }
 
+void pq4_set_packed_element(
+        uint8_t* data,
+        uint8_t code,
+        size_t bbs,
+        size_t nsq,
+        size_t vector_id,
+        size_t sq) {
+    // move to correct bbs-sized block
+    // number of blocks * block size
+    data += (vector_id / bbs) * ((nsq / 2) * bbs);
+
+    // get the vector_id inside the block
+    vector_id = vector_id % bbs;
+    bool shift = vector_id > 15;
+    vector_id = vector_id & 15;
+
+    // get the address of the vector in sq
+    size_t address;
+    if (vector_id < 8) {
+        address = vector_id << 1;
+    } else {
+        address = ((vector_id - 8) << 1) + 1;
+    }
+    if (sq & 1) {
+        address += 16;
+    }
+    address = (sq >> 1) * bbs + address;
+    uint8_t temp = data[address];
+    if (shift) {
+        temp = (code << 4) | (temp & 15);
+    } else {
+        temp = code | (temp & ~15);
+    }
+    data[address] = temp;
+}
+
 /***************************************************************
  * Packing functions for Look-Up Tables (LUT)
  ***************************************************************/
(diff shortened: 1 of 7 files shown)


faiss/impl/pq4_fast_scan.cpp
```diff
@@ -122,30 +122,59 @@ void pq4_pack_codes_range(
     }
 }
 
+uint8_t get_address(size_t& bbs, size_t& vector_id, size_t& sq, bool& shift) {
```
mdouze, line 125: passing arguments with & does not make sense here

AbdelrahmanElmeniawy: for "shift", it is mandatory because we need shift in the output.
for other args, I made that to avoid creating new 3 variables with 3 new memory locations  but after rethink, it is not much improvement

tests/test_index_composite.py
```diff
@@ -16,6 +16,43 @@
...
+        xt, xb, _ = get_dataset_2(d, ntotal, ntotal, 0)
+        index = faiss.index_factory(20, 'IDMap2,PQ5x4fs')
+        index.train(xt)
+        index.add_with_ids(xb, np.arange(ntotal).astype("int64"))
+        before = index.reconstruct_n(0, ntotal)
+        index.remove_ids(np.array(removed))
+        for i in range(ntotal):
+            if i in removed:
+                try:
+                    after = index.reconstruct(i)
```
mdouze, line 32: re-use of identifier i
these should all fail, correct?

tests/test_index_composite.py
```diff
@@ -16,6 +16,43 @@
...
+
+    def test_remove_last_vector(self):
+        self.do_test(993, [992])
+
+    # test remove element from every address 0 -> 31
+    def test_remove_every_address(self):
+        temp = np.arange(32).tolist()
+        removed = []
+        for i in range(32):
+            removed.append(i * 32 + temp[i])
```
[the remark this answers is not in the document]
AbdelrahmanElmeniawy: removed is every address(0 -> 32) but in different blocks
e.g. remove vector in address 0 in block 0 , address 1 in block 1 ............
removed = [0, 32 + 1, 2 * 32 + 2, ....]
                   [0,    33   ,       66      , 99, 132, .....]

tests/test_index_composite.py
AbdelrahmanElmeniawy: yes, if this vector i removed before, reconstruction should be failed.
I am not sure what do you mean by "re-use of identifier i"?

tests/test_index_composite.py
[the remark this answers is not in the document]
AbdelrahmanElmeniawy: I realised that my formula is actually 33 *  np.arange(32) 😅 
I will change it

PR #2521, opened 2022-10-10
Integrate IVF-Flat from RAFT

cjnolet:
This is a design proposal that demonstrates an approach to enabling optional support for [RAFT](https://github.com/rapidsai/raft) versions of IVF PQ and IVF Flat (and brute force w/ fused k-selection when k <= 64). There are still a few open issues and design discussions needed for the new RAFT index types to support the full range of features of that FAISS' current gpu index types.  

Checklist for the integration todos:
- [x] Rebase on current `main` branch
- [X] The raft handle has been plugged directly into the StandardGpuResources 
- [X] `FlatIndex` passing Googletests
- [x] Use `CodePacker` to support `copyFrom()` and `copyTo()`
- [X] `IVF-flat passing Googletests
- [ ] Raise appropriate exceptions for operations which are not yet supported by RAFT

Additional features we've discussed:
- [x] Separate IVF lists into individual memory chunks 
- [ ] Saving/loading

To build FAISS w/ optional RAFT support:
```
mkdir build
cd build
cmake ../ -DFAISS_ENABLE_RAFT=ON -DFAISS_ENABLE_GPU=ON
make -j 
```

For development/testing, we've also supplied a bash script to make things easier: `build.sh`

Below is a benchmark comparing the training of IVF Flat indices for RAFT and FAISS: 
image

The benchmark was produced using Googlebench in [this](https://github.com/tfeher/raft/tree/raft_faiss_bench) RAFT fork. We're going to provide benchmarks for the queries as well. There are still a couple bottlenecks to be removed in the IVF-Flat training implementation and we'll update the current benchmark when ready.


mlomeli1:
hi @cjnolet , thank you for working on this! Do you have any plans to remove the dependency on faiss that @wickedfoo flagged in his email? I am including the email content here in case you missed it:

I've been working include and build Raft inside our third-party code repository, so that we can include it from within “fbsource” which is our main code repository in which Faiss is the source of truth.

 One problem is that there exists a circular dependency, as Raft itself depends upon Faiss, namely the knn_brute_force_faiss.cuh header.

How should we handle this dependency? Ideally, there is a way for us to build/use Raft that doesn’t depend upon Faiss, so the dependency only goes one way when we eventually include Raft from within Faiss.

cjnolet:
@wickedfoo If you have a moment, I'm hoping to solicit your initial feedback on this integration path so far. We have a lot of the Googletests for IVF-flat passing, with the exception of the `TestGpuIndexIVFFlat.LargeBatch`, which I'm looking into.

cjnolet:
I've added a view to RAFT's Github project board to track tasks related to this integration: https://github.com/orgs/rapidsai/projects/55/views/14

faiss/gpu/GpuIndex.h
```diff
@@ -23,6 +23,9 @@ struct GpuIndexConfig {
     /// On Pascal and above (CC 6+) architectures, allows GPUs to use
     /// more memory than is available on the GPU.
     MemorySpace memorySpace;
+
+    /// Should the index dispatch down to RAFT?
+    bool use_raft = false;
```
wickedfoo, line 28: Can you add a `#define USE_NVIDIA_RAFT` or similar in this file to enable/disable RAFT compilation and wrap all RAFT-using blocks with it?

faiss/gpu/GpuIndexIVF.cu
```diff
@@ -426,18 +431,40 @@ void GpuIndexIVF::trainQuantizer_(idx_t n, const float* x) {
...
+        printf("Using raft to train quantizer for %d vectors\n", n);
+        const raft::handle_t& raft_handle =
+                resources_->getRaftHandleCurrentDevice();
+
+        raft::neighbors::ivf_flat::index_params raft_idx_params;
+        raft_idx_params.n_lists = nlist;
+        raft_idx_params.metric = raft::distance::DistanceType::L2Expanded;
+        raft_idx_params.add_data_on_build = false;
+        raft_idx_params.kmeans_trainset_fraction = 1.0;
+        raft_idx_params.kmeans_n_iters = 100;
```
wickedfoo, line 446: You can use the same clustering parameters (`IndexIVF::cp`) for this, and also derive `kmeans_trainset_fraction` based on `cp.min_points_per_centroid`/`cp.max_points_per_centroid` and the number of vectors passed for training? Can also look to see if any of the other clustering parameters have analogues in RAFT.

https://github.com/facebookresearch/faiss/blob/main/faiss/Clustering.h#L22

faiss/gpu/GpuIndexIVF.cu
```diff
@@ -426,18 +431,40 @@ void GpuIndexIVF::trainQuantizer_(idx_t n, const float* x) {
...
+
+        auto raft_index = raft::neighbors::ivf_flat::build(
+                raft_handle, raft_idx_params, x, n, (idx_t)d);
+
+        raft_handle.sync_stream();
+
+        // TODO: Validate this is all we need to do
+        quantizer->reset();
```
wickedfoo, line 454: FYI, as of this commit:

https://github.com/facebookresearch/faiss/commit/16d5ec755f3f7615668f160a4f23fcbf3dd6ac7b

both CPU and GPU indices can be used as a coarse quantizer (so it follows the CPU implementation better). For RAFT usage it seems ok to simply not allow pluggable CPU coarse quantizers, but if there's a way to pass in pre-assigned IVF clusters to look at for RAFT IVF indices, that would be nice.

faiss/gpu/GpuIndexIVFFlat.cu
```diff
@@ -110,30 +172,44 @@ void GpuIndexIVFFlat::copyFrom(const faiss::IndexIVFFlat* index) {
...
+        printf("Reconstructing %d original vectors and adding to GPU index\n",
+               ntotal);
+
+        // Quantizer should already have been updated above. Add reconstructed
+        // vectors to raft index
+        if (ntotal > 0) {
+            std::vector<float> buf_host(ntotal * d);
+            std::vector<idx_t> ids(ntotal);
+            std::iota(ids.begin(), ids.end(), 0);
+            index->reconstruct_n(0, ntotal, buf_host.data());
```
wickedfoo, line 205: For non-flat index types (e.g., CPU IVF indices here), arbitrary user IDs can be assigned to vectors. There is no guarantee that they are sequential starting from 0.

I think we may need a "batch reconstruct all vectors and provide all indices stored in the index" function as well in order to support this case. Or, this may simply be a bad idea as the vectors may have already been quantized on the Faiss side of things using SQ or PQ, and then yet again we'd re-quantize on the RAFT side of things post-reconstruction. 

@mdouze  would be interested in your thoughts here too.

faiss/gpu/GpuResources.cpp
```diff
@@ -153,6 +168,10 @@ cudaStream_t GpuResources::getDefaultStreamCurrentDevice() {
     return getDefaultStream(getCurrentDevice());
 }
 
+raft::handle_t& GpuResources::getRaftHandleCurrentDevice() {
```
wickedfoo, line 171: I was about to start on implementing a resources object (whether in StandardGpuResources or in a separate RaftGpuResources) that would route all Faiss GPU memory allocations to the RAFT memory manager as well. Do you think this is still needed, as I think we discussed this a while ago?

Or maybe it's fine just to have the two memory resources separate.

faiss/gpu/StandardGpuResources.cpp
```diff
@@ -375,6 +394,19 @@ cudaStream_t StandardGpuResourcesImpl::getDefaultStream(int device) {
     return defaultStreams_[device];
 }
 
+raft::handle_t& StandardGpuResourcesImpl::getRaftHandle(int device) {
+    initializeForDevice(device);
+
+    auto it = raftHandles_.find(device);
```
wickedfoo, line 400: `initializeForDevice` already constructs something in `raftHandles_` for this device, so we can simply `return raftHandles_[device];` like below instead of this?

faiss/gpu/StandardGpuResources.h
```diff
@@ -183,10 +207,13 @@ class StandardGpuResources : public GpuResourcesProvider {
     /// Export a description of memory used for Python
     std::map<int, std::map<std::string, std::pair<int, size_t>>> getMemoryInfo()
             const;
-
     /// Returns the current default stream
     cudaStream_t getDefaultStream(int device);
 
+    /// Returns the raft handle for the given device which can be used to
+    /// make calls to other raft primitives.
+    raft::handle_t& getRaftHandle(int device);
```
wickedfoo, line 215: What's inside a `raft::handle_t`? Is it bad to copy by value (versus passing a reference here)?

faiss/gpu/impl/RaftFlatIndex.cu
```diff
@@ -0,0 +1,101 @@
...
+    // For now, use RAFT's fused KNN when k <= 64 and L2 metric is used
+    if (k <= 64 && metric == MetricType::METRIC_L2 && vectors_.getSize(0) > 0) {
+        raft::handle_t& raft_handle = resources_->getRaftHandleCurrentDevice();
+
+        auto distance = exactDistance
+                ? raft::distance::DistanceType::L2Unexpanded
+                : raft::distance::DistanceType::L2Expanded;
+
+        auto index = raft::make_device_matrix_view<float>(
+                vectors_.data(), vectors_.getSize(0), vectors_.getSize(1));
```
wickedfoo, line 59: if `FlatIndex::useFloat16_` is true, then the data is stored in `vectorsHalf_` and is of type `half`. Is f32 (queries) x f16 (data) = f32 (restul distance) brute-force kNN supported by RAFT?

Even if f16 is used, we've always been reporting distances as f32 as true nearest neighbors might differ only beyond 2^-10 in the significand fraction.

wickedfoo (commented):
Looks pretty good overall, could we narrow this down to just GpuIndexFlat and GpuDistance to begin with, since I think the question on how (or even if) CPU IVF indices being copied to GPU RAFT or vice-versa is worth thinking about a bit more?

faiss/gpu/GpuResources.cpp
cjnolet: The nice thing about `rmm` (RAPIDS memory manager) is that it allows a user to set a default memory resource for the entire process. I do think it would still be useful if we had an `RmmGpuResources` in addition to the `StandardGpuResource`. The `raft::handle_t` is similar to `GpuResources` in that it's meant to hold on to various expensive resource handles so that they can be reused. The RAFT functions already use the default `rmm` resource for all temporary allocations so if we can have an `RmmGpuResources` right in FAISS then the user can set, for example, a pooled or managed resource using RMM and then both RAFT and FAISS will be using the exact same memory resource. We took a stab at building an `RmmGpuResources` awhile back [here](https://github.com/rapidsai/raft/blob/branch-23.04/cpp/include/raft/spatial/knn/faiss_mr.hpp). Would that be useful?

We would still want to manage a RAFT handle in addition to that though, because the `raft::handle_t` is ultimately what we pass to the RAFT APIs. That said, though, the handle doesn't have to create it's own resources- we can make sure the `raft::handle_t` uses the same resources like the main stream, stream pool, and cublas handle from the `GpuResources` instance.

faiss/gpu/StandardGpuResources.h
cjnolet: It's generally preferable to pass the same handle around (by reference). We've always passed it around that way, anyways within RAFT. Though with some of the recent changes it *may* be able to be passed around by value without causing excess copyies of the underlying resources.

cjnolet:
> Looks pretty good overall, could we narrow this down to just GpuIndexFlat and GpuDistance to begin with, since I think the question on how (or even if) CPU IVF indices being copied to GPU RAFT or vice-versa is worth thinking about a bit more?

Yes, I think that's a great idea. I had started working through the logic of implementing `copyFrom` and `copyTo` to be more like the existing FAISS GPU implementations but since RAFT was using a single chunk of memory for the IVF lists, the logic was getting overly complicated.

We have been working on two important features: [reconstruct_batch](https://github.com/rapidsai/raft/issues/1205) and [splitting IVFs into individual arrays](https://github.com/rapidsai/raft/issues/1170) which should it much more straightforward to use the new `CodePacker` API to implement `copyFrom()` and `copyTo()`. We're also working towards decoupling the coarse quantizers [here](https://github.com/rapidsai/raft/issues/956).

faiss/gpu/GpuIndexIVFFlat.cu
mdouze: Agree with Jeff here. Why not copy the data directly from the inverted lists? Like in 
https://github.com/facebookresearch/faiss/blob/868e17f29493075742170885f1f57c7b9e61d9ea/faiss/gpu/impl/IVFBase.cu#L325

faiss/gpu/GpuResources.cpp
mdouze: Just a high level question: how does that interact with the pytroch mem manager? People often combine pytorch and Faiss and currently the memory management is disjoint.

mdouze (commented):
Please notice that a few fields from GpuIndexIVF will move to a common parent with the CPU IndexIVF this week. 

https://github.com/facebookresearch/faiss/pull/2682/files#diff-facfbde9bd4277e014733ed6342e70693c1c65618947d29ae1929f36a99a16f4

This should not affect this PR significantly.

cjnolet:
> Agree with Jeff here. Why not copy the data directly from the inverted lists? Like in

@mdouze, the reconstruction was just an intermediate solution. We plan to use the new `CodePacker` once https://github.com/rapidsai/raft/issues/1170 is done on the RAFT side. We currently have PRs open for doing this. In the meantime, I can open a different PR with just the FlatIndex/GpuDistance dispatching.

cjnolet:
> Just a high level question: how does that interact with the pytroch mem manager? People often combine pytorch and Faiss and currently the memory management is disjoint.

We just integrated RMM into Pytorch (as of the latest 23.02 release of RAPIDS) so this is now able to be done with:

```python
torch.cuda.memory.change_current_allocator(rmm.rmm_torch_allocator)
```
More details in [this blog](https://medium.com/rapids-ai/pytorch-rapids-rmm-maximize-the-memory-efficiency-of-your-workflows-f475107ba4d4).  I think this will be helpful for FAISS+Pytorch integration as well.

faiss/gpu/GpuIndex.h
cjnolet: @wickedfoo, absolutely, I can do that. Just to clarify- you want to have `USE_NVIDIA_RAFT` instead of the `use_raft` member and not have both, right?

cjnolet: As a follow-on question: Do we want to assume RAFT should always be enabled when `USE_NVIDIA_RAFT` is defined or do we want to keep something that is configurable at runtime?

mdouze: I would vote to keep it configurable at runtime, as it makes comparison easier. 
We could even *not* guard the use_raft field in the .h file and just fail at runtime if USE_NVIDIA_RAFT is not set at compile time. This pushes the test to the .cpp / .cu file.

cjnolet:
Linking in RAFT's ivf-flat codepacker API PR: https://github.com/rapidsai/raft/pull/1632

tarang-jain:
Looks like the CLA check is failing. Does it need to be signed by me **individually**?

mdouze:
Thanks for fixing the CI errors. I think the PR is ready to be accepted.

tarang-jain:
It should compile with raft 23.10. There was a fix for [this issue](https://github.com/rapidsai/raft/issues/1756) that will be rolled out with the 23.10 release. With the 23.08 version of raft, the [LargeBatch googletest](https://github.com/facebookresearch/faiss/blob/c8d6f7bb2b09b8928570222131353f75acb4007b/faiss/gpu/test/TestGpuIndexIVFFlat.cpp#L310) will fail.

tarang-jain:
Does the build.sh need to be removed?

tarang-jain:
### Benchmarking results for Nearest Neighbor Search (Query)
Plot 1:
nTrain = 100000, nAdd = 100000, dim = 128, k = 15, nCentroids = 223, nprobe = 50
image

Plot 2:
nTrain = 100000, nAdd = 100000, dim = 256, k = 15, nCentroids = 223, nprobe = 50
image

Plot 3:
nTrain = 100000, nAdd = 100000, dim = 1024, k = 15, nCentroids = 223, nprobe = 50
image

mdouze:
Sorry for the delay, we are still busy importing the new raft version internally.

PR #2525, opened 2022-10-11
Use packaging.version to fix deprecation warning

Moisan:
This PR fixes the following warning that was raised by `loader.py`

```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```


mdouze:
I am not very keen on accepting this because it adds a dependency for a very simple functionality.

PR #2628, opened 2022-12-16
attempt to use switch..case in index_read

mdouze:
Summary:
to avoid too much spaghetti code, use switch... case but the foucc must be constexpr (does not work) or marco (ugly).
The code is not much better so let's avoid this.

Differential Revision: D42095262


mdouze:
Attempt not successful.

PR #2639, opened 2022-12-26
things should be removed were removed

cemilakkoc:
thats literally it


mdouze:
I find it more clear as written currently.

cemilakkoc:
> I find it more clear as written currently.

I found it not tbh, looks kinda ugly to me

PR #2650, opened 2023-01-05
about a magnitude faster LSQ/PLSQ training and encoding

alexanderguzhva:
Summary:
LocalSearchQuantizer::icm_encode_step was optimized:
* The loop orders were changes
* Memory allocations were optimized
* Different order for accessing 'binaries' tensor, because it is symmetric
* SIMD for the lookup of the best code.

Results are unchanged.

Also, fixes incorrect test in test_local_search_quantizer.py

Differential Revision: D42352124


faiss/impl/LocalSearchQuantizer.cpp
```diff
@@ -604,54 +606,67 @@ void LocalSearchQuantizer::icm_encode_step(
...
-                        // binaries[m, other_m, code, code2]
-                        objs[i * K + code] += binaries[binary_idx];
+                        size_t binary_idx = other_m * M * K * K + m * K * K +
+                                code2 * K + code;
+                        // binaries[m, other_m, code, code2].
+                        // It is symmetric over (m <-> other_m)
+                        //   and (code <-> code2).
+                        // So, replace the op with
+                        //   binaries[other_m, m, code2, code].
+                        objs[code] += binaries[binary_idx];
```
KinglittleQ, line 650: Why we need to replace the access order here?

alexanderguzhva: Because the accesses becomes sequential in for-loop for code variable, thus it allows compiler to use SIMD instructions, according to the assembler code. This access reordering allows to increase the training and encoding speed about ~5x alone, according to my benchmarks.

alexanderguzhva: Btw, the same trick can be applied to the GPU code. Please feel free to alter one, or I can do that as well.

KinglittleQ: Cool, this is an ingenious improvement! You are very welcome to improve the LSQ GPU code, if you have time.

Issue #2658, opened 2023-01-10
Precompute table should be enabled only if l2 quantizer and by_residual is true


chasingegg:
# Summary

There are comments which show that the `// Pre-compute distance tables for IVFPQ with by-residual and METRIC_L2`, but actually we do not precompute distance tables only if quantizer metric type is METRIC_INNER_PRODUCT in faiss/IndexIVFPQ.cpp's `initialize_IVFPQ_precomputed_table`, this would cause memory overhead for other cases. 

# Platform

OS: 

Faiss version: 

Installed from:  

Faiss compilation options: 

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [x] C++
- [ ] Python

# Reproduction instructions

mdouze:
Why do you say that? Here is the code that decides about precomputing the tables: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVFPQ.cpp#L436

chasingegg:
Yeah, but if we choose L2 and by_residual=false, faiss will still precomputing the table, but it will not be used but only involve the memory overhead.

PR #2659, opened 2023-01-10
Enable precomputed table only if L2 quantizer and by_residual is true

chasingegg:
Related issue #2658

### DIFF
--- a/faiss/IndexIVFPQ.cpp
+++ b/faiss/IndexIVFPQ.cpp
@@ -422,6 +422,7 @@ void initialize_IVFPQ_precomputed_table(
         const Index* quantizer,
         const ProductQuantizer& pq,
         AlignedTable<float>& precomputed_table,
+        bool& by_residual,
         bool verbose) {
     size_t nlist = quantizer->ntotal;
     size_t d = quantizer->d;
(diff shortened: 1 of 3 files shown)


faiss/IndexIVFPQ.cpp
```diff
@@ -422,6 +422,7 @@ void initialize_IVFPQ_precomputed_table(
         const Index* quantizer,
         const ProductQuantizer& pq,
         AlignedTable<float>& precomputed_table,
+        bool& by_residual,
```
mdouze, line 425: please pass in by value not reference

chasingegg: Updated.

faiss/IndexIVFPQ.cpp
```diff
@@ -433,10 +434,10 @@ void initialize_IVFPQ_precomputed_table(
     }
 
     if (use_precomputed_table == 0) { // then choose the type of table
-        if (quantizer->metric_type == METRIC_INNER_PRODUCT) {
+        if (!(quantizer->metric_type == METRIC_L2 && by_residual)) {
             if (verbose) {
                 printf("IndexIVFPQ::precompute_table: precomputed "
-                       "tables not needed for inner product quantizers\n");
+                       "tables not needed only if L2 quantizer and by_residual is enabled\n");
```
mdouze, line 440: please change this to " tables needed only for  L2 metric and by_residual is enabled" and make sure the line is < 80 characters.

chasingegg: Done.

PR #2712, opened 2023-02-16
faiss: use autovectorization for inner product

DenisYaroshevskiy:
Summary:
Using autovectorization to get the proper urnolling.

Previous version timings are:

Before
```
faiss_ip_10000                        2.10us   475.62K
faiss_n2_10000                        4.23us   236.30K
```

After
```
faiss_ip_10000                         1.21us   827.16K
faiss_n2_10000                       640.68ns     1.56M
```

Differential Revision: D43353199

### DIFF
--- a/faiss/utils/distances_simd.cpp
+++ b/faiss/utils/distances_simd.cpp
@@ -97,11 +97,32 @@ float fvec_inner_product_ref(const float* x, const float* y, size_t d) {
     return res;
 }
 
+float fvec_inner_product(const float* x, const float* y, size_t d) {
+    float res = 0.F;
+    for (size_t i = 0; i != d; ++i) {
+#pragma float_control(precise, off)
+        res += x[i] * y[i];
+    }
+    return res;
+}
+
 float fvec_norm_L2sqr_ref(const float* x, size_t d) {
     size_t i;
     double res = 0;
     for (i = 0; i < d; i++)
         res += x[i] * x[i];
+    return res;
+}
+
+float fvec_norm_L2sqr(const float* x, size_t d) {
+    // the double in the _ref is suspected to be a typo. Some of the manual
+    // implementations this replaces used float.
+    float res = 0;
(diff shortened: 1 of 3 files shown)


zcfh:
May I ask which benchmark you are running?

DenisYaroshevskiy:
> May I ask which benchmark you are running?

Sorry for not publishing benchmarks, it's not terribly fancy.
It's something like this: https://godbolt.org/z/Y1oj7xsjo
Quick Bench doesn't seem to work at the moment, or I'd share that.

zcfh:
Thanks, I'll try it

zcfh:
> > May I ask which benchmark you are running?
> 
> Sorry for not publishing benchmarks, it's not terribly fancy. It's something like this: https://godbolt.org/z/Y1oj7xsjo Quick Bench doesn't seem to work at the moment, or I'd share that.

Are there any requirements for the compiler version?
I used clang-17 but could not reproduce the optimization effect.

zcfh:
> > May I ask which benchmark you are running?
> 
> Sorry for not publishing benchmarks, it's not terribly fancy. It's something like this: https://godbolt.org/z/Y1oj7xsjo Quick Bench doesn't seem to work at the moment, or I'd share that.

What compiler version and compile options are you using?

zcfh:
>  Is there something you are trying to do?
`fvec_inner_product` is a hot function in my program. I am looking for optimization points for it, and I saw this commit mentioning that using automatic vectorization can bring better performance. I only improved by 40% on the local test benchmark, so I want to know if my compiler or compilation options are wrong.

> To the best of my knowledge, this codegen: https://godbolt.org/z/9e3TMozqc is pretty good
Compiler and the previous code do roughly the same thing but compiler unrolls - which is important in high IPC code like this.

It seems that no special options are used, so I will try to analyze the reason myself.

DenisYaroshevskiy:
The specific improvement will vary by the machine.
Are you using march=native or some similar option?

PR #2722, opened 2023-02-20
Build with generic docker image

mdouze:
try ubunu


cjnolet:
@mdouze we do have a rapids container that contains all the RAFT dependencies. It's pretty heavy as it contains all of rapids so I'm not sure how much it would help. I am happy to help provision a `Dockerfile` with the needed dependencies to build w/ RAFT, though when using raft header-only, `rapids-cmake` takes care of downloading many of the additional dependencies upstream and making cmake targets available.

mdouze:
IIUC there are 3 options here: 
1. Use default containers and install dependencies via Dockerfiles 
2. Use default containers and install dependencies via commands in circleci (or cmake)
3. Use specialized containers with dependencies (in particular nvidia) already installed

My preference would be for option 2 because the commands are useful too for people who compile Faiss without docker. Asking the other team members for their opinion...

cjnolet:
@mdouze option 2 does sound like the most reasonable to me

mdouze:
superseded by @algoriddle's work

Issue #2727, opened 2023-02-23
Expose IDSelector filtering through c_api


kaelen-moda:
I was wanting to play around with subset filtering on my searches using some go bindings against the [c_api](https://github.com/facebookresearch/faiss/tree/main/c_api).

Unfortunately, it doesn't seem like the [search header](https://github.com/facebookresearch/faiss/blob/main/c_api/Index_c.h#L103) is fully in line with the [internal implementation](https://github.com/facebookresearch/faiss/blob/main/faiss/Index.h#L124-L130) that exposes SearchParameters which contains the IDSelector filter.

Am I missing something obvious or are these out of sync? Is there a particular issue that is preventing the c_api from being in sync? The optional params? Would something like faiss_Index_search_w_params be doable here? 

https://github.com/facebookresearch/faiss/pull/2483

Enet4:
> Is there a particular issue that is preventing the c_api from being in sync?

The issue is that the process of synchronizing the C API is purely manual and depends on volunteer work. The additional parameter in `search` was only introduced five months ago. With that said, a new function with the extra parameter would be doable and welcomed, so long as the `SearchParameters` and ID selector types are also given their C version in the API.

kaelen-moda:
Thanks for the fast reply @Enet4! I worked on the implementation and submitted PR here: https://github.com/facebookresearch/faiss/pull/2732 

Happy to take some C/C++ style pointers as I'm pretty rusty  (puns not intended but welcome) :)

PR #2732, opened 2023-02-27
Search params c_api implementation

kaelen-moda:
https://github.com/facebookresearch/faiss/issues/2727

Implements search_with_params function on c_api for index.

Implemented c_api equivalents of SearchParameters and SearchParametersIVF. 

My C/C++ is pretty rusty so I imagine my arguments to the new functions for each search parameters could be refined. Happy to take suggestions :)

### DIFF
--- a/c_api/example_c.c
+++ b/c_api/example_c.c
@@ -109,6 +109,8 @@ int main() {
         }
         free(I);
         free(D);
+        free(sel);
+        free(params);
     }
 
     printf("Saving index to disk...\n");
--- a/c_api/example_c.c
+++ b/c_api/example_c.c
@@ -111,8 +111,8 @@ int main() {
         }
         free(I);
         free(D);
-        free(sel);
-        free(params);
+        faiss_SearchParameters_free(params);
+        faiss_IDSelectorRange_free(sel);
     }
 
     printf("Saving index to disk...\n");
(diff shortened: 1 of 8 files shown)


mdouze:
@Enet4 could you look at this PR? Thanks!

c_api/example_c.c
```diff
@@ -92,6 +93,27 @@ int main() {
...
+        printf("I=\n");
+        for (int i = 0; i < 5; i++) {
+            for (int j = 0; j < k; j++)
+                printf("%5lld (d=%2.3f)  ", I[i * k + j], D[i * k + j]);
+            printf("\n");
+        }
+        free(I);
+        free(D);
+        free(sel);
+        free(params);
```
Enet4, line 115: While it might appear to work, objects constructed through functions in the API should also be freed through the API.
```suggestion
        faiss_SearchParams_free(params);
        faiss_IDSelectorRange_free(sel);
```

Enet4 (commented):
No major objections to this. I only noticed one thing worth fixing before it's merged.

c_api/example_c.c
kaelen-moda: @Enet4 Good catch, updated with wrapper free's

kaelen-moda:
Hey @Enet4, does this PR look good to you?

Enet4 (approved):
Seems to be in order now.

PR #2742, opened 2023-03-02
Proposal IDSelectorCombination

kaelen-moda:
Adds support for an IDSelector that takes in two IDSelectors and can perform a boolean operation on their is_member outcomes.

Current implementation is pretty naive and doesn't try to do any optimizations on the types of IDSelectors combined.

Also test cases are definitely lacking but can add more once approach is agreed upon.

### DIFF
--- a/faiss/python/__init__.py
+++ b/faiss/python/__init__.py
@@ -189,9 +189,9 @@ add_ref_in_constructor(BufferedIOWriter, 0)
 add_ref_in_constructor(BufferedIOReader, 0)
 
 add_ref_in_constructor(IDSelectorNot, 0)
-add_ref_in_constructor(IDSelectorAnd, 0)
-add_ref_in_constructor(IDSelectorOr, 0)
-add_ref_in_constructor(IDSelectorXOr, 0)
+add_ref_in_constructor(IDSelectorAnd, slice(2))
+add_ref_in_constructor(IDSelectorOr, slice(2))
+add_ref_in_constructor(IDSelectorXOr, slice(2))
 
 # seems really marginal...
 # remove_ref_from_method(IndexReplicas, 'removeIndex', 0)
--- a/faiss/python/__init__.py
+++ b/faiss/python/__init__.py
@@ -189,6 +189,9 @@ add_ref_in_constructor(BufferedIOWriter, 0)
 add_ref_in_constructor(BufferedIOReader, 0)
 
 add_ref_in_constructor(IDSelectorNot, 0)
+add_ref_in_constructor(IDSelectorAnd, 0)
+add_ref_in_constructor(IDSelectorOr, 0)
+add_ref_in_constructor(IDSelectorXOr, 0)
 
 # seems really marginal...
 # remove_ref_from_method(IndexReplicas, 'removeIndex', 0)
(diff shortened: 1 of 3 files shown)


mdouze:
Looks good, would you mind making one `IDSelector` class for each binary operation? Eg. `IDSelectorAnd`, `IDSelectorOr`, etc.? This would be easier to read. 
Please also add reference to the sub-classes like here: 

https://github.com/facebookresearch/faiss/blob/main/faiss/python/__init__.py#L191

kaelen-moda:
Thanks for the review @mdouze. I agree with your suggestion about a unique IDSelector for each binary operation. Looks much better now. 

Added the python changes following the IDSelectorNot example. I sort of get the purpose but I'm not fully confident in my understanding so please check that to ensure correctness.

tests/test_search_params.py
```diff
@@ -81,6 +99,18 @@ def do_test_id_selector(self, index_key, id_selector_type="batch", mt=faiss.METR
                 if i not in ssubset
             ]).astype('int64')
             sel = faiss.IDSelectorNot(faiss.IDSelectorBatch(inverse_subset))
+        elif id_selector_type == "or":
+            lhs_sel = faiss.IDSelectorBatch(lhs_subset)
+            rhs_sel = faiss.IDSelectorBatch(rhs_subset)
+            sel = faiss.IDSelectorOr(lhs_sel, rhs_sel)
```
mdouze, line 105: to check if the python references worked correctly, please use 

```
sel = faiss.IDSelectorOr(
    faiss.IDSelectorBatch(lhs_subset), 
    faiss.IDSelectorBatch(rhs_subset)
)
```

This will crash if the references are not tracked correctly

mdouze (commented):
please address 2 points in the code

faiss/python/__init__.py
```diff
@@ -189,6 +189,9 @@ def replacement_function(*args):
 add_ref_in_constructor(BufferedIOReader, 0)
 
 add_ref_in_constructor(IDSelectorNot, 0)
+add_ref_in_constructor(IDSelectorAnd, 0)
```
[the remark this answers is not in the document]
kaelen-moda: Updated with slice(2), having trouble running these tests locally so currently having to rely on CI to check.

QQ but why wouldn't the IDSelectorNot follow a similar pattern or since it only needs one reference a single 0 is sufficient?

tests/test_search_params.py
kaelen-moda: Not obvious the distinction between the two as my python is pretty basic, could you explain or provide a pointer to the difference between the two?

faiss/python/__init__.py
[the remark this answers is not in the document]
mdouze: Right about IDSelectorNot

tests/test_search_params.py
mdouze: I added a bit of doc about this in the wiki 

https://github.com/facebookresearch/faiss/wiki/Faiss-code-structure#object-ownership-python

kaelen-moda: Thanks for the link! Makes more sense now.

PR #2755, opened 2023-03-10
Id selector c api

kaelen-moda:
Implements c_api interfaces for IDSelectorNot, IDSelectorAnd, IDSelectorOr, IDSelectorXOr


kaelen-moda:
Another c_api change, appreciate a review from you @Enet4 :)

Enet4 (approved):
I could not find anything standing out as incorrect, so if the example works, then it should be fine.

kaelen-moda:
Thanks for review @Enet4 . Yes, the example works well from my runs

PR #2771, opened 2023-03-17
Fix bug of HNSW

hhy3:


hhy3:
> Could you explain what this is?

When popping the first element of the heap, it could be invalid. So in this case `nvalid` should not decrease by 1. So it should be checked first. Otherwise, the performance would be impaired.

mdouze:
I believe this can happen only if a -1 id gets pushed on the heap, which does never happen, correct?

hhy3:
> I believe this can happen only if a -1 id gets pushed on the heap, which does never happen, correct?

It'll be set to -1 when popping the min element. https://github.com/facebookresearch/faiss/blob/main/faiss/impl/HNSW.cpp#L877
And after I made this modification the recall improved a lot.

alexanderguzhva:
Confirmed.

mdouze:
Thanks a lot for this fix! it increases the accuracy of  HNSW by a lot!

PR #2779, opened 2023-03-24
Relax Time Limit for some CI jobs

wx257osn2:
It seems that recent CI pipelines were failed due to timeout of `Linux GPU nightlies (CUDA 10.2)` , so I have made `no_output_timeout` doubled. (Reason to _doubled_ instead of 1.5x or something else is... well, I don't have knowledge to make a decision that this is excessive or not enough, but anyway `60m` is not enough so...)

**EDITED**: `OSX (ARM) nightlies` was also failed due to timeout, so I relaxed it to 30min in the same way.

This PR doesn't affect the software behavior.


wx257osn2:
It seems that the problem has been solved by #2798, so I close this PR.

Issue #2790, opened 2023-03-26
Bug: Build Frome Srouce 'GpuIndexIVFPQ' object has no attribute 'setNumProbes'


matchyc:
# Summary

Building Frome Srouce 'GpuIndexIVFPQ' object has no attribute 'setNumProbes'

Run `bench_gpu_sift1m.py` 
**Exact Search** Totally OK
**Approximate Search** with IVFPQ  **NO**:  'GpuIndexIVFPQ' object has no attribute 'setNumProbes'

# Platform

OS:  Ubuntu 20.04

Faiss version: 1.7.3 (main branch)

Installed from:  Build From Source
 

Faiss compilation options: 

Running on:
- [ ] CPU
- [x] GPU

Interface: 
- [ ] C++
- [x] Python

# Reproduction instructions

**Strictly** Follow `INSTALL.md` Section "Building from source"

except cmake step command:
cmake -B build . -DFAISS_ENABLE_GPU=ON -DFAISS_ENABLE_PYTHON=ON -DCMAKE_BUILD_TYPE=Release -DPython_EXECUTABLE=\<my conda environment interpreter path\>

- cmake version: cmake version 3.26.0
- gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1) 
- CUDA version: 11.4

Note: 
1. Working well using GpuFlatIndex
2. Working well if faiss installed by conda (conda install faiss-gpu)

mdouze:
the function was removed, please set the number of probes directly, with 

```
index.nprobe = 123
```

PR #2797, opened 2023-03-27
remove setNumProbes

mdouze:
Summary:
This is the last code instance of setNumProbes
Removing because some people still seem run into errors due to this.

Differential Revision: D44421600


mdouze:
related to #2790

PR #2889, opened 2023-05-31
Fix windows CI

wx257osn2:
#2882 added [a for loop, which has unsigned index, qualified with `#pragma omp parallel for`](https://github.com/facebookresearch/faiss/pull/2882/files#diff-5a89dcb99a1cce3f297c7f7dfc8e221306b281d4ced6dac1e0fc0fa54188195fR449-R452), but it seems that [MSVC doesn't support unsigned index with `#pragma omp parallel for`](https://app.circleci.com/pipelines/github/facebookresearch/faiss/4220/workflows/ee72de05-6ead-42d9-8ec5-44772e9fd41b/jobs/22529?invite=true#step-104-333) (I think this would not be conformed to OpenMP specification, but...)

I (finally) change the loop with signed index. This changes introduce the precondition `n <= std::numeric_limits<std::make_signed_t<std::size_t>>::max()` , but usually this is `true` I think, so I just put this limitation as a comment instead of any `FAISS_ASSERT` or something like that.

### DIFF
--- a/faiss/utils/utils.cpp
+++ b/faiss/utils/utils.cpp
@@ -28,6 +28,7 @@
 #include <omp.h>
 
 #include <algorithm>
+#include <type_traits>
 #include <vector>
 
 #include <faiss/impl/AuxIndexStructures.h>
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@wx257osn2 , thanks for your commit. Which MSVC version is this?

wx257osn2:
@alexanderguzhva According to [build log of CI](https://circleci.com/api/v1.1/project/github/facebookresearch/faiss/22604/output/104/0?file=true&allocation-id=64773ec8f0f54c7e978de38c-0-build%2F1E1Y0BUL), 

> `** Visual Studio 2019 Developer Command Prompt v16.11.18`
> `-- The CXX compiler identification is MSVC 19.16.27048.0`

PR #2893, opened 2023-06-05
Install C API Library on `make install`

njwhite:
Currently only the C API headers are installed when you run `make install` after configuring with `-DFAISS_ENABLE_C_API=ON`.


alibeklfc:
Thanks! These install rules have since landed in c_api/CMakeLists.txt (lines 127-159), so closing.

PR #2919, opened 2023-06-20
Fix warnings

daniel-ju:
Fixes a type mismatch and unreferenced parameter warning.

### DIFF
--- a/faiss/Index.h
+++ b/faiss/Index.h
@@ -65,7 +65,7 @@ struct Index {
     using component_t = float;
     using distance_t = float;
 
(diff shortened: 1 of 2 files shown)


mdouze:
The size of field d cannot be changed otherwise the file format will break.

daniel-ju:
> The size of field d cannot be changed otherwise the file format will break.

@mdouze Could you point me to why the file format would break? I don't quite follow.
If it cannot be changed, can we perform a static cast when assigning field `d` instead to resolve the warning?

PR #2886, opened 2023-05-31
Add sve targets

vorj:
related: #2884 

This PR contains below changes:

- Add new optlevel `sve`
    - ARM SVE is _extension_ of ARMv8, so it should be treated similar to AVX2 IMO
- Add targets for ARM SVE, `faiss_sve` and `swigfaiss_sve`
    - These targets will be built when you give `-DFAISS_OPT_LEVEL=sve` at build time
    - Design decision: Don't fix SVE register length.
        - The python package of faiss is "fat binary" (for example, the package for avx2 contains `_swigfaiss_avx2.so` and `_swigfaiss.so`)
        - SVE is scalable instruction set (= doesn't fix vector length), but actually we can specify the vector length at compile time.
            - [with `-msve-vector-length=` option](https://developer.arm.com/documentation/101726/4-0/Coding-for-Scalable-Vector-Extension--SVE-/SVE-Vector-Length-Specific--VLS--programming)
            - When this option is specified, the binary can't work correctly on the CPU which has other vector length rather than specified at compile time
        - When we use fixed vector length, SVE-supported faiss python package will contain 7 shared libraries like `_swigfaiss.so` , `_swigfaiss_sve.so` , `_swigfaiss_sve128.so` , `_swigfaiss_sve256.so` , `_swigfaiss_sve512.so` , `_swigfaiss_sve1024.so` , and `_swigfaiss_sve2048.so` . The package size will be exploded.
        - For these reason, I don't specify the vector length at compile time and `faiss_sve` detects the vector length at run time.
- Add a mechanism of detecting ARM SVE on runtime environment and importing `swigfaiss_sve` dynamically
    - Currently it only supports Linux, but there is no SVE environment with non-Linux OS now, as far as I know

NOTE: I plan to make one more PR about add some SVE implementation after this PR merged. This PR only contains adding sve target.

PR #2931, opened 2023-06-28
Add BMI2; Add LTO; Upgrade SQ4, SQ6, SQ8 for AVX2

alexanderguzhva:
This PR does the following:
* Adds `FAISS_USE_BMI2` option for CMake that enables BMI2 instruction set for x86. 
  * This option is not enabled by default, because BMI2 instructions are known to be very slow on all AMD CPUs prior to Zen3 architecture. 
  * This option may be enabled safely on Intel CPUs
  * Internal Meta infra needs to define `USE_BMI2` compiler directive and add `-mbmi2` to compiler flags in TARGETS file
* Adds `FAISS_USE_LTO` option for CMake which enables Link-Time Optimization, which should improve the performance over the whole library. This option is not enabled by default, because it increases the compilation time.
  * Internal Meta infra does not need to do anything
* Adds BMI2 code for SQ4 and SQ6, and upgrades the code for SQ8. The IPC (instructions per cycle, the larger, the better) values for `decode_8_component()` calls according to godbolt.org and uica.uops.info are the following:
  * SQ4, clang-15: baseline IPC is 2.42, candidate IPC is 4.93
  * SQ6, clang-15: baseline IPC is 3.94, candidate IPC is 4.89
  * SQ8, clang-15: baseline IPC is 2.49, candidate IPC is 3.20
* Adds missing `-mpopcnt` and `-mf16c` compiler flags to `CMakeLists.txt` for unit tests along with BMI2 ones.  

The actual performance gain in a production environment with BMI2 enabled needs to be measured, but I don't have the appropriate hardware, only a low-power laptop :)

### DIFF
--- a/faiss/MatrixStats.cpp
+++ b/faiss/MatrixStats.cpp
@@ -12,7 +12,7 @@
 #include <stdarg.h> /* va_list, va_start, va_arg, va_end */
 
 #include <faiss/utils/utils.h>
-#include <inttypes.h>
+#include <cinttypes>
 #include <cmath>
 #include <cstdio>
 
--- a/tests/test_residual_quantizer.py
+++ b/tests/test_residual_quantizer.py
@@ -1117,7 +1117,7 @@ class TestCrossCodebookComputations(unittest.TestCase):
             K = 1 << rq.nbits.at(m)
             cpp_table = codebook_cross_prods[ofs:ofs + K * kk].reshape(kk, K)
             ofs += kk * K
-            np.testing.assert_allclose(py_table, cpp_table, rtol=1e-4)
+            np.testing.assert_allclose(py_table, cpp_table, rtol=2e-4)
 
         cent_norms = faiss.vector_to_array(rq.cent_norms)
         np.testing.assert_array_almost_equal(


mdouze:
Thanks! 

Does BMI2 imply AVX2 ? It would be good to have a monotonous set of CPU features... 
There is a PR for ARM that also adds SVE support (https://github.com/facebookresearch/faiss/pull/2886) + your AVX512 code. 
So it seems we are going to support more and more SIMD sub-features.

alexanderguzhva:
you may wish to keep the CMakeLists.txt as it is, and sometimes later in the future, as soon as modern processors will provide a good support for BMI2, this BMI2 option becomes a default one and gets merged with AVX2. Because, technically, AVX2 is identical to x86 option.
As of internal Meta usage, it is a good idea to enable this BMI2.

mdouze:
What I was thinking of is to have dynamic dispatching to CPU implementations for selected functions. 
The device-specific implementations would be in say a separate namespace, eg. 

````
distances.cpp 
faiss::fvec_norms() 
dynamic dispatch according to cpuid

distances_avx2.cpp (compiled with -mavx2)
--> faiss::avx2::fvec_norms()

distances_sve.cpp (compiled with -msve)
--> faiss::sve::fvec_norms()

````

the current version where we re-compile all the code for separate flavors is not very scalable. WDYT?

alexanderguzhva:
1. The fastest code is obtained by doing a compilation for a given
platform, because the most interesting pieces of the code are the fastest
when ones are inlined. Otherwise, a dynamic dispatch may create an
additional overhead, unless the functions are large and virtual. Maybe,
Faiss should have IndexScalarQuantizerDefault, IndexScalarQuantizerAVX2,
IndexScalarQuantizerSVE internally? Sure doable, but this requires some
integration of C++ templates or macros.
2. There is nothing wrong about providing multiple versions of a library,
say, for PyTorch https://pytorch.org/get-started/locally/.  So, the root of
the evil is in the lack of ppl to take care of this, which is a known
feature.
3. gcc and clang have a feature called 'function multiversioning' that does
exactly this. Check
https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html .
Unfortunately, it is not supported in msvc. Also, I see that it is useful
for somewhat large functions like gemm that can exist as a separate entity,
but I don't know whether it is suitable for small always-inline functions,
because I've never used this feature. Technically, it is possible to go
this path and declare that the MSVC compiler is no longer supported and
that MinGW needs to be used for Windows, which triggers its own side
effects. Or, maybe, disable such a versioning for Windows.
4. There is a somewhat complicated approach for pyfaiss. It is possible to
compile a C++ library right during `pip install` \ `conda install` as some
packages do. Although, I don't believe that it is the right way to solve
the original problem.
5. One more way is to simplify the compilation of Faiss to the ultimate
degree so that it would be possible for anyone to compile Faiss easily.
6. this is how FFmpeg does it, say,
https://github.com/FFmpeg/FFmpeg/tree/master/libavcodec
7. this is how ggml does it,
https://github.com/ggerganov/ggml/blob/master/src/ggml.c
8. One more suggestion. Hardware with SVE support is not easily found. How
should unit tests be run?
9. Can a python wrapper do such a cpuid-based selection and load of one
several DLLs provided in a package?

The easiest way is to stick with the current policy (provide default
generic / AVX2 / NEON versions and allow users to build a faster one of
their own). I could imagine going with platform-specific versions of every
Index class, but it somewhat similar to packing several platform-specific
versions inside a single file.


чт, 29 июн. 2023 г. в 17:03, Matthijs Douze ***@***.***>:

vorj:
@mdouze @alexanderguzhva IMHO, it should be completely separated binaries for each architecture options in order to keep the execution speed (unfortunately). Current code of Faiss was changed like [`code_distance`](https://github.com/facebookresearch/faiss/commits/main/faiss/impl/code_distance), [`hamming_distance`](https://github.com/facebookresearch/faiss/commits/main/faiss/utils/hamming_distance), and so on by @alexanderguzhva, and this is one of major separation for supporting multiple architecture (another one is creating directory `arch` and separating directories by each architecture). Wouldn't it be enough to isolate architecture-intermingled code of current Faiss e.g. `distances_simd.cpp`  like that?

---

> 8. One more suggestion. Hardware with SVE support is not easily found. How should unit tests be run?

Graviton3 on AWS EC2 is one of major environment supporting SVE, but in the context of CI, your suggestion is correct because currently Circle CI doesn't provide runners on c7g instances... QEMU will help most test cases, but [tests for performance regression](https://github.com/facebookresearch/faiss/blob/69cb87768392dc6fd871a9aa8dd42187622ddcdb/tests/test_fast_scan.py#L40-L76) will be failed.

> 9. Can a python wrapper do such a cpuid-based selection and load of one several DLLs provided in a package?

There are some projects like [`workhorsy/py-cpuinfo`](https://github.com/workhorsy/py-cpuinfo) and [`fpelliccioni/cpuid-py`](https://github.com/fpelliccioni/cpuid-py). It would also be possible to achieve this by creating python wrapper for [`pytorch/cpuinfo`](https://github.com/pytorch/cpuinfo).

> default generic / AVX2 / NEON versions

**This is a over details** , but _NEON version_ doesn't exist because NEON (accurately, ASIMD) is part of ARMv8 ISA, not extension (actually **real** _NEON_ is the extension of ARMv7, but current Faiss doesn't support ARMv7 so we don't need to consider it). Currently there are only `generic` and `avx2` , then NEON is always enabled on `generic` in aarch64. #2886 will adds new optimization flag `sve` , because SVE is extension of ARMv8.

PR #2974, opened 2023-07-27
Make parallel OMP execution for add_with_ids in IndexIVF

andriygav:
When working with faiss, I found a significant speedup add_with_ids method of faiss indexing by parallelizing the cluster search step for indexed vectors.

Before the update, indexing of 5000 vectors took ~3 minutes, after the update ~1 minute. Before the update, 3/4 cores were idle most of the time, after the update, all the cores were maximum loaded during the time of quantizer->assign in add_with_ids method in IndexIVF index type.

Value bs is set to 256 according to the experiment on my index. In general, it can be calculated by using the number of available cores for OMP.

Index type: OPQ32_128,IVF262144,PQ32 on CPU .
Virtual Machine: 4 cores, 16 GB RAM.

P.S. Also it is small change to #2973 PR. Use OMP not for all indexes but only to IVF index.


mdouze:
Thanks for the PR, but the use case that is addressed here is very unusual: you use a flat coarse quantizer that is very large (262k centroids). Therefore, I don't think this PR is useful in practice. See 

https://github.com/facebookresearch/faiss/wiki/Guidelines-to-choose-an-index

andriygav:
@mdouze I completely disagree with you. Why is a large index very unusual?

We use faiss index for indexing millions vectors now, and soon we will use it for billions vectors. Low indexes with 4k clusters are unusual indexes in real life. In real life, we have billions of vectors, but not thousands.

A large index size cannot be indexed all at once request, so it must be indexed by using batches. And as an example we use 5k batch size.

The main problem that solved this PR is that faiss not use all available core resources during assign clusters to vector. Perhaps this can be solved more precisely, but this part of the code https://github.com/facebookresearch/faiss/blob/main/faiss/utils/distances.cpp#L263 is not executed in parallel for each vector and this is a big problem. This PR solved this problem at high execution level in assign function, but can also be fixed at low execution level by using parallel distances calculation for each vector independently.

mdouze:
OPQ32_128,IVF262144,PQ32 is not a practical index because it uses brute force assignment on the quantizer side. This explains why adding 5000 vectors takes 1 to 3 minutes. It should be something like OPQ32_128,IVF262144_HNSW32,PQ32. 
Besides, there is no reason why 3/4 compute cores are idle. This suggests that Faiss and especially MKL was not installed properly, please try to use the conda packages instead.

andriygav:
Thanks for the response! I will try using this type of index.

I think the reason of idle in this line of code https://github.com/facebookresearch/faiss/blob/main/faiss/utils/distances.cpp#L263-L290. That line of code are not parallelised for each vector. That why I used OMP for parallelised it.

Issue #2978, opened 2023-07-30
Is the comment of HNSW::shrink_neighbor_list wrong?


yockie:
https://github.com/facebookresearch/faiss/blob/821a401ae9ecee7f6b18c5dea70021a88acbc88b/faiss/impl/HNSW.cpp#L225C42-L225C42

I read HNSW::shrink_neighbor_list source very carefully, I'm sure that this comment is wrong. It enumerates vertices from nearest to farthest, not from farthest to nearest.

Here's my analysis：
（1）notice the definition of NodeDistFarther
```
struct NodeDistFarther {
    float d;
    int id;
    NodeDistFarther(float d, int id) : d(d), id(id) {}
    bool operator<(const NodeDistFarther& obj1) const {
        return d > obj1.d;
    }
};

std::priority_queue<NodeDistFarther> input
```
（2）
It means that the top of input is the node that d is the minimum，and d means the distance。
The smaller the d, the closer the distance（It is also true for inner product metric because d is the opposite of the inner product value）。
（3）Here is a simple test of "std::priority_queue\<NodeDistFarther\>"
![image](https://github.com/facebookresearch/faiss/assets/2568971/7d710ac1-84f7-4d31-8692-2f8c22dd85c0)

mdouze:
Hmmm it looks like you're right. 
Would you mind submitting a PR with the corrected comment?

yockie:
> Hmmm it looks like you're right. Would you mind submitting a PR with the corrected comment?

OK，here is the PR： #2980

PR #2980, opened 2023-07-31
fix hnsw shrink_neighbor_list comment

yockie:
This pr is to fix the issue #2978 .


mlomeli1:
please fill in the CLA @yockie so we can merge it, thanks!

PR #2983, opened 2023-08-01
simplify raft build

algoriddle:


CMakeLists.txt
```diff
@@ -63,9 +63,7 @@ if(FAISS_ENABLE_GPU)
 endif()
 
 if(FAISS_ENABLE_RAFT)
-  rapids_cpm_init()
-  include(cmake/thirdparty/get_raft.cmake)
-  include(cmake/thirdparty/get_cutlass.cmake)
+  find_package(raft COMPONENTS compiled distributed)
```
algoriddle, line 66: @cjnolet I wanted to try this, based on the documentation: https://docs.rapids.ai/api/raft/stable/build/#using-c-pre-compiled-shared-libraries

Wouldn't this be the preferred and simplest way to add the dependencies?

cjnolet: Hey @algoriddle, We probably want to keep the cpm stuff here so that RAFT will automatically be built if it's not already installed. I suspect the issue we're facing is that two different versions of RAFT are being installed (e.g. different version from conda than what's being pulled inside cpm) and this is putting the headers from both on the search path, thus causing build issues.

cjnolet: In fact, I'm pretty sure that's the case here because if the same version of RAFT were being downloaded from conda that was specified in `cpm` then removing one of them wouldn't have fixed a failing RAFT build. I think we should lock both to 23.08 nightlies.

cjnolet: @algoriddle, it looks like we're fetching 23.06 for `cpm`. Are we installing 23.08 from conda by chance?

algoriddle: Is there an advantage of building RAFT if it's not installed? The Faiss build scripts don't build other dependencies (eg. MKL) either, but we expect them to be installed in the build environment. The downside of building RAFT when not available is that we can accidentally create a dependency on the locally built version - either a compile time or runtime dependency, neither of which is desired. We would like to only depend on the version of libraft that we explicitly install and compile with.

We install the nightly of libraft in the environment in what we call the 'cmake' contbuild (which doesn't use `conda build`, but invokes cmake+make directly. See the initialization of the build environment here: https://github.com/facebookresearch/faiss/blob/main/.circleci/config.yml#L191). I suggest  we keep it this way (use the nightly of libraft), because it will allow us to notice breaking changes in upstream.

When I will create the workflow for building a conda package, I will be pinning the version of libraft to an appropriate version upstream, which I presume will be 23.08 initially. In that case the version pinning will be done in the conda package specification (meta.yaml). We will want the cmake script to build against that version of libraft and create a conda package dependency.

cjnolet: @algoriddle 

> I suggest we keep it this way

Yeah that's totally fine w/ me. In the FAISS docs, we could just point the users to RAFT itself if they want to build it and install it from source.

cjnolet: @algoriddle, we have also been kicking around the idea on RAFT to produce a static link library, which might actually make it easier for FAISS since 
1) only the actual used bits would be included in the FAISS binary, and 
2) downstream users of pre-build FAISS binaries would not have to worry about installing RAFT at all. 

The second part there would also alleviate the potential that a user might have a different version fo RAFT already installed on their system when using FAISS. 

Would there be any interest in using a static link library for RAFT if we produce one?

algoriddle: Is there any risk of static linking? If an end user uses faiss, but also loads libraft dynamically into a process (because they use some libraft features directly), would there be any problems or limitations imposed on the use of faiss (with a statically linked instance of libraft) + a dynamically loaded instance of libraft at the same time? Is there any global state in libraft? Are there any issues sharing data pointers between two instances of libraft (one statically linked, the other dynamically loaded) within the same process?

Issue #3027, opened 2023-08-24
Can HNSWPQ nbits parameter be made configurable?


jmazanec15:
# Summary
For [IndexHNSWPQ](https://github.com/facebookresearch/faiss/blob/v1.7.4/faiss/IndexHNSW.cpp#L879),  I see the value for [`nbits`](https://github.com/facebookresearch/faiss/blob/v1.7.4/faiss/IndexPQ.h#L36) is hardcoded to 8. I wanted to know if this could possibly be made configurable?

I saw this issue from several years ago: #259, where reasoning for not going above 8 nbits on CPU was over concerns of keeping table in l1/l2 cache. However, I am wondering if improved recall may be worth the tradeoff in additional latency. Additionally, for graph-based algorithms, where fewer distance computations are made, I am wondering if the latency impact from not fitting table in l1 cache may not be as pronounced.

mdouze:
Suppport for >8 bits per PQ has improved a lot, so we should support it in HNSW as well. 

Easy to address here with a default value of 8 

https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L889

the only issue is that the PQ parameters are intermixed with the HNSW parameters but there is no easy way around it wihtout breaking compatibiiuty.

jmazanec15:
Sure, let me see if I can make the change without breaking bwc. I assume that we could use default of 8 as a way to avoid this in most cases.

PR #3031, opened 2023-08-31
make nbits configurable for graph indices based on PQ

mdouze:
Summary:
As requested in

https://github.com/facebookresearch/faiss/issues/3027

Indeed, PQ sizes with nbits > 8 are good tradeoffs, so it is interesting to support them.

Differential Revision: D48860659


jmazanec15:
Thanks @mdouze! I can run some tests on this if that would be helpful!

mdouze:
> Thanks @mdouze! I can run some tests on this if that would be helpful!

I am merging this PR. LMK if you have feedback on the functionality (positive or negative).

PR #3034, opened 2023-08-31
Fix is_trained in IndexHNSWSQ

naveentatikonda:
### Description
Even though `is_trained` is set as `true` in [IndexScalarQuantizer](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexScalarQuantizer.cpp#L34). It is again overwritten as `false` in [IndexHNSW](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L910) which is failing at this [validation check](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L363) while ingesting vectors. Raising this PR with a small change to fix it.


mdouze:
Right, is_trained is true for the FP16 encodings that do not require training.

PR #3055, opened 2023-09-12
Small refactoring of inverted lists

alexanderguzhva:


faiss/invlists/OnDiskInvertedLists.cpp
```diff
@@ -524,7 +524,7 @@ void OnDiskInvertedLists::free_slot(size_t offset, size_t capacity) {
         it++;
     }
 
-    size_t inf = 1UL << 60;
+    size_t inf = 1ULL << 60;
```
mdouze, line 527: should be (size_t)1 << 60. UL/ULL is too platform deopendent

PR #3089, opened 2023-10-11
use precomputed norms for raft brute_force knn calls

benfred:


mdouze:
could you give some context?

mdouze:
Would it be possible to support this for incremental adds?

benfred:
> Would it be possible to support this for incremental adds?

@mdouze  This is already supported here in the `RaftFlatIndex` class here with this change.  This class subclasses the `FlatIndex` class - that supports incremental adds, and keeps the precomputed norms for the index up to date when new data is added to the index via the `FlatIndex::add` code.   We're using these precomputed norms here - and this functionality is being tested here https://github.com/facebookresearch/faiss/blob/9bb6b4be0d95042d10beba65706cacc234592128/faiss/gpu/test/test_raft.py#L55-L63

mdouze:
Thanks for the clarification! Importing to internal version...

faiss/gpu/GpuDistance.cu
```diff
@@ -236,89 +236,84 @@ void bfKnn(GpuResourcesProvider* prov, const GpuDistanceParams& args) {
         raft::device_resources& handle = res->getRaftHandleCurrentDevice();
         auto stream = res->getDefaultStreamCurrentDevice();
 
-        idx_t dims = args.dims;
-        idx_t num_vectors = args.numVectors;
-        idx_t num_queries = args.numQueries;
+        int64_t dims = args.dims;
```
algoriddle, line 239: @benfred Can you clarify why you're changing these to int64_t?

Issue #3086, opened 2023-10-10
Documentation errors


Pe4enIks:
# Summary

The documentation contains errors.

The [IndexPQ](https://faiss.ai/cpp_api/struct/structfaiss_1_1IndexPQ.html) constructor has duplicate parameter descriptions, metric parameter description doesn't exist.

```
IndexPQ(int d, size_t M, size_t nbits, MetricType metric = METRIC_L2)
  Constructor.
  
  Parameters:
  - d – dimensionality of the input vectors
  - M – number of subquantizers
  - nbits – number of bit per subvector index
  - d – dimensionality of the input vectors
  - M – number of subquantizers
  - nbits – number of bit per subvector index
```

Should I write about this here, or maybe fill in a form that I haven't found yet, because there are several such bugs.?

mdouze:
Right, this is because the doc is duplicated in Doxygen. 

https://github.com/facebookresearch/faiss/blob/main/faiss/IndexPQ.h#L28

If you want to fix it, please submit a PR. Thanks!

PR #3092, opened 2023-10-12
Documentation fixes

Pe4enIks:
As a follow-up to this issue #3086, I've fixed some bugs in the doxygen-generated documentation.

### DIFF
--- a/faiss/IndexFlatCodes.h
+++ b/faiss/IndexFlatCodes.h
@@ -34,13 +34,6 @@ struct IndexFlatCodes : Index {
 
     void reset() override;
 
-    /** Reconstruct vectors i0 to i0 + ni - 1
-     *
-     * this function may not be defined for some indexes
-     * @param i0          index of the first vector in the sequence
-     * @param ni          number of vectors in the sequence
-     * @param recons      reconstucted vector (size ni * d)
-     */
     void reconstruct_n(idx_t i0, idx_t ni, float* recons) const override;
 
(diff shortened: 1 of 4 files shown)


faiss/IndexFlatCodes.h
```diff
@@ -34,7 +34,13 @@ struct IndexFlatCodes : Index {
 
     void reset() override;
 
-    /// reconstruction using the codec interface
+    /** Reconstruct vectors i0 to i0 + ni - 1
+     *
```
mdouze, line 38: please move this doc to the parent as it applies to all children

faiss/IndexPQ.h
```diff
@@ -26,10 +26,6 @@ struct IndexPQ : IndexFlatCodes {
     ProductQuantizer pq;
 
     /** Constructor.
```
mdouze, line 28: if there is a choice, please priviledge putting the doc in a /** */ in front of the function rather than inline (inline is worse with multi-line doc)

mdouze (commented):
Thanks for the PR! Please consider my comments below.

faiss/IndexFlatCodes.h
Pe4enIks: Got it, it confused me a bit because Index.h already has this doc, so I think I should just remove the inline doc.

mdouze:
Thanks! importing to internal Faiss.

PR #3139, opened 2023-11-22
fix swig ownership when assigning to struct pointer field

mdouze:
Summary: When faiss is called from SWIG, it is usually assumed that Python will do all the memory management, therefore assigning to a pointer in C++ should not transfer ownerhsip (default SWIG behavior). This diff fixed that.

Differential Revision: D51529085


junjieqi:
@mdouze do we still need this one?

mnorris11:
Even though the final assert fails since the ownership disappears, it seems like the memory does not increase. This typemap looks right to me when I test it. We can discuss on next steps.

Issue #3087, opened 2023-10-10
[Feature Request]Parent join support


heemin32:
# Summary
Similar to what the Lucene does in https://github.com/apache/lucene/pull/12434, I would like to see parent join support in Faiss.

For example, search method accepts [bitset](https://github.com/benwtrent/lucene/blob/e14327288e61dacd278b748f332d907758c07b26/lucene/core/src/java/org/apache/lucene/util/BitSet.java#L27) parameter which contains parent id of child in the graph. Then, result is deduped per parent during graph navigation.

mdouze:
That's an interesting database functionality. 
I don't see immediately how this could be supported in Faiss, but let's keep it in mind.

heemin32:
I am thinking similar pattern that how Lucene does. In search method, it take BitSet as parameter. This will enforce a constraint on how doc id should be defined though.

DocId should be ordinal of entire documents starting from zero. Doc Id should be assigned in a way that child doc Id is smaller than parent doc Id. This way, parent doc id can be found quickly with small memory footprint using BitSet. For example, let's say we have two parent docs p1 and p2 with doc id 2 and 4 respectively. Also, we have two child docs of p1 with doc id 0 and 1, one child doc of p2 with doc id 3. Then, BitSet is represented as 00101

navneet1v:
@mdouze I see @heemin32 has updated the proposal, I think this will be a good feature to go in side by side of IDSelector Interface.

This will provide the a better way to get top K similar results. Please provide your thoughts.

heemin32:
@mdouze Any further thought on this? As @navneet1v mentioned, we can add something like IDDeduper(or IDGroup) and pass it through search parameter. During the result collection, we can dedupe the the ID and only store the nearest ID/Distance.

mdouze:
AFAICS this is of broad enough interest to put into main Faiss. 
You have to realize that adding a new result filtering criterion means touching ~20 index implementations, at least to raise an error signaling it is not supported.
Workarounds are: 
- over-fetch and dedup afterwards
- implement search externally.

heemin32:
I think we can support the feature incrementally starting hnsw index. For all other index, we can throw an error message like what we do for search parameter. `FAISS_THROW_IF_NOT_MSG(!params, "search params not supported for this index");`

navneet1v:
@mdouze 
The workarounds provided won't work well. 

[...]

PR #3140, opened 2023-11-22
Add vector deduplication for HNSW

heemin32:
Add vector deduplication for HNSW

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -54,6 +54,7 @@ set(FAISS_SRC
   impl/AuxIndexStructures.cpp
   impl/CodePacker.cpp
   impl/IDSelector.cpp
+  impl/IDDeduper.cpp
   impl/FaissException.cpp
   impl/HNSW.cpp
   impl/NSG.cpp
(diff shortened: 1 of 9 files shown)


mdouze:
Thanks for doing the effort of implementing this deduplication functionality. 
However, as I said in https://github.com/facebookresearch/faiss/issues/3087 I don't think it is of sufficient interest. 
We are trying to limit the amount of code in Faiss because otherwise it is more work to maintain.

heemin32:
@mdouze Thanks for the comment. Could you tell how to decide if there are sufficient interest or not? OpenSearch uses faiss for knn search engine and there are many interest on document deduplication feature. In other words, what would make your mind change to accept this PR?

cjnolet:
I don't want to take any attention away from FAISS here, but multi-valued keys have been requested several times in the RAFT / cuVS library. It's on our roadmap to hopefully implement sometime within the year. If the feature ends up performing well, we would be interested in contributing it to FAISS. At the very least, we could write up an example of how to accomplish this functionality using the Boolean filter function.

PR #3143, opened 2023-11-23
Search and return codes

mdouze:
This PR adds a functionality where an IVF index can be searched and the corresponding codes be returned. It also adds a few functions to compress int arrays into a bit-compact representation.

### DIFF
--- a/benchs/bench_all_ivf/bench_all_ivf.py
+++ b/benchs/bench_all_ivf/bench_all_ivf.py
@@ -7,6 +7,7 @@ import argparse
 import os
 import sys
 import time
+import json
 
 import faiss
 import numpy as np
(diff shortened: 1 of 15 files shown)


alexanderguzhva:
can `search_and_return_codes()` be merged with the regular `search(..., uint8_t* return_codes = nullptr)` by just doing some check that the codes will get populated if `return_codes` is not `nullptr`?

mdouze:
> can `search_and_return_codes()` be merged with the regular `search(..., uint8_t* return_codes = nullptr)` by just doing some check that the codes will get populated if `return_codes` is not `nullptr`?

Could be, but (1) the default IVF search code is already very complicated and (2) we'll have to call it with store_pairs=True anyways because you can't enqueue the codes in the heap.

PR #3146, opened 2023-11-27
Fix potential buffer overflow

baderouaich:
Size of pointer `sub_x` used instead of size of its data. This is likely to lead to a buffer overflow if the user is not lucky enough to be in a x32 bit machine where `sizeof(float*) == sizeof(float)`. 
You probably intend to write `sizeof(*sub_x)` or `sizeof(float)` ?


mdouze:
absolutetly and indeed sizeof(sub_x) = 8 != sizeof(float)

PR #3161, opened 2023-12-07
Introduce result collector for HNSW

heemin32:
Introduce result collector for HNSW

The result collector is to be used by caller to implement their own logic to collect result. One of example is deduplicating result based on group id. https://github.com/facebookresearch/faiss/issues/3087

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -162,6 +162,8 @@ set(FAISS_HEADERS
   impl/ProductQuantizer.h
   impl/Quantizer.h
   impl/ResidualQuantizer.h
+  impl/ResultCollector.h
+  impl/ResultCollectorFactory.h
   impl/ResultHandler.h
   impl/ScalarQuantizer.h
   impl/ThreadedIndex-inl.h
--- a/faiss/Index.h
+++ b/faiss/Index.h
@@ -38,11 +38,12 @@
 
 namespace faiss {
 
-/// Forward declarations see impl/AuxIndexStructures.h, impl/IDSelector.h and
-/// impl/DistanceComputer.h
+/// Forward declarations see impl/AuxIndexStructures.h, impl/IDSelector.h,
+/// impl/DistanceComputer.h, and impl/ResultCollectorFactory.h
 struct IDSelector;
 struct RangeSearchResult;
 struct DistanceComputer;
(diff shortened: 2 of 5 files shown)


mdouze:
Thanks for the PR. 

It would indeed be useful to have a `ResultCollector` object to support knn / range search transparently, and possibly deduplication. 
The result collector would be a class of the form: 

```c++
struct ResultCollector {
    float threshold; // result will be added if distance < threshold 
    virtual bool add_result(idx_t id, float dis) = 0; // returns whether the result was added and the threshold was updated.
};
```

It can be provided to the Index's search function for a given query (like `HNSW::search_from_candidates`). 

The advantage of having an abstract class for `ResultCollector` rather than a template is that it can be provided from an external caller. Besides, since the comparison with the `threshold` (with `C::cmp`) is performed prior to calling the virtual function, there is no possibly slow virtual function call for all computed distances, only for the ones that pass the comparison. 

There are already ResultCollector objects in [IndexIVFPQ](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVFPQ.cpp#L754) and [IndexIVFPQFastScan](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/simd_result_handlers.h). 

The ResultCollector in this PR is of too limited scope to be useful, since it does not enable the range search support and calling externally.

heemin32:
Thanks for the comment. I get the point that it does not support for range search. However, what do you mean by saying "calling externally"?

mdouze:
I don't think the resultcollector could replace search and range_search functions because these take batches of queries while the resultcollector is for lower level calls and not thread-safe. 
It would remain an internal function of the indexes.

heemin32:
1. If ResultCollector is not passed through top level method like search and range_search, it cannot be overridden by another collector. 
2. Because it is not thread-safe, we need factory or something else to create ResultCollector per thread anyway

Another option is passing ResultCollector factory through HNSWSearchParameter as this PR and make its interface more generic so that both knn search and range search can be supported transparently.

mdouze:
Please hold on implementing anything. We are currently discussing how to refactor the code with more `ResultHandler`s.

heemin32:
Good to know that discussion is going on regarding ResultHandler. Hope that new plan also include the deduplication use case.

Issue #3014, opened 2023-08-18
Support for signed int8 and float16 for HNSW and IVF using Flat and PQ encoders


naveentatikonda:
# Summary

Using float32 datatype to ingest vectors using IndexHNSW or IndexIVF (using Flat or PQ encoders) is getting expensive in terms of storage and memory especially for large scale use cases. Adding support for **signed int8 and fp16** vector datatype support helps to reduce these memory footprints. 

As of now, it supports [uint8 for indexing Binary vectors](https://github.com/facebookresearch/faiss/wiki/Binary-indexes) and [Scalar Quantizer has the fp16 support](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexScalarQuantizer.cpp#L34). Do you have any plans to support them for other methods and encoders mentioned above ?

mdouze:
HNSW and IVF both support FP16, SQ8 and PQ compression, see 
https://github.com/facebookresearch/faiss/wiki/The-index-factory

navneet1v:
@naveentatikonda while creating the index you need to send `SQfp16` instead Flat to take use of floats as 16bit.

naveentatikonda:
@mdouze Thank you for your reply. It is actually hard to understand the encoder and decoder logic in this [fp16 quantizer](https://github.com/facebookresearch/faiss/blob/main/faiss/utils/fp16-inl.h). Can you please confirm if we provide a quantized vector as input where each dimension is of type `fp16` and try to ingest them using `HNSW` and  `SQfp16` encoder, will faiss support it ?

If it is supported, will it quantize this vector further and does the vector value change ?

naveentatikonda:
> @mdouze Thank you for your reply. It is actually hard to understand the encoder and decoder logic in this [fp16 quantizer](https://github.com/facebookresearch/faiss/blob/main/faiss/utils/fp16-inl.h). Can you please confirm if we provide a quantized vector as input where each dimension is of type `fp16` and try to ingest them using `HNSW` and `SQfp16` encoder, will faiss support it ?
> 
> If it is supported, will it quantize this vector further and does the vector value change ?

@mdouze Did you get a chance to look into my question?

mdouze:
Sorry I did not understand the question. 
If you use SQfp16, the internal representation will be FP16.
You cannot provide a quantized vector, the index will do it at add time. 
Is that clear?

naveentatikonda:
[...]

PR #3166, opened 2023-12-07
Add SIMD NEON Optimization for QT_FP16 in Scalar Quantizer

naveentatikonda:
### Description
We have SIMD support for x86 architecture using AVX2 optimization. But, we don't have a similar optimization for ARM architecture in Scalar Quantizer for the Quantization Type `QT_FP16`. This PR adds SIMD support for ARM using NEON optimization.

### Issues resolved 
https://github.com/facebookresearch/faiss/issues/3014

### DIFF
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -91,6 +91,20 @@ struct Codec8bit {
         return _mm256_fmadd_ps(f8, one_255, half_one_255);
     }
 #endif
+
+#ifdef __aarch64__
+    static FAISS_ALWAYS_INLINE float32x4x2_t
+    decode_8_components(const uint8_t* code, int i) {
+        float32_t result[8] = {};
+        for (size_t j = 0; j < 8; j++) {
+            result[j] = decode_component(code, i + j);
+        }
+        float32x4_t res1 = vld1q_f32(result);
+        float32x4_t res2 = vld1q_f32(result + 4);
+        float32x4x2_t res = vzipq_f32(res1, res2);
+        return vuzpq_f32(res.val[0], res.val[1]);
+    }
+#endif
 };
 
 struct Codec4bit {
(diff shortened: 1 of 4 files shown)


faiss/impl/ScalarQuantizer.cpp
```diff
@@ -728,6 +745,59 @@ struct SimilarityL2<8> {
...
+    }
+
+    FAISS_ALWAYS_INLINE void add_8_components(float32x4x2_t x) {
+        float32x4x2_t yiv = vld1q_f32_x2(yi);
+        yi += 8;
+
+        float32x4_t sub0 = vsubq_f32(yiv.val[0], x.val[0]);
+        float32x4_t sub1 = vsubq_f32(yiv.val[1], x.val[1]);
+
+        float32x4_t accu8_0 = vaddq_f32(accu8.val[0], vmulq_f32(sub0, sub0));
```
alexanderguzhva, line 770: `vfmaq_f32()`

faiss/impl/ScalarQuantizer.cpp
```diff
@@ -1024,55 +1195,57 @@ struct DistanceComputerByte<Similarity, 8> : SQDistanceComputer {
  * specialization
  *******************************************************************/
 
-template <class Sim>
+template <class Sim, class Sim_default>
```
alexanderguzhva, line 1198: I don't understand why it is crucial to introduce `class Sim_default` here. @naveentatikonda, could you please give some comments? Thanks.

alexanderguzhva (changes requested):

faiss/impl/ScalarQuantizer.cpp
naveentatikonda: In ARM architecture, as we don't have the templates implemented for all the Quantizers(except QT_fp16) with SIMDWIDTH 8 it is throwing errors when we were trying to build it as it is not able to find the corresponding templates . So, I introduced a new argument `Sim_default` without messing up this switch case or adding more if else conditions. Pls let me know if there is any better way to do this.

faiss/impl/ScalarQuantizer.cpp
naveentatikonda: ack

faiss/impl/ScalarQuantizer.cpp
mdouze: Agreed with @alexanderguzhva. 
Let's try to not make ScalarQuantizer even more complex than it already is.

naveentatikonda: Thank you for the suggestions. Will implement the 2nd approach suggested by @alexanderguzhva

alexanderguzhva:
@naveentatikonda Windows code does not seem to run properly. Also, there are still two Similarity typenames in a template.

naveentatikonda:
I also ran benchmarking tests for all other SQ quantizers after making the changes and the recall on ARM architecture is matching with the corresponding recall on x86 architecture.

naveentatikonda:
@mdouze @alexanderguzhva All the checks have passed. Can you please review this PR

mdouze:
LGTM. 
Importing to internal version. Let's see if @alexanderguzhva  has more comments.

alexanderguzhva:
@naveentatikonda @mdouze lgtm

naveentatikonda:
@mdouze @alexanderguzhva Thank you for reviewing the PR.
@mdouze Can you please merge these changes ASAP

naveentatikonda:
@mdouze Did you get a chance to review and merge the changes in this PR ? Can you please take a look

mdouze:
Sorry a glitch with the import, redoing...

PR #3190, opened 2024-01-04
Generalize ResultHanlder, support range search for HNSW and Fast Scan

mdouze:
Summary:
This diff adds more result handlers in order to expose them externally.
This enables range search for HSNW and Fast Scan, and nprobe parameter support for FastScan.

Differential Revision: D52547384


mdouze:
Sorry I was not aware of this... 
Is there a changelog for knowwhere's modifications to Faiss? 
Maybe we should sync from time to time with common meetings.

alexanderguzhva:
I think that Faiss should remain as the baseline and I need to actively merge changes into it
Let me create a list of differences between the baseline and the patched version from Zilliz

PR #3233, opened 2024-02-01
HNSW search use efSearch from params if provided

xinhuitian:
I found that HNSW search method uses its own efSearch to create candidates, which should be got from params if provided I think?


mdouze:
Yes that makes sense, I don't understand how we missed this. 
Thanks

PR #3238, opened 2024-02-02
Fix AVX2 build on Windows

### LINKED ISSUE #3193: AVX2 support doesn't compile with MSVC
# Summary

The AVX2 implementation of FAISS appears to use intrinsics that are GNU specific (or at least don't port to Windows via MSVC).

# Platform

Windows 10.0.19045 
MSVC 2022

Faiss version: git tag v1.7.4

Installed from: Attempting to compile

Faiss compilation options:  

Probably doesn't much matter but using OpenBLAS compiled myself
CMake `set(FAISS_OPT_LEVEL "avx2" CACHE INTERNAL "")` and link with faiss_avx2 instead of faiss

Running on:
- [X] CPU
- [ ] GPU

Interface: 
- [X] C++
- [ ] Python

# Reproduction instructions

1. Get any BLAS implementation in order to satisfy the dependency requirements for building this library
2. Standard CMake dance (build directory, etc)
3. `cmake -DFAISS_OPT_LEVEL=avx2 -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=OFF ..` (assuming the source directory is one level up from the build directory)
4. `cmake --build . --config Release --target faiss_avx2 ..`

# Fix instructions

This boils down to 3 distinct issues:

1. The definition of `_mm_prefetch` in MSVC (and actually in the [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_prefetch) suggesting GCC / clang are mistaken) takes a `const char *` as the first argument, not `const void *`.  This base is peppered with calls directly passing float pointers.
2. `__m128i_u` and `__m256i_u` are not portable.  However they are the exact same datatype as the non `_u` variants, just with some [implied side effects](https://stackoverflow.com/a/68939636/1155387) so I'm guessing it is safe to define them away to their non `_u` variants `#if defined(_MSC_VER) && !defined(__clang__)`.
3. By default Visual Studio implemented OpenMP 2.0, which doesn't allow signed indexes in for loops.  The following code violates this https://github.com/facebookresearch/faiss/blob/d87888b13e7eb339bb9c45825e9d20def6665171/faiss/utils/distances_fused/simdlib_based.cpp#L262-L266  

However, issue 3 can be worked around with some trickery, involving override CMake's FindMP module by setting `OpenMP_CXX_FLAGS="-openmp:llvm"`.  This turns on the "better" support (which is OpenMP 2.5 with a few 3.1 features including unsigned index types) for OpenMP in Visual Studio as noted [here](https://devblogs.microsoft.com/cppblog/openmp-task-support-for-c-in-visual-studio/)

borrrden:
Tested with both MSVC (with /openmp:llvm) and clang-cl (no particular extra flags needed).  This PR is separated into two commits (three after I found out that lines need to be 80 chars or less):

1. Changes needed for clang-cl (and probably stock clang too)
2. Changes needed for MSVC

So FAISS can decide either to require using LLVM for Windows (not a hard thing to do these days since it is fully supported inside Visual Studio) and discarding the second commits, or taking them all and documenting the need to use /openmp:llvm

Closes #3193

### DIFF
--- a/faiss/impl/LocalSearchQuantizer.cpp
+++ b/faiss/impl/LocalSearchQuantizer.cpp
@@ -628,7 +628,9 @@ void LocalSearchQuantizer::icm_encode_step(
                         {
                             size_t binary_idx = (other_m + 1) * M * K * K +
                                     m * K * K + code2 * K + code;
-                            _mm_prefetch(binaries + binary_idx, _MM_HINT_T0);
+                            _mm_prefetch(
+                                    (const char*)(binaries + binary_idx),
+                                    _MM_HINT_T0);
                         }
                     }
 #endif
--- a/faiss/impl/platform_macros.h
+++ b/faiss/impl/platform_macros.h
@@ -40,11 +40,13 @@
 
 #include <intrin.h>
 
+#ifndef __clang__
 inline int __builtin_ctzll(uint64_t x) {
     unsigned long ret;
     _BitScanForward64(&ret, x);
     return (int)ret;
 }
+#endif
 
 // cudatoolkit provides __builtin_ctz for NVCC >= 11.0
 #if !defined(__CUDACC__) || __CUDACC_VER_MAJOR__ < 11
@@ -55,9 +57,11 @@ inline int __builtin_ctz(unsigned long x) {
 }
 #endif
 
+#ifndef __clang__
 inline int __builtin_clzll(uint64_t x) {
     return (int)__lzcnt64(x);
 }
+#endif
 
 #define __builtin_popcount __popcnt
 #define __builtin_popcountl __popcnt64
--- a/faiss/impl/platform_macros.h
+++ b/faiss/impl/platform_macros.h
@@ -66,6 +66,11 @@ inline int __builtin_clzll(uint64_t x) {
 #define __builtin_popcount __popcnt
 #define __builtin_popcountl __popcnt64
 
+#ifndef __clang__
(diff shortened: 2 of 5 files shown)


borrrden:
I actually don't know what the format check wants from me...if somebody could explain then I'll take whatever action necessary.

borrrden:
The format check forced me to do something that broke the actual build?  I don't really get it....I'm going to put that line in question back to one line even though it violates the 80 char rule

borrrden:
Nevermind sorry about that.  I copied and pasted from the CircleCI output to try to get exactly what it was telling me to do and ended up copying and pasting the diff plus signs.

borrrden:
One more note is that I noticed some warnings when using -Wall about used reserved defines.  This is for the popcnt stuff, and the __builtin defines which clang actually has.  The build is ok without doing anything here so I assume if these implementations are sane then it won't be a big deal, but it's another area for possible improvement for clang on Windows.

borrrden:
What is the story here?  This PR keeps getting out of date.  Should I keep rebasing it or wait until someone notices that it is here?

mdouze:
Don't worry ;-) 
Once the PR is imported, we take it over and it will be integrated into Faiss.

borrrden:
Ok, I started getting that impression but thanks for the confirmation!

Issue #3246, opened 2024-02-08
Disable compute_sdc_table if HNSWPQ index is read only


jmazanec15:
# Summary

In [index_read](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L966) for HNSWPQ, the `sdc_table` is always computed. This SDC table is used to optimize distance computations during graph construction in HNSW through symmetric_distance operation:
1.  https://github.com/facebookresearch/faiss/blob/main/faiss/IndexPQ.cpp#L95-L106
2. https://github.com/facebookresearch/faiss/blob/main/faiss/impl/HNSW.cpp#L316
3. https://github.com/facebookresearch/faiss/blob/main/faiss/impl/HNSW.cpp#L240

This table comes with the cost of `M*2^nbits*2^nbits*4` bytes of memory. For systems that will use this index type as a template multiple times (this is how we integrated with OpenSearch), this can accumulate a large memory overhead: `500*(64*2^8*2^8*4)/1024^3 = 7.8125 GB`.

This SDC table is not used during `search`. That being said, I am wondering if we can change the read logic to avoid computing this table  if the index is read only.

So, my proposal is to switch [index_read](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L966) to:
```
if (h == fourcc("IHNp") && !(io_flags & IO_FLAG_READ_ONLY)) {
   dynamic_cast<IndexPQ*>(idxhnsw->storage)->pq.compute_sdc_table();
}
```

From my understanding, this should not break any of the typical functionality of the HNSWPQ index type when it is read only. The only case the could be potentially problematic is when a user is reading the index from disk and then extracting PQ from the index and using it for other things (like maybe building another HNSWPQ index). That being said, I am not sure if this is a typical/valid use case.

Alternative options I can think of:
1. Add different io flag to specifically skip this pre-computed table - in general, this table is implementation detail so I think itd be better to avoid exposing this in the interface if possible
2. Implement logic on our side to read index and manually skip this - this may work but would be more difficult to maintain

I can raise a PR to make change, but wanted to first get feedback on the approach.

Running on:
- [X] CPU
- [ ] GPU

Interface: 
- [X] C++
- [ ] Python

mdouze:
Thanks for this analysis. This sounds good, IMO the flag could be `IO_FLAG_SKIP_PRECOMPUTE_TABLE`

[...]

PR #3250, opened 2024-02-13
Skip HNSWPQ sdc init with readonly flag

jmazanec15:
## Description

Related issue: #3246 

When reading HNSWPQ from disk, if index read only flag is set, skip initializing the sdc_table. This is okay because the sdc computations are only used during graph construction, which will not happen when index is readonly.

In addition, adds cpp test case verifying functionality and build test util header file to share creation of temporary files amongst tests.


jmazanec15:
Test failure seems unrelated and was unable to repro locally:
```
The following tests FAILED:
        121 - TestFastScan.knnVSrange (Failed)
Errors while running CTest
```

mdouze:
I would rather introduce a new flag then, `IO_FLAG_PQ_SKIP_SDC_TABLE` for example.
The fact that readonly means that SDC is not needed in HSNW + PQ is very non obvious.

mdouze:
Could you check if all times the SDC tables are used, a exception is raised if the table is not initialized?

jmazanec15:
Sure, 

`pq->sdc_compute_table()` produces `pq->sdc_table` and is called in:
1. [IndexHNSW::train](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L684) - use through PQDistanceComputer
2. [IndexNSG::train](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexNSG.cpp#L299) - use through PQDistanceComputer
3. [PolysemousTraining::optimize_ranking](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/PolysemousTraining.cpp#L845) - used internally
4. [PolysemousTraining::optimize_pq_for_hamming](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/PolysemousTraining.cpp#L845) - used internally
4. [index_read::read_index](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L966)

`pq->sdc_table` is used in:
1. [IndexPQ::PQDistanceComputer](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexPQ.cpp#L118-L119) - if size == 0, sdc is set to null. symmetric_dis() then will properly throw. 
1. [PolysemousTraining::optimize_ranking](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/PolysemousTraining.cpp#L882) - `pq.compute_sdc_table` called earlier in method, so it will not be null ([ref](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/PolysemousTraining.cpp#L845C9-L845C31))
1. [ProductQuantizer::search_sdc](https://github.com/facebookresearch/faiss/blob/main/faiss/impl/ProductQuantizer.cpp#L818) - handled explictly

mdouze:
Thanks, importing the PR...

jmazanec15:
@mdouze I see there was a failure internally. Is this related to this change or is it problem elsewhere? I dont think I can access logs.

mdouze:
Sorry the process was stuck internally, aiming to merge today.

mdouze:
Merge done

PR #3304, opened 2024-03-21
Improve filtering & search parameters propagation

alexanderguzhva:


faiss/impl/simd_result_handlers.h
```diff
@@ -217,6 +221,20 @@ struct ResultHandlerCompare : SIMDResultHandlerToFloat {
         return lt_mask;
     }
 
+    uint32_t get_lt_mask_for_range_search(size_t b) {
```
mdouze, line 224: Maybe this mask could take into account the sel, then we don't need separate conditions for with / without sel.

mdouze (commented):
Thanks for undertaking this.
Could you consider my suggestion to avoid too much code duplication? Thanks

faiss/impl/simd_result_handlers.h
alexanderguzhva: A good suggestion. Let me see...

alexanderguzhva:
@mdouze I've removed `uint32_t get_lt_mask_for_range_search(size_t b)` that you've mentioned, it is a change for the next PR, which will be related to range search.

Also, I think that the following code constructions
```
        if (this->sel != nullptr) {
            while (lt_mask) {
                // find first non-zero
                int j = __builtin_ctz(lt_mask);
                auto real_idx = this->adjust_id(b, j);
                lt_mask -= 1 << j;
                if (this->sel->is_member(real_idx)) {
                    T dis = d32tab[j];
                    if (C::cmp(heap_dis[0], dis)) {
                        heap_replace_top<C>(k, heap_dis, heap_ids, dis, real_idx);
                    }
                }
            }
        } else {
            while (lt_mask) {
                // find first non-zero
                int j = __builtin_ctz(lt_mask);
                lt_mask -= 1 << j;
                T dis = d32tab[j];
                if (C::cmp(heap_dis[0], dis)) {
                    int64_t idx = this->adjust_id(b, j);
                    heap_replace_top<C>(k, heap_dis, heap_ids, dis, idx);
                }
            }
        }
```
is the compromise. I don't want to collapse this `if-else statement`, because it provides the optimal performance behavior for both branches. Creating the code, such as `if (sel == nullptr || sel->is_member())` will lead to suboptimal performance, because `sel->is_member()` has to evaluate `this->adjust_id(b, j);` all the time, despite `C::cmp()` results

alexanderguzhva:
rebased to master

mdouze:
I will add a python test to this unless you have one ready.

Issue #3258, opened 2024-02-20
sa_decode error on IVFPQFastScan


mdouze:
# Summary

The `sa_decode` function does not work for IVFPQFastScan.

Faiss version: 1.7.4

Installed from: anaconda

Faiss compilation options: installed from package

Running on:
- [ x] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [ x] Python

# Reproduction instructions

see https://gist.github.com/mdouze/219a9b84c00fb75e98ccbc0fb56ba6e8
compares with IVFPQ decode version. 
Besides, the sa_decode does not work for flat PQ index

PR #3312, opened 2024-03-22
Fix IVFPQFastScan decode function

junjieqi:
Summary:
as the [#issue3258](https://github.com/facebookresearch/faiss/issues/3258) mentioned, the IVFPQFastScan should have same decoding result as IVFPQ. However, current result is not as expected.

In this PR/Diff, we are going to fix the decoding function

Differential Revision: D55264781

### DIFF
--- a/faiss/IndexIVFPQFastScan.cpp
+++ b/faiss/IndexIVFPQFastScan.cpp
@@ -286,9 +286,28 @@ void IndexIVFPQFastScan::compute_LUT(
     }
 }
 
-void IndexIVFPQFastScan::sa_decode(idx_t n, const uint8_t* bytes, float* x)
+void IndexIVFPQFastScan::sa_decode(idx_t n, const uint8_t* codes, float* x)
         const {
-    pq.decode(bytes, x, n);
+    size_t coarse_size = coarse_code_size();
+
+#pragma omp parallel
+    {
+        std::vector<float> residual(d);
+
+#pragma omp for
+        for (idx_t i = 0; i < n; i++) {
+            const uint8_t* code = codes + i * (code_size + coarse_size);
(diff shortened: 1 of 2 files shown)


faiss/IndexIVFPQFastScan.cpp
```diff
@@ -286,9 +286,28 @@ void IndexIVFPQFastScan::compute_LUT(
     }
 }
 
-void IndexIVFPQFastScan::sa_decode(idx_t n, const uint8_t* bytes, float* x)
+void IndexIVFPQFastScan::sa_decode(idx_t n, const uint8_t* codes, float* x)
         const {
-    pq.decode(bytes, x, n);
+    size_t coarse_size = coarse_code_size();
+
+#pragma omp parallel
```
alexanderguzhva, line 293: `parallel if (n > 1)`
or, even better, is n is greater than a certain threshold
please check other `sa_decode` implementations for the appropriate threshold

junjieqi: Hi @alexanderguzhva, thanks for the suggestion. I checked other sa_decode. They also don't have the condition. Try to understand it more. `#pragma omp parallel if (n > 1)` means we only parallel execution when n is greater than 1. I'm wondering how we should choose the threshold in general. Do we have any benchmark to run or other methods?

alexanderguzhva: no benchmarks, this is a magic constant. Typically, things like 100 or 1000 are used.
https://github.com/facebookresearch/faiss/blob/0c96b0d7e0e399458e7fb0703015f03dbecb614d/faiss/IndexScalarQuantizer.cpp#L186

PR #3326, opened 2024-03-28
Fix missing overload variable in Rocksdb ivf demo

Warmchay:
**Bugs:** 
When following rocksdb_ivf demo to build executable file, its output as:
```bash
faiss/demos/rocksdb_ivf/RocksDBInvertedLists.h:52:35: error: 'faiss::InvertedListsIterator* faiss_rocksdb::RocksDBInvertedLists::get_iterator(size_t) const' marked 'override', but does not override
   52 |     faiss::InvertedListsIterator* get_iterator(size_t list_no) const override;
      |                                   ^~~~~~~~~~~~
make[2]: *** [CMakeFiles/demo_rocksdb_ivf.dir/build.make:90: CMakeFiles/demo_rocksdb_ivf.dir/RocksDBInvertedLists.cpp.o] Error 1
```

**Solution:**
Add relevant variable `void* inverted_list_contex` corresponding `get_iterator`'s base virtual function.

### DIFF
--- a/demos/rocksdb_ivf/RocksDBInvertedLists.cpp
+++ b/demos/rocksdb_ivf/RocksDBInvertedLists.cpp
@@ -101,7 +101,7 @@ void RocksDBInvertedLists::resize(size_t /*list_no*/, size_t /*new_size*/) {
 }
 
 InvertedListsIterator* RocksDBInvertedLists::get_iterator(
-        size_t list_no) const {
+        size_t list_no, void* inverted_list_contex) const {
     return new RocksDBInvertedListsIterator(db_.get(), list_no, code_size);
 }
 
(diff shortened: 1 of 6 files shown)


mdouze:
@algoriddle would you mind taking a look?

mdouze:
It would be useful to compiling this in CI so that we catch these kinds of API pdates.

demos/rocksdb_ivf/RocksDBInvertedLists.cpp
```diff
@@ -101,7 +101,8 @@ void RocksDBInvertedLists::resize(size_t /*list_no*/, size_t /*new_size*/) {
 }
 
 InvertedListsIterator* RocksDBInvertedLists::get_iterator(
-        size_t list_no) const {
+        size_t list_no,
+        void* inverted_list_contex) const {
```
algoriddle, line 105: thank you for this PR, can you correct the typo please (also below)? contex -> context

Warmchay: No problem~

PR #3348, opened 2024-04-04
Unroll loop in lookup_2_lanes

carll99:
The current loop goes from 0 to 31.  It has an if statement to do an assignment for j < 16 and a different assignment for j >= 16.  By unrolling the loop to do the j < 16 and the j >= 16 iterations in parallel the if j < 16 is eliminated and the number of loop iterations is reduced in half.

This change results in approximately a 14% reduction in the execution time for the bench_ivf_fastscan.py workload on Power 10 when compiled with CMAKE_INSTALL_CONFIG_NAME=Release.

### DIFF
--- a/faiss/utils/simdlib_emulated.h
+++ b/faiss/utils/simdlib_emulated.h
@@ -532,16 +532,20 @@ struct simd32uint8 : simd256bit {
     // The very important operation that everything relies on
     simd32uint8 lookup_2_lanes(const simd32uint8& idx) const {
         simd32uint8 c;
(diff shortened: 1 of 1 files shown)


carll99:
Looks like the patch fails the formatting check for clang.  Looks like I need spaces in the index calculation.

I am new to using the github.com interface.  I just use the git interface from the command line and send patches to the GDB and GCC mailing lists.  I fixed my patch locally, not sure how I now "fix" my pull request?  Just delete the request and start over or is there an easier way?  Sorry, this is all a learning experience for me.

PR #3365, opened 2024-04-12
Remove unused variables in faiss/IndexIVF.cpp

r-barnes:
Summary:
LLVM-15 has a warning `-Wunused-but-set-variable` which we treat as an error because it's so often diagnostic of a code issue. Unused variables can compromise readability or, worse, performance.

This diff either (a) removes an unused variable and, possibly, it's associated code, or (b) qualifies the variable with `[[maybe_unused]]`, mostly in cases where the variable _is_ used, but, eg, in an `assert` statement that isn't present in production code.

 - If you approve of this diff, please use the "Accept & Ship" button :-)

Reviewed By: dmm-fb

Differential Revision: D56065763

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,12 +38,17 @@ jobs:
       - run:
           name: Install clang-format
           command: |
-            apt-get update
-            apt-get install -y git-core clang-format-11
+            apt-get update -y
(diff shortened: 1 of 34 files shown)


junjieqi:
The code has been merged and close this PR

PR #3384, opened 2024-04-24
Fix IndexBinary.assign Python method

### LINKED ISSUE #3343: Error using assign method with IndexBinary in FAISS
### Description:
I am encountering an issue when attempting to use the `assign` method on an `IndexBinary` object in FAISS. The objective is to assign each input vector to the label of its nearest neighbor within the index. However, when I call the `assign` method, I receive a `TypeError` indicating a mismatch in the number or type of arguments.

### Steps to Reproduce:
1. Import FAISS and other necessary libraries.
2. Generate random floating-point vectors and quantize them using `sentence_transformers.quantization.quantize_embeddings`.
3. Create an `IndexBinary` object and set its parameters.
4. Attempt to train the index with quantized vectors.
5. Initialize an empty numpy array for nearest neighbor labels.
6. Call the `assign` method with the quantized vectors and the labels array.

### Expected Behavior:
The `assign` method should correctly assign each input vector to the label of its nearest neighbor within the index without any errors.

### Actual Behavior:
When calling the `assign` method, the following `TypeError` is thrown:
```
TypeError: Wrong number or type of arguments for overloaded function 'IndexBinary_assign'.
Possible C/C++ prototypes are:
    faiss::IndexBinary::assign(faiss::idx_t,uint8_t const *,faiss::idx_t *,faiss::idx_t) const
    faiss::IndexBinary::assign(faiss::idx_t,uint8_t const *,faiss::idx_t *) const
```

### Environment:
- FAISS version: 1.8.0
- Python version: 3.11.6

### Additional Context:
I am using FAISS for vector similarity search in a binary space. The vectors are first quantized to a binary format before being used with the `IndexBinary` object. Below is a snippet of the code that leads to the error:

```python
import faiss
from sentence_transformers.quantization import quantize_embeddings
import numpy as np

d = 1024                          
nb = 100000  
k=1

xb = np.random.random((nb, d)).astype('float32')
vectors = quantize_embeddings(xb, "ubinary")
count = len(vectors)

index = faiss.index_binary_factory(d, "BIVF64")
index.nprobe = 4
index.train(vectors)

list_nos = np.empty((count, k), dtype=np.int64)

index.assign(n=count, x=vectors, labels=list_nos, k=k)
```

I have ensured that the quantized vectors are of type `uint8` as required. Any insights into what might be causing this error or how to resolve it would be greatly appreciated.

asadoughi:
Differential Revision: D56526842

Fixes #3343

### DIFF
--- a/faiss/python/class_wrappers.py
+++ b/faiss/python/class_wrappers.py
@@ -956,10 +956,44 @@ def handle_IndexBinary(the_class):
             sel = IDSelectorBatch(x.size, swig_ptr(x))
         return self.remove_ids_c(sel)
 
+    def replacement_assign(self, x, k, labels=None):
+        """Find the k nearest neighbors of the set of vectors x in the index.
+        This is the same as the `search` method, but discards the distances.
+
+        Parameters
+        ----------
+        x : array_like
+            Query vectors, shape (n, d) where d is appropriate for the index.
+            `dtype` must be uint8.
+        k : int
+            Number of nearest neighbors.
+        labels : array_like, optional
+            Labels array to store the results.
+
+        Returns
+        -------
+        labels: array_like
+            Labels of the nearest neighbors, shape (n, k).
+            When not enough results are found, the label is set to -1
+        """
+        n, d = x.shape
+        x = _check_dtype_uint8(x)
+        assert d == self.code_size
+        assert k > 0
+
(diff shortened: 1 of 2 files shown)


asadoughi:
Closed via 03750f5

Issue #3351, opened 2024-04-05
Implement timeout for slow functions 


mdouze:
In Faiss, slow functions periodically call `InterruptCallback::check()` 

https://github.com/facebookresearch/faiss/blob/main/faiss/impl/AuxIndexStructures.h#L152

The callback is arbitrary. It does nothing by default. 
In python there is an InterruptCallback that checks if the user pressed Ctrl-C. 

It would be useful to implement an InterruptCallback that breaks out of a function that lasts too long, eg. a clustering.

mdouze:
1. C++ only first 
2. support timeout from python

mdouze:
For the C++ version it can be done with just one additional test. 
It requires subclassing InterruptCallback to implement `want_interrupt` to return true if more than x seconds elapsed since the counter was reset. 

Something like: 

```c++
struct TimeoutCallback: InterruptCallback {
    double t0 = 0; 
    double timeout = 0; 
    void set_timeout(double t) {
        t0 = get_time(); 
        timeout = t; 
    } 
   
   bool want_interrupt () override {
      if (timeout == 0) {
         return false; 
      }
      if (get_time() - t0 > timeout) {
         timeout = 0;  
         return true; 
      } 
     return false;
   }

};
``` 

can be tested by running a very long k-means clustering (see `Clustering` object).

keeth:
looks like this is a one-per-process global timeout, not compatible with multiple threads calling into FAISS, is that right?

PR #3397, opened 2024-04-29
TimeoutCallback

asadoughi:
Summary: https://github.com/facebookresearch/faiss/issues/3351

Differential Revision: D56732720

### DIFF
--- a/faiss/gpu/perf/PerfClustering.cpp
+++ b/faiss/gpu/perf/PerfClustering.cpp
@@ -17,6 +17,7 @@
 #include <vector>
 
 #include <cuda_profiler_api.h>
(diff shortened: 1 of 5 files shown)


faiss/gpu/perf/PerfClustering.cpp
```diff
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <faiss/MetricType.h>
```
mdouze, line 8: this was probably added randomly by vscode

mdouze (commented):
Maybe make a test out of it, that does not depend on GPU.

asadoughi:
Replaced by #3413 as the bot decided to not update this diff with the second commit

PR #3416, opened 2024-05-07
PowerPC, improve code generation for function fvec_L2sqr

carll99:
The code generated for function fvec_L2sqr generated by OpenXL do not perform as good as the codes generated by gcc on Power. The macros to enable imprecise floating point operation don’t cover Power with OpenXL. This patch adds the OpenXL compiler options for the PowerPC macros to achieve better performance.


alexanderguzhva:
@carll99 does it improve the code for dot product as well, did you have a chance to take a look? :)

carll99:
We took a look at the dot product and the performance appears to be the same for the dot product.

mdouze:
lgtm

PR #3430, opened 2024-05-13
[ARM NEON] Get rid of redundant instructions in ScalarQuantizer

alexanderguzhva:
This PR removes unneeded ARM NEON SIMD instructions for ScalarQuantizer.

The removed instructions are completely redundant, and I believe that it is a funky way of converting two `float32x4_t` variables (which hold 4 float values in a single SIMD register) into a single `float32x4x2_t` variable (two SIMD registers packed together). 

Clang compiler is capable of eliminating these instructions. The only GCC that can eliminate these unneeded instructions is GCC 14, which was released very recently (Apr-May 2024).

@mdouze


alexanderguzhva:
rebased to master

mdouze:
So I trust you that the new syntax will compile on all GCCs and clang.

PR #3433, opened 2024-05-14
Missed printing 'D'

saarthdeshpande:
'I' was printed twice and 'D' (distance vector) was not printed. Fixed.

### DIFF
--- a/tutorial/cpp/6-HNSW.cpp
+++ b/tutorial/cpp/6-HNSW.cpp
@@ -57,10 +57,10 @@ int main() {
 
         index.search(nq, xq, k, D, I);
 
-        printf("I=\n");
+        printf("D=\n");
         for (int i = nq - 5; i < nq; i++) {
             for (int j = 0; j < k; j++)
-                printf("%5zd ", I[i * k + j]);
+                printf("%5zd ", D[i * k + j]);
             printf("\n");
         }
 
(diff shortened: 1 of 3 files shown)


tutorial/cpp/6-HNSW.cpp
```diff
@@ -55,12 +55,10 @@ int main() {
...
         }
 
-        index.search(nq, xq, k, D, I);
-
-        printf("I=\n");
+        printf("D=\n");
         for (int i = nq - 5; i < nq; i++) {
             for (int j = 0; j < k; j++)
-                printf("%5zd ", I[i * k + j]);
+                printf("%5zd ", D[i * k + j]);
```
junjieqi, line 61: Format specifies type 'ssize_t' (aka 'long') but the argument has type 'float' [-Werror,-Wformat]
                printf("%5zd ", D[i * k + j]);
                        ~~~~    ^~~~~~~~~~~~
                        %5f

junjieqi: @saarthdeshpande would you like to fix this part?

saarthdeshpande: Sure

saarthdeshpande: Done! Added the changes for other tutorial files too.

PR #3445, opened 2024-05-15
Delete unnecessary print(f) functions in Tests

fxdawnn:
Summary: This commit focus on reducing the length of the unnecessary log and strengthen the codebase.

Differential Revision: D57405802


alexanderguzhva:
is there a guarantee that removing ALL prints help to fix problems identified by unit tests?

Issue #3231, opened 2024-01-31
[RFC] HIP backend for AMD GPU support


iotamudelta:
We would like to contribute a HIP backend to Faiss to support AMD GPUs. We have a working prototype that passes all unit tests on Navi hardware (6800XT, 7900XTX). The prototype features a statically hipified version of the existing CUDA backend with manual AMD specific changes (build system, PTX to amdgcn builtins, ...).

Assuming you are interested in this work, how would we go about upstreaming it best?

Would a static HIP backend (in the faiss::hip namespace) be preferred? If not, what architecture would preferable (e.g., overriding faiss::gpu)?

Unlike the CUDA backend, we ultimately need to support multiple "warp sizes" (wave fronts) at runtime - 32 for Navi and 64 for MI series. There are some questions pertaining to uses of kWarpSize that will not work out of the box (sizing shared memory, some of the replacements in cmake, static uses at dispatch sites, ...). We will need guidance on how such support would be architected and integrated best.

Lastly, we have done only minor performance analysis and/or tuning with the current prototype - are there any public benchmarks and/or particular protocols you have used for the other backends that we should use as a reference? We have used the GPU benchmark scripts with the SIFT data sets to assess performance so far.

As part of this, we're currently using a GPU_MAX_SELECTION_K of 1024 - what would a recommend protocol look like to decide for 1024 vs 2048 - we'd ideally like to use only one independent of HW/SW generation.

To browse the current prototype as a reference, please see https://github.com/iotamudelta/faiss/tree/wf32 .

mdouze:
FYI there is an open (but inactive) pull request for AMD: 
https://github.com/facebookresearch/faiss/pull/3126
Let's discuss how we can coordinate this effort.

iotamudelta:
@mdouze Thanks for the pointer! I see your concerns in that PR, we'll talk internally how to address.

iotamudelta:
@mdouze We had some further discussions internally. @ItsPitt and I can maintain an AMD/HIP backend for now. AMD can contribute two servers for CI - same setup as the PyTorch CI.

I'd recommend closing #3126 as overcome by events. With that in mind, what's the preferred strategy to make this all happen - I assume we'll want multiple steps.

Current state of AMD/HIP is in https://github.com/iotamudelta/faiss/tree/rocm_support

iotamudelta:
Solved - ROCm support is integrated into tip of tree.

PR #3462, opened 2024-05-20
ROCm support

iotamudelta:
* add hipify at configure time
* ROCm specific code paths behind USE_ROCM guards
* support for wavefront 32 (Navi) and 64 (MI)
* use builtins to match inline PTX
* support C API on ROCm
* support Python API on ROCm

---------


iotamudelta:
This superseedes #3126 and addresses #3231

As discussed in #3231 , @ItsPitt (Michael Pittard) and @iotamudelta (myself) can maintain the ROCm backend. AMD can provide two MI200-class servers for CI with a similar arrangement to the PyTorch CI.

mdouze:
Assuming this is WIP because the Python tests are not there yet. Otherwise I can write them but could take some time. 
Sorry for insisting but many people (especially in the ML community) know Faiss only via Python.

iotamudelta:
@mdouze thanks a lot for having a look! yes, we can figure out the Python support prior to getting it merged.

Are there any other immediate issues you see that we should be working on? Do you prefer getting this all merged as one big PR or split up (if so: how)?

Again - thanks a lot!

mdouze:
We are looking into compiling this in the CI
@ramilbakhshyiev

iotamudelta:
@mdouze we sorted the Python support out. I am not sure if the Windows CI target failing after my latest merge of `main` is real or not.

@ramilbakhshyiev anything we can help with w.r.t. getting this into CI? Thanks!

ramilbakhshyiev:
@iotamudelta We will be trying this soon, our plan is to use a g4ad instance from AWS. Once we try it on that, we will most likely come back with some feedback.

iotamudelta:
@ramilbakhshyiev I'm not sure g4ad will work - according to https://aws.amazon.com/ec2/instance-types/ , it features a AMD Radeon Pro V520 w/ RDNA1 architecture which is not supported by ROCm (see: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html )

ramilbakhshyiev:
@iotamudelta Is there any plan to support it? I think AWS is a widely used platform and it only seems to support g4ad, right?

iotamudelta:
@ramilbakhshyiev I do not know about support plans. ROCm does support RDNA2 and RDNA3 GPUs.

ramilbakhshyiev:
Based on some of my very high-level research, it seems that some users still get ROCm on g4ad even though it's unsupported by AMD. Do you have any opinion on using that to build this in our CI? If we don't have an option to build and test on g4ad then I believe that our only option is to use CI only for builds.

iotamudelta:
@ramilbakhshyiev it's hard to answer this - building CI is must-have, I think. If we can get started with that, it'd be a huge step forward. We can try to test on the AWS instance but we cannot guarantee correctness, stability, or performance on it.

As stated earlier, we could also use two of the PyTorch CI MI200-class machines but it sounds that would be too much effort to integrate?

ramilbakhshyiev:
@iotamudelta Hey, apologies for missing that earlier offer, I missed that context when picking this up. I think getting AMD to provide backends would work great and I think we can support the integration. I have glanced through #3231 and PyTorch but did not figure out AMD servers register runners with GitHub. Any chance you can point me to that? I assume you're using PyTorch's GH App to retrieve API tokens and register runners to use.

ramilbakhshyiev:
> @iotamudelta Hey, apologies for missing that earlier offer, I missed that context when picking this up. I think getting AMD to provide backends would work great and I think we can support the integration. I have glanced through #3231 and PyTorch but did not figure out AMD servers register runners with GitHub. Any chance you can point me to that? I assume you're using PyTorch's GH App to retrieve API tokens and register runners to use.

@iotamudelta @ItsPitt

iotamudelta:
@ramilbakhshyiev we'll check on our side how the integration works in the PyTorch CI - if you could check on yours as well to connect the dots?

In the meantime, getting even the build functionality done would allow us to make some progress with integrating the code support.

ramilbakhshyiev:
@iotamudelta Pinged folks on our end regarding the PyTorch setup.

Re: build only setup, tried building this PR and running into conda dependency issues: https://github.com/facebookresearch/faiss/actions/runs/9863975175/job/27237915065?pr=3622

Is installing hip complier from conda-forge the best practice or do you have your own repository that I should be using?

ramilbakhshyiev:
@iotamudelta I have learned how we can register the runners you provide. Just to confirm: would these machines be continuously running? Would we get SSH access to them to configure them or would you need to configure them yourself?

iotamudelta:
@ramilbakhshyiev concerning setup - we'd prefer the solution PyTorch employs in their docker containers where it pulls directly from repo.radeon.com and versions the docker image w/ our ROCm version.

Concerning the machines, they'd be continuously running (unless there are HW issues or servicing, ofc). Let me check about ssh access.

ramilbakhshyiev:
@iotamudelta All of FAISS builds happen directly on the machines. Do you know if there is a way to install it using conda? If not, if we start with a container image, do you envision any issues if we install the rest with conda as usual?

iotamudelta:
@ramilbakhshyiev using our repos for the ROCm stack and rest from conda should compose, that's how PyTorch is setting things up: https://github.com/pytorch/pytorch/blob/main/.ci/docker/ubuntu-rocm/Dockerfile

ramilbakhshyiev:
@iotamudelta I think this is the code that sets up ROCm inside containers: https://github.com/pytorch/pytorch/blob/main/.ci/docker/common/install_rocm.sh#L9. Since we use ephemeral build machines for now, I will look into starting without introducing containers and we start using containers when we build on the machines you provided because those would be long-running. Wdyt?

ramilbakhshyiev:
@iotamudelta And another question to confirm: for build-only setup, we do not need physical GPUs on the machines, right? That's the case with our Nvidia builds but I just wanted to confirm that this is true for ROCm before we try to build it on a runner that doesn't have AMD GPUs.

iotamudelta:
@ramilbakhshyiev sounds good - I've reached out to our DevOps team of these machines.

Concerning build: there's a workaround where we can set `/opt/rocm/bin/target.lst` with the targeted architecture(s) and have `hipcc` pick that up:
```
gfx900
gfx906
gfx908
gfx90a
gfx1030
gfx1100
```
as an example - for 200 we'd only need to target `gfx90a` (to limit finalization time). If we wanted to build a release wheel, we would extend the list as needed.

ItsPitt:
@ramilbakhshyiev I have been using ROCm 6.1 on Ubuntu 22.04 for this effort. I have yet to run into any issues. You probably won't need some of the extra steps in that config. Here is the official install for ROCm 6.1.2: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/tutorial/quick-start.html

It looks pretty close. You might just need to add the latest 6.1.2 version number and make a few changes. Here are the lines I use in my docker:
`mkdir --parents --mode=0755 /etc/apt/keyrings
 wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
echo 'deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/6.1.2 jammy main' | sudo tee /etc/apt/sources.list.d/rocm.list
apt update && apt install -y rocm-dev6.1.2 rocm-libs6.1.2`

The miopen and amdgpu stuff can probably be removed. And the build target looks good. I hope this helps.

ramilbakhshyiev:
@ItsPitt Thank you! I think there is progress but it fails further down the line.

1. Can you please check if the installation is sufficient or if we are missing something? [latest build files](https://github.com/facebookresearch/faiss/pull/3622/files)
2. Can you please see what the next set of errors is and how to tackle those? [build output](https://github.com/facebookresearch/faiss/actions/runs/9903546385/job/27359255640?pr=3622)

ramilbakhshyiev:
@iotamudelta @ItsPitt I noticed `hipify.sh` file being introduced that creates the `faiss/gpu-rocm` subdir. Does it need to be called explicitly?

ItsPitt:
@ramilbakhshyiev Yes, it will need to be run before the cmake step. I run it from the top level faiss/ directory. 

`faiss/gpu/hipify.sh`

This will make a hip version of the gpu directory. It will also run hipify on c_api for support.

ItsPitt:
For your flags, you should only need `-DFAISS_ENABLE_GPU=ON`. No need for `-DFAISS_ENABLE_ROCM=ON`.

ramilbakhshyiev:
@ItsPitt I introduced the `ENABLE_ROCM` to be explicit to follow current patterns in the library. Happy to discuss now or when we have a PR to enable ROCm in CI.

I did add a step to call `hipify.sh`, the HIP module error still stayed. Can you please take a look and let me know if you have any thoughts? https://github.com/facebookresearch/faiss/actions/runs/9911542730/job/27384449480?pr=3622

ItsPitt:
@ramilbakhshyiev It looks like it cannot find the hip cmake module. You should be able to resolve this by adding the following to the cmake file:

`set(CMAKE_MODULE_PATH "/opt/rocm/hip/cmake" ${CMAKE_MODULE_PATH})`

The line above `find_package(HIP REQUIRED)` so it's inside the `if (USE_ROCM)` should work. 



> @ItsPitt I introduced the `ENABLE_ROCM` to be explicit to follow current patterns in the library. Happy to discuss now or when we have a PR to enable ROCm in CI.

Understood, sounds good!

ItsPitt:
Also, here is the documentation for including FindHIP. It has a few more examples: https://github.com/ROCm/HIP/tree/master/samples/2_Cookbook/12_cmake_hip_add_executable#including-findhip-cmake-module-in-the-project

ramilbakhshyiev:
Thanks @ItsPitt! I tried a few versions and it didn't solve the problem. Upon some investigating, I was only able to find `FindHIP.cmake` in `/opt/rocm-6.1.0/lib/cmake/hip/` (searching under the full `/opt` path). Do you know why the install happens this way and what I should do about it? I am trying that specific path right now but it still seems pretty far from the path expected in the documentation.

ramilbakhshyiev:
Made it to the next step but it failed then:
```
-- Found MKL: /home/runner/miniconda3/lib/libmkl_rt.so;-lpthread;-lm;-ldl  
CMAKE_MODULE_PATH: = /home/runner/work/faiss/faiss/cmake
CMAKE_MODULE_PATH: = /opt/rocm-6.1.0/lib/cmake/hip/;/home/runner/work/faiss/faiss/cmake
-- Found HIP: /opt/rocm-6.1.0 (found version "6.1.40091-a8dbc0c19") 
CMake Error at /opt/rocm-6.1.0/lib/cmake/hip/FindHIP.cmake:299 (message):
  Unable to find the clang compiler path.  Set ROCM_PATH or HIP_PATH in env
Call Stack (most recent call first):
  faiss/gpu-rocm/CMakeLists.txt:325 (find_package)
-- Configuring incomplete, errors occurred!
```

ItsPitt:
@ramilbakhshyiev I did some digging and was able to repo this. You just need to add `-DCMAKE_PREFIX_PATH=/opt/rocm` to your cmake build flags. Hope this helps!

ramilbakhshyiev:
Thanks @ItsPitt! It started building but now fails because it cannot find libm, I'm looking into it right now:
```
[ 79%] Linking HIP shared library libfaiss.so
clang++: warning: argument unused during compilation: '--offload-arch=gfx90a' [-Wunused-command-line-argument]
ld.lld: error: cannot open /lib64/libm.so.6: No such file or directory
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

ramilbakhshyiev:
Hey @ItsPitt, I want to provide an update. `libm` is available under `$CONDA_PREFIX/x86_64-conda-linux-gnu/sysroot/usr/lib64` on the runner machine but not directly in under `/lib64`. So far, I have tried to hint it to the linker with `target_link_directories` and `CMAKE_PREFIX_PATH` and now trying with updating the `LD_LIBRARY_PATH`. I have combed through the PR and not sure what breaks the library paths. If you have any thoughts or ideas, happy to try those.

ItsPitt:
@ramilbakhshyiev I have not run into that issue before. However, I am setting `LD_LIBRARY_PATH=/opt/rocm/lib` in my docker. I'm not sure if that's the root of your issue, but I'll link my whole docker below. I hope this helps!

https://github.com/ItsPitt/faiss_docker/tree/rocm_wf64_wf32

ramilbakhshyiev:
@ItsPitt @iotamudelta I have some updates. I have been able to build this PR in CI but with some workarounds (some are pretty brute force): [link to full diff](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19). This includes rebase onto main but here are the main files to pay attention to:
 * [.github/actions/build_cmake/action.yml](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-c5dee0272afe1826fd7f0936f07026a8c97865bd18ae5234490d22c4b0800800): install ROCm, call hipify.sh, do not build tests and do not run tests for ROCm
 * [CMakeLists.txt](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a): use `FAISS_USE_ROCM` instead of existence of /opt/rocm for enabling ROCm
 * [faiss/gpu/hipify.sh](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-6e994a0a8174cb70ea0a9e389db0e6004512fca7e206011074a0a1bfb8ba2df8): symlink required libraries from conda install -- we can move this to the ROCm install section of `action.yml` above; otherwise, it would be great to hear ideas on how to avoid this
 * [faiss/gpu/CMakeLists.txt](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-7f711500a7d6e68f360f6155318a3538423f4e42e980cb21c1d3899c3848f1d0): set `CMAKE_PREFIX_PATH` and `CMAKE_MODULE_PATH`, will remove messaging and LD_LIBRARY_PATH

Can you please review and let me know what you think? it's not ideal but at least we have a way to compile ROCm version.

Main questions:
1. Any updates on ssh access to the machines you would provide? If not, we can proceed with you runner scripts that we send to you to register the runners.
2. Should we be able to compile the testing code without having an AMD GPU on the host? It seemed to have failed for that reason.
3. In your setup, does ROCm install into /opt/rocm or /opt/rocm-{version}? I wanted to understand why it's installed versioned in our case and why CMake files are in a different spot than what the documentation mentions.
4. Do you have the list of libraries I symblinked in hipify.sh in /opt/rocm/lib? If yes, I will try to get the `LD_LIBRARY_CONFIG` updated in the CMake setup and use that.

If this is ok, I can review with the team if we are ok to proceed to rebase and review this PR to merge it in and then add the compile-only steps.

ItsPitt:
@ramilbakhshyiev  Glad to hear it's building!

> * [CMakeLists.txt](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a): use `FAISS_USE_ROCM` instead of existence of /opt/rocm for enabling ROCm

This makes sense. We could maybe enable gpu automatically if ROCm is enabled to avoid having to enable both. But this doesn't have to be done now.

> * [faiss/gpu/hipify.sh](https://github.com/facebookresearch/faiss/compare/922a4edeb2b6bc0ff8b5e1d2d3d6e44b1891dfca..c1e654e3c63c84eca1246d6588c73d3b5bfd5d19#diff-6e994a0a8174cb70ea0a9e389db0e6004512fca7e206011074a0a1bfb8ba2df8): symlink required libraries from conda install -- we can move this to the ROCm install section of `action.yml` above; otherwise, it would be great to hear ideas on how to avoid this

Moving this to `action.yml` would probably be best in avoiding issues when building on other systems. Since other systems won't have this file path and still need to run this script.

> 1. Any updates on ssh access to the machines you would provide? If not, we can proceed with you runner scripts that we send to you to register the runners.

I believe there's been some movement. I'll find out more details and get back to you on this.

> 2. Should we be able to compile the testing code without having an AMD GPU on the host? It seemed to have failed for that reason.

I'll look into this and get you an answer. Of course, if the testing code tries to make a device call as some kinda validation during building, this would cause an issue. Not sure that's what's going on here, just something to check. 

> 3. In your setup, does ROCm install into /opt/rocm or /opt/rocm-{version}? I wanted to understand why it's installed versioned in our case and why CMake files are in a different spot than what the documentation mentions.

When ROCm installs, it makes the symbolic link `/opt/rocm` to `/opt/rocm-{version}`. The actual install will be in `/opt/rocm-{version}`. Using the link just lets a user avoid updating the version number whenever they update ROCm. But using either should work.

This also helps for installing multiple versions of ROCm:
https://rocm.blogs.amd.com/software-tools-optimization/rocm-installation/README.html

> 4. Do you have the list of libraries I symblinked in hipify.sh in /opt/rocm/lib? If yes, I will try to get the `LD_LIBRARY_CONFIG` updated in the CMake setup and use that.

These look to be system libs and not part of ROCm.

ramilbakhshyiev:
@ItsPitt ack and thank you! I am currently trying to figure out the build errors with testing code, it actually seems to be failing to link against glib and also can't find one of the ROCm libs (I assume). Question on that: I assume AMD integrates into CMake/make setup, is that correct? Are there any recommended ways to debug why make does not find the libraries that it's looking for? I mean specific to ROCm/HIP setup.

ItsPitt:
@ramilbakhshyiev Do you know which ROCm lib(s) it was failing to link against?

It is correct. Out of curiosity, do you know which version of CMake you're using? It might be best if we start synchronizing dependency versions to eliminate factors between the working docker and the conda build.

ramilbakhshyiev:
@ItsPitt We use CMake `3.26` and the library that it couldn't find was `libamdhip64.so.6`. I provided the excerpt with errors below. Not sure why conda install 3.26, I will try with the version you specify in the dockerfile you provided.

```
2024-07-18T22:46:20.0232041Z [ 98%] Linking HIP executable TestGpuDistance
2024-07-18T22:46:20.2657473Z ld.lld: error: undefined symbol: __libc_single_threaded
2024-07-18T22:46:20.2659015Z >>> referenced by TestGpuDistance.hip
2024-07-18T22:46:20.2660505Z >>>               CMakeFiles/TestGpuDistance.dir/TestGpuDistance.hip.o:(testTransposition(bool, bool, faiss::MetricType, bool, float))
2024-07-18T22:46:20.2661882Z >>> referenced by TestGpuDistance.hip
2024-07-18T22:46:20.2663202Z >>>               CMakeFiles/TestGpuDistance.dir/TestGpuDistance.hip.o:(testTransposition(bool, bool, faiss::MetricType, bool, float))
2024-07-18T22:46:20.2664489Z >>> referenced by TestGpuDistance.hip
2024-07-18T22:46:20.2665766Z >>>               CMakeFiles/TestGpuDistance.dir/TestGpuDistance.hip.o:(testTransposition(bool, bool, faiss::MetricType, bool, float))
2024-07-18T22:46:20.2667013Z >>> referenced 3 more times
2024-07-18T22:46:20.2706393Z clang++: error: linker command failed with exit code 1 (use -v to see invocation)
2024-07-18T22:46:20.2754798Z make[2]: *** [faiss/gpu-rocm/test/CMakeFiles/TestGpuDistance.dir/build.make:101: faiss/gpu-rocm/test/TestGpuDistance] Error 1
2024-07-18T22:46:20.2756387Z make[2]: Leaving directory '/home/runner/work/faiss/faiss/build'
2024-07-18T22:46:20.2758640Z make[2]: Target 'faiss/gpu-rocm/test/CMakeFiles/TestGpuDistance.dir/build' not remade because of errors.
2024-07-18T22:46:20.2760522Z make[1]: *** [CMakeFiles/Makefile2:2209: faiss/gpu-rocm/test/CMakeFiles/TestGpuDistance.dir/all] Error 2
2024-07-18T22:46:25.0384186Z [100%] Linking HIP executable TestGpuSelect
2024-07-18T22:46:25.0995353Z ld.lld: error: undefined symbol: __libc_single_threaded
2024-07-18T22:46:25.0996060Z >>> referenced by TestGpuSelect.hip
2024-07-18T22:46:25.0997162Z >>>               CMakeFiles/TestGpuSelect.dir/TestGpuSelect.hip.o:(testForSize(int, int, int, bool, bool))
2024-07-18T22:46:25.0998075Z >>> referenced by TestGpuSelect.hip
2024-07-18T22:46:25.0999176Z >>>               CMakeFiles/TestGpuSelect.dir/TestGpuSelect.hip.o:(testForSize(int, int, int, bool, bool))
2024-07-18T22:46:25.1000179Z >>> referenced by TestGpuSelect.hip
2024-07-18T22:46:25.1001209Z >>>               CMakeFiles/TestGpuSelect.dir/TestGpuSelect.hip.o:(testForSize(int, int, int, bool, bool))
2024-07-18T22:46:25.1002295Z >>> referenced 2 more times
2024-07-18T22:46:25.1037876Z clang++: error: linker command failed with exit code 1 (use -v to see invocation)
2024-07-18T22:46:25.1083260Z make[2]: *** [faiss/gpu-rocm/test/CMakeFiles/TestGpuSelect.dir/build.make:101: faiss/gpu-rocm/test/TestGpuSelect] Error 1
2024-07-18T22:46:25.1085674Z make[2]: Target 'faiss/gpu-rocm/test/CMakeFiles/TestGpuSelect.dir/build' not remade because of errors.
2024-07-18T22:46:25.1087147Z make[1]: *** [CMakeFiles/Makefile2:2237: faiss/gpu-rocm/test/CMakeFiles/TestGpuSelect.dir/all] Error 2
2024-07-18T22:46:25.1088279Z make[2]: Leaving directory '/home/runner/work/faiss/faiss/build'
2024-07-18T22:46:28.9987835Z [100%] Linking CXX executable faiss_test
2024-07-18T22:46:29.1223839Z /home/runner/work/faiss/faiss/build/tests/faiss_test: error while loading shared libraries: libamdhip64.so.6: cannot open shared object file: No such file or directory
2024-07-18T22:46:29.1229698Z CMake Error at /home/runner/miniconda3/share/cmake-3.26/Modules/GoogleTestAddTests.cmake:112 (message):
2024-07-18T22:46:29.1230847Z   Error running test executable.
2024-07-18T22:46:29.1231217Z 
2024-07-18T22:46:29.1232350Z     Path: '/home/runner/work/faiss/faiss/build/tests/faiss_test'
2024-07-18T22:46:29.1233055Z     Result: 127
2024-07-18T22:46:29.1233429Z     Output:
2024-07-18T22:46:29.1233756Z       
2024-07-18T22:46:29.1233969Z 
2024-07-18T22:46:29.1234149Z Call Stack (most recent call first):
2024-07-18T22:46:29.1235416Z   /home/runner/miniconda3/share/cmake-3.26/Modules/GoogleTestAddTests.cmake:225 (gtest_discover_tests_impl)
2024-07-18T22:46:29.1236371Z 
2024-07-18T22:46:29.1236378Z 
2024-07-18T22:46:29.1245951Z make[2]: *** [tests/CMakeFiles/faiss_test.dir/build.make:567: tests/faiss_test] Error 1
```

ramilbakhshyiev:
I tried with 3.24 cmake and got the same error. I saw a commit on the dockerfile you provided to lower the cmake version from 3.29 to 3.24. What was the reason for doing that?

Meanwhile I am trying to match more versions to see if it helps.

ramilbakhshyiev:
I think I got a working build. I will clean things up, test it again, and update you with what I needed to change.

ItsPitt:
@ramilbakhshyiev Great to hear!

As for the CMake question, I hadn't fully validated the new version of CMake in my docker, so I rolled it back. No specific issue. It can more than likely be bumped once it's fully working.

faiss/gpu/CMakeLists.txt
```diff
@@ -304,6 +318,12 @@ foreach(header ${FAISS_GPU_HEADERS})
   )
 endforeach()
 
+if (USE_ROCM)
+find_package(HIP REQUIRED)
+find_package(hipBLAS REQUIRED)
+target_link_libraries(faiss_gpu PRIVATE hip::host roc::hipblas)
+target_compile_options(faiss_gpu PRIVATE)
+else()
```
ramilbakhshyiev, line 326: 1. indentation
2. Can we please move `find_package` statements to the top-level `CMakeLists.txt` guarded by `USE_ROCM`?
3. For `target_link_libraries`, can we please use the condition syntax like we do for RAFT? `$<$<BOOL:${USE_ROCM}>:hip::host>`

faiss/gpu/StandardGpuResources.cpp
```diff
@@ -365,8 +365,8 @@ void StandardGpuResourcesImpl::initializeForDevice(int device) {
 
     // Our code is pre-built with and expects warpSize == 32, validate that
     FAISS_ASSERT_FMT(
-            prop.warpSize == 32,
-            "Device id %d does not have expected warpSize of 32",
+            prop.warpSize == 32 || prop.warpSize == 64,
```
ramilbakhshyiev, line 368: Is this ROCm specific? If so, can we allow 64 only for ROCm?

faiss/gpu/test/CMakeLists.txt
```diff
@@ -17,12 +17,16 @@
...
-find_package(CUDAToolkit REQUIRED)
-
 # Defines `gtest_discover_tests()`.
 include(GoogleTest)
 add_library(faiss_gpu_test_helper TestUtils.cpp)
+if(USE_ROCM)
+find_package(HIP REQUIRED)
+target_link_libraries(faiss_gpu_test_helper PUBLIC faiss gtest hip::host)
+else()
+find_package(CUDAToolkit REQUIRED)
```
ramilbakhshyiev, line 27: Same as above:
1. Can we rely on top-level find package call?
2. Can we use conditional inclusion like we do with RAFT?

faiss/gpu/test/CMakeLists.txt
```diff
@@ -39,14 +43,25 @@ faiss_gpu_test(TestGpuIndexBinaryFlat.cpp)
 faiss_gpu_test(TestGpuMemoryException.cpp)
 faiss_gpu_test(TestGpuIndexIVFPQ.cpp)
 faiss_gpu_test(TestGpuIndexIVFScalarQuantizer.cpp)
+if(USE_ROCM)
+faiss_gpu_test(TestGpuDistance.hip)
+faiss_gpu_test(TestGpuSelect.hip)
+else()
 faiss_gpu_test(TestGpuDistance.cu)
 faiss_gpu_test(TestGpuSelect.cu)
+
```
ramilbakhshyiev, line 52: Can we use $HIP_OR_CU here?

faiss/gpu/test/CMakeLists.txt
```diff
@@ -39,14 +43,25 @@ faiss_gpu_test(TestGpuIndexBinaryFlat.cpp)
...
 add_executable(demo_ivfpq_indexing_gpu EXCLUDE_FROM_ALL
   demo_ivfpq_indexing_gpu.cpp)
 
+if (USE_ROCM)
+target_link_libraries(demo_ivfpq_indexing_gpu
+  PRIVATE faiss gtest_main hip::host)
+else()
 target_link_libraries(demo_ivfpq_indexing_gpu
   PRIVATE faiss gtest_main CUDA::cudart)
+endif()
```
ramilbakhshyiev, line 67: Can we do conditional here please?

faiss/python/CMakeLists.txt
```diff
@@ -125,13 +139,20 @@ else()
...
 if(FAISS_ENABLE_GPU)
-  find_package(CUDAToolkit REQUIRED)
-  if(FAISS_ENABLE_RAFT)
-    find_package(raft COMPONENTS compiled distributed)
+  if(USE_ROCM)
+    find_package(HIP REQUIRED)
+    target_link_libraries(swigfaiss PRIVATE hip::host)
+    target_link_libraries(swigfaiss_avx2 PRIVATE hip::host)
+    target_link_libraries(swigfaiss_avx512 PRIVATE hip::host)
+  else()
```
ramilbakhshyiev, line 147: Can we use conditional statements here and elsewhere in this file too?

faiss/gpu/CMakeLists.txt
```diff
@@ -197,6 +197,16 @@ function(generate_ivf_interleaved_code)
     "64|2048|8"
   )
 
+  if(USE_ROCM)
+    set(CU_OR_HIP "hip")
```
ramilbakhshyiev, line 201: Could you name it something like `GPU_EXT_PREFIX`?

ramilbakhshyiev:
@ItsPitt @iotamudelta I found a workaround to build the tests in our CI environment so I think we should be good for build-only setup for now. I left some initial comments on the PR. Can you please rebase as well? Once we address the comments, I will run the PR by the team and we will figure what else needs to happen so we could merge it. After that, I will publish the PR to enable build-only.

ramilbakhshyiev:
Here's a successful build with testing enabled: https://github.com/facebookresearch/faiss/actions/runs/10085927308/job/27887614699?pr=3622

faiss/gpu/StandardGpuResources.cpp
ItsPitt: We have both wavefront 32 (E.g. navi) and 64 (E.g. MI250) devices. So this offers support for both.

ramilbakhshyiev: It sounds like Nvidia is 32 only and ROCm is 32 or 64. Should we lock it accordingly in code?

ItsPitt: If that is desired, I could rework that assert using a ROCm flag to only allow a warpSize of 64 (and 32) on ROCm devices. It shouldn't be an issue at all!

ItsPitt:
@ramilbakhshyiev I'll work on these changes and get back to you.

faiss/gpu/StandardGpuResources.cpp
ramilbakhshyiev: Yeah, I think I would do that.

ramilbakhshyiev:
> @ramilbakhshyiev I'll work on these changes and get back to you.

Sounds great, thank you!

ItsPitt:
@ramilbakhshyiev I believe I address all issues. If there are more or I missed any, just let me know!

Fixed indentations.
Moved `find_package` to the top level CMake.
Changed all `target_link_libraries` for `hip::host` to `$<$<BOOL:${USE_ROCM}>:hip::host>`.
Changed the warpSize assert to allow a size of 64 only on ROCm. While allowing a size of 32 on all devices.
Moved `$HIP_OR_CU` to the top level CMake for use in faiss/gpu/CMakeLists.txt. Also changed the name to `$GPU_EXT_PREFIX`.
Replaced file extensions (such as .hip) with `$GPU_EXT_PREFIX`

ramilbakhshyiev:
@ItsPitt That's great, thank you so much! Could you resolve conflicts and rebase one more time? I will be import this and send to team for review as soon as you do.

iotamudelta:
@ramilbakhshyiev rebased and merged - hopefully correctly. Thanks for working on this!

faiss/gpu/test/CMakeLists.txt
```diff
@@ -17,12 +17,15 @@
...
 # Defines `gtest_discover_tests()`.
 include(GoogleTest)
 add_library(faiss_gpu_test_helper TestUtils.cpp)
-target_link_libraries(faiss_gpu_test_helper PUBLIC faiss gtest CUDA::cudart $<$<BOOL:${FAISS_ENABLE_RAFT}>:raft::raft> $<$<BOOL:${FAISS_ENABLE_RAFT}>:raft::compiled>)
+if(USE_ROCM)
+  target_link_libraries(faiss_gpu_test_helper PUBLIC faiss gtest $<$<BOOL:${USE_ROCM}>:hip::host>)
+else()
+  find_package(CUDAToolkit REQUIRED)
+  target_link_libraries(faiss_gpu_test_helper PUBLIC faiss gtest CUDA::cudart $<$<BOOL:${FAISS_ENABLE_RAFT}>:raft::raft> $<$<BOOL:${FAISS_ENABLE_RAFT}>:raft::compiled>)
+endif()
```
ramilbakhshyiev, line 28: @iotamudelta @ItsPitt This PR breaks the TestCodePacking tests internally. While I am looking at it, do you think we remove collapse these if-else statements throughout the CMakeLists files where possible? That was the intent behind my original comment to use the inline condition includes.

ItsPitt: @ramilbakhshyiev Just to clarify, are these if-else conditions breaking the TestCodePacking, or is that a separate issue?

ramilbakhshyiev: Separate issue. I just didn't notice these yesterday when I was going over the diff after you addressed the comments, my apologies.

ramilbakhshyiev: If you're curious, here are the errors I am getting:
```
Faiss assertion 'err__ == cudaSuccess' failed in int faiss::gpu::getCurrentDevice() at fbcode/faiss/gpu/utils/DeviceUtils.cu:23; details: CUDA error 35 CUDA driver version is insufficient for CUDA runtime version
```

I think that's because of the call to get warp size from the device properties. I'm trying to understand why that fails internally.

ramilbakhshyiev: I figured out the internal error and published a change for that. It should fix this PR failures internally.

ramilbakhshyiev:
@iotamudelta I already started landing the PR so I will pick up your latest commit in a PR that I need to do to prep for enabling build-only AMD CI runs.

@iotamudelta @ItsPitt Any updates on the runners you offered to provide?

iotamudelta:
@ramilbakhshyiev awesome, thank you! We'll push the remaining diff up in another PR.

Concerning test machines - @ItsPitt was able to log into them. We hope it'll be straightforward to hook them up to Faiss.

ramilbakhshyiev:
OK, let me get the current state to build-only for now and then we'll proceed to building and testing on the machines you provide. So the plan is that we will provide you with the scripts to register and you will run them on the machines? For security reasons, we will probably need to do this over zoom or email, you might have the context from doing this for PyTorch.

faiss/gpu/utils/BlockSelectKernel.cuh
```diff
@@ -26,45 +26,47 @@ __global__ void blockSelect(
...
-    if (i < in.getSize(1)) {
-        heap.addThreadQ(*inStart, (IndexType)i);
-    }
-
-    heap.reduce();
-
-    for (int i = threadIdx.x; i < k; i += ThreadsPerBlock) {
-        outK[row][i] = smemK[i];
-        outV[row][i] = smemV[i];
+    if constexpr ((NumWarpQ == 1 && NumThreadQ == 1) || NumWarpQ >= kWarpSize) {
```
simshi, line 29: why not a static_assert, add one more indention looks ugly 😄

ramilbakhshyiev: @iotamudelta ^^

ItsPitt: I tested this (and the others cases like it). When changing from an 'if' statement to a 'static_assert' I received compile errors. From what I can tell, it is checking this logic at build time for resolving warpSize. With an 'if', it simply falls through if this check fails and does not error out the build. Then during run time it resolves the warpSize and works correctly. With a 'static_assert', it just fails the build.

DocMAX:
Can we run Faiss with ZLUDA?

ItsPitt:
> Can we run Faiss with ZLUDA?

ZLUDA is outside of the scope of this PR and has not been tested. However, this PR offers native support for wavefront 32 and 64 AMD GPUs.

PR #3487, opened 2024-05-29
add skip_storage flag to HNSW

mdouze:
Summary:
Sometimes it is not useful to serialize the storage index along with a HNSW index. This diff adds a flag that supports skipping the storage of the index.

Searchign and adding to the index is not possible until a storage index is added back in.

Differential Revision: D57911060

### DIFF
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -5,8 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-// -*- c++ -*-
-
 #include <faiss/index_io.h>
 
 #include <faiss/impl/io_macros.h>
@@ -531,7 +529,11 @@ Index* read_index(IOReader* f, int io_flags) {
     Index* idx = nullptr;
     uint32_t h;
     READ1(h);
-    if (h == fourcc("IxFI") || h == fourcc("IxF2") || h == fourcc("IxFl")) {
+    if (h == fourcc("null")) {
+        // denotes a missing index, useful for some cases
+        return nullptr;
+    } else if (
+            h == fourcc("IxFI") || h == fourcc("IxF2") || h == fourcc("IxFl")) {
         IndexFlat* idxf;
         if (h == fourcc("IxFI")) {
             idxf = new IndexFlatIP();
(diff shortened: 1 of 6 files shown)


faiss/impl/index_write.cpp
```diff
@@ -765,6 +763,12 @@ void write_index(const Index* idx, IOWriter* f) {
         WRITE1(h);
         write_index_header(idxhnsw, f);
         write_HNSW(&idxhnsw->hnsw, f);
+        if (io_flags & IO_FLAG_SKIP_STORAGE) {
+            uint32_t n4 = fourcc("null");
+            WRITE1(n4);
+        } else {
+            write_index(idxhnsw->storage, f);
+        }
         write_index(idxhnsw->storage, f);
```
alexanderguzhva, line 772: writing the storage twice, dupilcate rows 770 and 772

alexanderguzhva (changes requested):

faiss/impl/index_read.cpp
```diff
@@ -531,7 +530,10 @@ Index* read_index(IOReader* f, int io_flags) {
     Index* idx = nullptr;
     uint32_t h;
     READ1(h);
-    if (h == fourcc("IxFI") || h == fourcc("IxF2") || h == fourcc("IxFl")) {
+    if (h == fourcc("null")) {
```
junjieqi, line 533: When do we see the caller provide null ?

faiss/impl/index_write.cpp
```diff
@@ -765,6 +763,12 @@ void write_index(const Index* idx, IOWriter* f) {
         WRITE1(h);
         write_index_header(idxhnsw, f);
         write_HNSW(&idxhnsw->hnsw, f);
+        if (io_flags & IO_FLAG_SKIP_STORAGE) {
+            uint32_t n4 = fourcc("null");
+            WRITE1(n4);
```
junjieqi, line 768: is it gonna to output in stdout?

tests/test_graph_based.py
```diff
@@ -133,6 +133,38 @@ def test_ndis_stats(self):
...
+
+        # add storage afterwards
+        index.storage = faiss.clone_index(index.storage)
+        index.own_fields = True
+
+        Dnew, Inew = index.search(self.xq, 5)
+        np.testing.assert_array_equal(Dnew, Dref)
+        np.testing.assert_array_equal(Inew, Iref)
+
+        if False:
```
junjieqi, line 160: where do we define the False?

junjieqi:
format & internal linter

PR #3492, opened 2024-06-02
Update .gitignore

abhiramvad:
Adding build folder to gitignore, so that they don't show up in the commit tree while building from source

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.dylib
 *.pyc
 *~
+/build/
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
lgtm, totally makes sense when building for OSS

abhiramvad:
> lgtm, totally makes sense when building for OSS

Thanks, I have signed the CLA, but the check is not passing. Can you help me on this please?

PR #3523, opened 2024-06-17
Update RAFT to 24.06

divyegala:
As of [commit](https://github.com/facebookresearch/faiss/commit/df0dea6c6d8951056763dc03528b3973c6ba26e2) in branch `main`, CAGRA was inoperable because FAISS was using RAFT 24.04 which was missing a feature. This PR updates RAFT to 24.06 which is our next stable version and contains the said missing feature.


ramilbakhshyiev:
@divyegala My apologies, I didn't notice that you published a PR and I published and pushed this change after testing it out and confirming with you over email. However, I had a question: do you think we could just change to use `24.*` so we don't have to upgrade for every stable version? Also, can we rely solely on rapidsai channel (assuming this one has stable builds) and remove rapidsai-nightly from our conda recipe?

Issue #3532, opened 2024-06-21
Build error on Faiss due to missing #includes


mfixman:
# Summary

Compiling Faiss using cmake on my Linux desktop produces a build error.

```
~/packages/faiss $ cmake -B build . -DBUILD_SHARED_LIBS=ON -DFAISS_ENABLE_C_API=ON
# [...]
~/packages/faiss $ cmake --build build -j20
# [...]
~/packages/faiss/tests/test_ivf_index.cpp:232:26: error: no matching function for call to ‘fi
nd(std::set<long unsigned int>::iterator, std::set<long unsigned int>::iterator, long unsigned int&)’    
  232 |                 std::find(                                                                       
      |                 ~~~~~~~~~^                                                                       
  233 |                         lists_probed.cbegin(),                                                   
      |                         ~~~~~~~~~~~~~~~~~~~~~~                                                   
  234 |                         lists_probed.cend(),                                                     
      |                         ~~~~~~~~~~~~~~~~~~~~                                                     
  235 |                         query_vector_listno) != lists_probed.cend())                             
      |                         ~~~~~~~~~~~~~~~~~~~~                                                     

```

This can be solved by including `set` and `algorithm` to `test_ivf_index.cpp`.

# Platform

OS: Arch Linux

Faiss version: `e758973fa08164728eb9e136631fe6c57d7edf6c`

Installed from: Attempting to compile with cmake

Faiss compilation options: `-DBUILD_SHARED_LIBS=ON -DFAISS_ENABLE_C_API=ON`

Running on:
- [ ] CPU
- [ ] GPU
- [x] Compiling for GPU 

Interface: 
- [ ] C++
- [ ] Python
- [x] Cmake

# Reproduction instructions

```
~/packages/faiss $  cmake -B build . -DBUILD_SHARED_LIBS=ON -DFAISS_ENABLE_C_API=ON
~/packages/faiss $ cmake --build build -j20

[ 87%] Building CXX object c_api/CMakeFiles/faiss_c.dir/gpu/StandardGpuResources_c.cpp.o                 
In file included from /home/mfixman/packages/faiss/build/_deps/googletest-src/googletest/include/gtest/gt
est-printers.h:114,                                                                                      
                 from /home/mfixman/packages/faiss/build/_deps/googletest-src/googletest/include/gtest/gt
est-matchers.h:48,                                                                                       
                 from /home/mfixman/packages/faiss/build/_deps/googletest-src/googletest/include/gtest/in
[...]

PR #3533, opened 2024-06-21
Adding missing includes which are necessary for building.

mfixman:
The current version of Faiss cannot be built due to missing `#include`s on a test file `tests/test_ivf_index.cpp`. This is better described in issue [#3532](https://github.com/facebookresearch/faiss/issues/3532).

This adds the missing imports.

### DIFF
--- a/tests/test_ivf_index.cpp
+++ b/tests/test_ivf_index.cpp
@@ -12,6 +12,8 @@
 #include <cstring>
 #include <map>
 #include <random>
+#include <set>
+#include <algorithm>
 
 #include <gtest/gtest.h>
 


tests/test_ivf_index.cpp
```diff
@@ -12,6 +12,8 @@
 #include <cstring>
 #include <map>
 #include <random>
+#include <set>
+#include <algorithm>
```
junjieqi, line 16: How about removing algorithm? I don't see where we use it

mfixman: It's used in a `std::find` call in this file.

junjieqi: Hi @mfixman, I think you are right. I'm wondering if you could put #include<alogrithm> in between #include <omp.h> and #include <cstddef>. The format doesn't pass. Once you fix it, I could merge it

junjieqi:
fixed

PR #3552, opened 2024-06-26
Fix seg faults in CAGRA C++ unit tests

divyegala:
The issue was that `uniform_int_distribution` generates numbers in the range `[0, ntotal]` and not `[0, ntotal)`, which was an oversight on my part. This PR also attempts to reduce the tolerance for `copyTo` tests as we have seen those fail intermittently.

cc @ramilbakhshyiev @mdouze @cjnolet

### DIFF
--- a/faiss/IndexHNSW.cpp
+++ b/faiss/IndexHNSW.cpp
@@ -463,9 +463,7 @@ void IndexHNSW::search_level_0(
             vt.advance();
         }
 #pragma omp critical
-        {
-            hnsw_stats.combine(search_stats);
-        }
+        { hnsw_stats.combine(search_stats); }
     }
     if (is_similarity_metric(this->metric_type)) {
 // we need to revert the negated distances
--- a/faiss/IndexHNSW.cpp
+++ b/faiss/IndexHNSW.cpp
@@ -463,7 +463,9 @@ void IndexHNSW::search_level_0(
             vt.advance();
         }
 #pragma omp critical
-        { hnsw_stats.combine(search_stats); }
+        {
+            hnsw_stats.combine(search_stats);
+        }
     }
     if (is_similarity_metric(this->metric_type)) {
 // we need to revert the negated distances
@@ -948,7 +950,7 @@ void IndexHNSWCagra::search(
 
             std::random_device rd;
             std::mt19937 gen(rd());
-            std::uniform_int_distribution<idx_t> distrib(0, this->ntotal);
+            std::uniform_int_distribution<idx_t> distrib(0, this->ntotal - 1);
 
             for (idx_t j = 0; j < num_base_level_search_entrypoints; j++) {
                 auto idx = distrib(gen);
--- a/faiss/gpu/test/TestGpuIndexCagra.cu
+++ b/faiss/gpu/test/TestGpuIndexCagra.cu
@@ -198,7 +198,7 @@ void copyToTest(
         faiss::MetricType metric,
         double expected_recall,
         bool base_level_only) {
-    for (int tries = 0; tries < 5; ++tries) {
+    for (int tries = 0; tries < 100; ++tries) {
         Options opt;
         if (opt.buildAlgo == faiss::gpu::graph_build_algo::NN_DESCENT &&
             metric == faiss::METRIC_INNER_PRODUCT) {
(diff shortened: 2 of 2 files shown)


faiss/gpu/test/TestGpuIndexCagra.cu
```diff
@@ -198,7 +198,7 @@ void copyToTest(
         faiss::MetricType metric,
         double expected_recall,
         bool base_level_only) {
-    for (int tries = 0; tries < 5; ++tries) {
+    for (int tries = 0; tries < 100; ++tries) {
```
ramilbakhshyiev, line 201: Do we still need to increase this or is this a testing artifact? I think this increases probability but does not make this certain.

faiss/gpu/test/TestGpuIndexCagra.cu
```diff
@@ -330,6 +330,9 @@ void copyToTest(
                 recall_score.view(),
                 copy_ref_dis_mds_opt,
                 ref_dis_mds_opt);
+        std::cout << "run: " << tries
+                  << ", recall_score: " << *recall_score.data_handle()
```
ramilbakhshyiev, line 334: Can we get the actual value once and re-used just to be safe?

faiss/gpu/test/TestGpuIndexCagra.cu
```diff
@@ -468,7 +471,7 @@ int main(int argc, char** argv) {
     testing::InitGoogleTest(&argc, argv);
 
     // just run with a fixed test seed
-    faiss::gpu::setTestSeed(100);
+    faiss::gpu::setTestSeed(101);
```
ramilbakhshyiev, line 474: What is the purpose of this?

faiss/gpu/test/TestGpuIndexCagra.cu
divyegala: Testing artifact, just to see if we get intermittent accuracy drops. I'll change this once we are comfortable merging this PR in.

faiss/gpu/test/TestGpuIndexCagra.cu
divyegala: As mentioned offline, the intermittent accuracy drops looked to me like they were a consequence of random number generating train/query vectors with not good distribution. To verify that hypothesis, I changed the seed. At least locally, I don't see one-off accuracy drops anymore with change in seed.

faiss/gpu/test/TestGpuIndexCagra.cu
ramilbakhshyiev: Assuming this is a testing artifact as well.

faiss/gpu/test/TestGpuIndexCagra.cu
ramilbakhshyiev: Got it. I think this is fine but it seems to have been succeeding even with the previous seed value on at least 1 build. Just FYI.

divyegala: Just to make sure we don't run into that case ever again, and seeding will help with that.

faiss/gpu/test/TestGpuIndexCagra.cu
divyegala: Yes, testing artifact too.

faiss/IndexHNSW.cpp
```diff
@@ -948,7 +948,7 @@ void IndexHNSWCagra::search(
 
             std::random_device rd;
             std::mt19937 gen(rd());
-            std::uniform_int_distribution<idx_t> distrib(0, this->ntotal);
+            std::uniform_int_distribution<idx_t> distrib(0, this->ntotal - 1);
```
alexanderguzhva, line 951: can this->ntotal be 0?

divyegala: Only if vectors have not been added to the index, then calling `search` is invalid in any case.

faiss/gpu/test/TestGpuIndexCagra.cu
```diff
@@ -330,6 +330,9 @@ void copyToTest(
                 recall_score.view(),
                 copy_ref_dis_mds_opt,
                 ref_dis_mds_opt);
+        std::cout << "run: " << tries
+                  << ", recall_score: " << *recall_score.data_handle()
+                  << std::endl;
         ASSERT_TRUE(*recall_score.data_handle() > expected_recall);
```
ramilbakhshyiev, line 336: Please consider to change this to the line below as well. It should output the values on error which is convenient.

```
ASSERT_GT(*recall_score.data_handle(), expected_recall);
```

PR #3602, opened 2024-07-01
Non-Blocking AVX512 Build on self-hosted github runner

mengdilin:
Start beta-testing AVX512 Build on self-hosted github runner with label faiss-aws-m7i.xlarge. This build is non-blocking on the PRs right now (via the parameter `continue-on-error`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,6 +66,7 @@ jobs:
       - uses: ./.github/actions/build_cmake
         with:
           opt_level: avx512
+        continue-on-error: true # non-blocking mode for now
   linux-x86_64-GPU-cmake:
     name: Linux x86_64 GPU (cmake)
     needs: linux-x86_64-cmake
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,7 @@ jobs:
   linux-x86_64-AVX512-cmake:
     name: Linux x86_64 AVX512 (cmake)
     continue-on-error: true # non-blocking mode for now
+    needs: linux-x86_64-cmake
     runs-on: faiss-aws-m7i.xlarge
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -56,11 +56,11 @@ jobs:
...
   linux-x86_64-AVX512-cmake:
     name: Linux x86_64 AVX512 (cmake)
-    if: false # TODO: enable when GitHub Actions adds AVX-512 hosts
-    needs: linux-x86_64-cmake
-    runs-on: ubuntu-latest
+    continue-on-error: true # non-blocking mode for now
+    runs-on: faiss-aws-m7i.xlarge
     steps:
       - name: Checkout
+        continue-on-error: true # non-blocking mode for now
```
ramilbakhshyiev, line 63: Do we need this line here?

.github/workflows/build.yml
```diff
@@ -56,11 +56,11 @@ jobs:
           opt_level: avx2
   linux-x86_64-AVX512-cmake:
     name: Linux x86_64 AVX512 (cmake)
-    if: false # TODO: enable when GitHub Actions adds AVX-512 hosts
-    needs: linux-x86_64-cmake
```
ramilbakhshyiev, line 60: Could we please gate avx-512 behind the plain vanilla build like all others?

`needs: linux-x86_64-cmake`

.github/workflows/build.yml
mengdilin: I observed a difference in UI behaviors for `continue-on-error` when set on a job vs on a step: setting it on a job means the action is successful but people will still see the red check mark on the github PR page. I wanted the red check mark to go away to avoid confusion and that means we also need `continue-on-error` set on the step level. There has been a discussion about this behavior in: https://github.com/orgs/community/discussions/15452

ramilbakhshyiev: Interesting. So it's different when it's set at job level and the step level. Do we need to set it on the build_cmake as well then?

mengdilin: Ah nice catch!

ramilbakhshyiev (approved):

PR #3609, opened 2024-07-03
Adding missing includes which are necessary for building

junjieqi:
Summary:
The current version of Faiss cannot be built due to missing #includes on a test file tests/test_ivf_index.cpp. This is better described in issue #3532.

https://github.com/facebookresearch/faiss/pull/3533

Differential Revision: D59314273


junjieqi:
@mfixman just a quick update

PR #3611, opened 2024-07-04
Add search functionality to FlatCodes

mdouze:
Summary: Using the new dispatcher functions, add search func to flat codes. To test it, make IndexLattice a subclass of FlatCodes and check the resonstruction there.

Differential Revision: D59367989

### DIFF
--- a/faiss/IndexFlatCodes.cpp
+++ b/faiss/IndexFlatCodes.cpp
@@ -12,6 +12,8 @@
 #include <faiss/impl/DistanceComputer.h>
 #include <faiss/impl/FaissAssert.h>
 #include <faiss/impl/IDSelector.h>
+#include <faiss/impl/ResultHandler.h>
+#include <faiss/utils/extra_distances.h>
 
 namespace faiss {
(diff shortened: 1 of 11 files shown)


faiss/IndexFlatCodes.cpp
```diff
@@ -114,4 +111,163 @@ void IndexFlatCodes::permute_entries(const idx_t* perm) {
...
+            const idx_t idx1,
+            const idx_t idx2,
+            const idx_t idx3,
+            float& dis0,
+            float& dis1,
+            float& dis2,
+            float& dis3) {
+        uint8_t* cp = code_buffer.data();
+        for (idx_t i : {idx0, idx1, idx2, idx3}) {
+            memcpy(cp, codes + i * code_size, code_size);
```
alexanderguzhva, line 163: Let's assume that the code size is 1536 floats. This means that every function call costs extra 4 * 4 * 1536 bytes read and written. If I understand correctly, it should affect the performance noticeably.

PR #3627, opened 2024-07-11
add dispatcher for VectorDistance and ResultHandlers

mdouze:
Summary: Add dispatcher function to avoid repeating dispatching code for distance computation and result handlers.

Reviewed By: asadoughi

Differential Revision: D59318865


asadoughi:
D59318865 was already committed.

PR #3632, opened 2024-07-13
Update IDSelector to support vector filtering by distance

jaepil:
Currently, there is no way to select vectors using distance with `IDSelector`. This PR adds overloaded version of `is_member(idx, [d])` function to `IDSelector`, so the user can make a decision based on the distance.

Let me know, if Faiss already supports similar functionality in different ways.


mnorris11:
Hi @jaepil , sorry for the delay. I took a look through it.

Is range_search not suitable to your use case?

If range_search is not right here, my first questions are:
1. Can you share more info on why you need distance for your use case?
2. Can you add a unit test demonstrating/testing it?
3. Is this applicable for other indexes besides HNSW?

mnorris11:
Closing PR at team's request.

PR #3635, opened 2024-07-16
1720 - expose FAISS version field to c_api

bshethmeta:
Summary: Add a util function to return the version in the c api.

Differential Revision: D59817407

### DIFF
--- a/faiss/Index.h
+++ b/faiss/Index.h
@@ -20,6 +20,18 @@
 #define FAISS_VERSION_MINOR 8
 #define FAISS_VERSION_PATCH 0
 
+// Macro to combine the version components into a single string
+#ifndef FAISS_STRINGIFY
+#define FAISS_STRINGIFY(ARG) #ARG
+#endif
+#ifndef FAISS_TOSTRING
+#define FAISS_TOSTRING(ARG) FAISS_STRINGIFY(ARG)
+#endif
+#define VERSION_STRING                                          \
+    FAISS_TOSTRING(FAISS_VERSION_MAJOR)                         \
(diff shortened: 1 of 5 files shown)


faiss/Index.h
```diff
@@ -20,6 +20,17 @@
...
 #define FAISS_VERSION_PATCH 0
 
+// Macro to combine the version components into a single string
+#ifndef STRINGIFY
+#define STRINGIFY(ARG) #ARG
+#endif
+#ifndef TOSTRING
+#define TOSTRING(ARG) STRINGIFY(ARG)
+#endif
+#define VERSION_STRING            \
```
Enet4, line 30: This definition should probably be prefixed with `FAISS_`, otherwise it might collide with other symbols.

faiss/utils/utils.h
```diff
@@ -37,6 +37,9 @@ std::string get_compile_options();
  * Get some stats about the system
  **************************************************/
 
+// Expose FAISS version as a string
+std::string get_version();
```
Enet4, line 41: It feels a bit of a waste to produce an allocated string that is already known at compile time. It also does not contribute to exposing this version to the C API.

Enet4 (commented):
I assume that this is a work in progress, but I might as well leave some feedback per its current state.

faiss/utils/utils.h
bshethmeta: > It also does not contribute to exposing this version to the C API

I am not sure I understand, can you explain?

Enet4: `std::string get_version()` is not a function signature compatible with C due to the C++ string type.

Ultimately, this PR claims to extend the C API while not touching the C API at all. :)

PR #3653, opened 2024-07-19
Add ARM64 build to build_cmake actions for SVE

mengdilin:
Add instructions to download arm64 specific conda dependencies and cmake command and run it on CI. This should prepare us to turn on CI with SVE optimization

### DIFF
--- a/.github/actions/build_cmake/action.yml
+++ b/.github/actions/build_cmake/action.yml
@@ -28,9 +28,20 @@ runs:
         conda update -y -q conda
         echo "$CONDA/bin" >> $GITHUB_PATH
 
-        # install base packages
-        conda install -y -q -c conda-forge gxx_linux-64=11.2 sysroot_linux-64=2.28
-        conda install -y -q python=3.11 cmake make swig mkl=2023 mkl-devel=2023 numpy scipy pytest
+        conda install -y -q python=3.11 cmake make swig numpy scipy pytest
+
+        # install base packages for ARM64
+        if [ "${{ runner.arch }}" = "ARM64" ]; then
+          conda install -y -q -c conda-forge openblas gxx_linux-aarch64 sysroot_linux-aarch64
+        fi
+
+        # install base packages for X86_64
+        if [ "${{ runner.arch }}" = "X64" ]; then
+          # TODO: unpin versions for gxx_linux-64 and sysroot_linux-64 and merge it with ARM64 below
+          conda install -y -q -c conda-forge gxx_linux-64=11.2 sysroot_linux-64=2.28
+          conda install -y -q mkl=2023 mkl-devel=2023 
+        fi
+
 
         # install CUDA packages
         if [ "${{ inputs.gpu }}" = "ON" ] && [ "${{ inputs.raft }}" = "OFF" ]; then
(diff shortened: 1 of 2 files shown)


.github/actions/build_cmake/action.yml
```diff
@@ -67,6 +69,35 @@ runs:
               -DCMAKE_CUDA_FLAGS="-gencode arch=compute_75,code=sm_75" \
               .
         make -k -C build -j$(nproc)
+    - name: Configure build environment in Linux aarch64
```
ramilbakhshyiev, line 72: Significant overlap with the configure section about, could we please merge?

.github/actions/build_cmake/action.yml
```diff
@@ -20,17 +20,27 @@ runs:
       with:
         python-version: '3.11'
         miniconda-version: latest
-    - name: Configure build environment
```
ramilbakhshyiev, line 23: This is only built on Linux, right? If so, we can keep it.

.github/actions/build_cmake/action.yml
```diff
@@ -49,7 +59,7 @@ runs:
         else
           conda install -y -q pytorch -c pytorch
         fi
-    - name: Build all targets
+    - name: Build all targets on Linux
```
ramilbakhshyiev, line 62: Same "on Linux" comment as above.

.github/actions/build_cmake/action.yml
```diff
@@ -20,17 +20,27 @@ runs:
...
         # initialize Conda
         conda config --set solver libmamba
         conda update -y -q conda
         echo "$CONDA/bin" >> $GITHUB_PATH
 
-        # install base packages
-        conda install -y -q -c conda-forge gxx_linux-64=11.2 sysroot_linux-64=2.28
-        conda install -y -q python=3.11 cmake make swig mkl=2023 mkl-devel=2023 numpy scipy pytest
+        # install base packages for X86_64
+        if [ "${{ inputs.gpu }}" = "ON" ]; then
```
ramilbakhshyiev, line 32: I think we need to guard this section with the arm64 below

.github/actions/build_cmake/action.yml
```diff
@@ -20,17 +20,27 @@ runs:
...
+          # TODO: unpin versions for gxx_linux-64 and sysroot_linux-64 and merge it with ARM64 below
+          conda install -y -q -c conda-forge gxx_linux-64=11.2 sysroot_linux-64=2.28
+          conda install -y -q mkl=2023 mkl-devel=2023 
+        fi
+
+        # install base packages for ARM64
+        if [ "${{ inputs.gpu }}" = "ARM64" ]; then
```
ramilbakhshyiev, line 39: not `input.gpu`, probably `runner.arch`, same as above

.github/actions/build_cmake/action.yml
ramilbakhshyiev: I'd merge this into an if-else with the below section: if ARM64 else

.github/workflows/build.yml
```diff
@@ -88,6 +88,19 @@ jobs:
         with:
           gpu: ON
           raft: ON
+  linux-arm64-SVE-cmake:
+    name: Linux arm64 (cmake)
```
ramilbakhshyiev, line 92: might want to add "SVE" in here

PR #3687, opened 2024-07-24
Add warning on adding nbits to LSH index factory

bshethmeta:
Summary: We will write a warning if nbits is not specified while using index factory with LSH. The warning lets users know we will be using default d as nbits.

Differential Revision: D60187935


mdouze:
Hmm not sure we want to warn users for that, since having nbits = d is the most common use case. 
Often the pre-transform is setting the number of dimensions and the LSH step just does the binarization.

bshethmeta:
@mdouze I will revert today.

PR #3688, opened 2024-07-25
add get_version() for c_api.

bshethmeta:
Summary: Test PR.

Differential Revision: D60207739

### DIFF
--- a/c_api/utils/utils_c.h
+++ b/c_api/utils/utils_c.h
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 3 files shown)


c_api/utils/utils_c.h
```diff
@@ -0,0 +1,27 @@
...
+#define FAISS_UTILS_C_H
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+const char* get_version();
```
Enet4, line 21: All functions in the C API have the `faiss_` prefix, otherwise they might collide with other functions (whereas the C++ library puts everything in namespaces).

Issue #3610, opened 2024-07-03
The correctness of HNSW implementation for `HNSW::upper_beam` != 1


alexanderguzhva:
# Summary

I wonder whether this (https://github.com/facebookresearch/faiss/blob/f8217045dc11f2fd8358e1f47864524fb290ae1d/faiss/impl/HNSW.cpp#L893) branch in HNSW is designed in the way as it is implemented. 

Currently, it seems that HNSW may use a beam search, which is controlled by a field `HNSW::upper_beam`. According  to the code, `upper_beam` serves not only as a beam search parameter, but as a replacement of `ef` parameter, so it is used across all HNSW levels. Logically speaking, I would expect `upper_beam` to control upper levels only, and leave `ef` for the last level. So,
```
MinimaxHeap candidates_upper_levels(upper_beam);
for (int level = max_level; level >= 1; level--) {
    // use `upper_beam`
}

// last level uses `ef`
MinimaxHeap candidates_last_level(ef);
candidates_last_level.copy_from(candidates_upper_levels);
...
```

@mdouze Could you please comment on the topic? Thanks.

asadoughi:
Closing as it was addressed in #3692.

PR #3692, opened 2024-07-25
Some small improvements.

alexanderguzhva:
A collection of small tweaks that might be needed in the future.
* Adds another statistics for HNSW called `nhops`, which is a number of different nodes visited through a single query search
* AVX-512 code for `PQDecoder8`
* Binary heap utilities for data stored in `std::pair<dis,ids>` format
* Completely remove `HNSW::upper_beam`
* AVX-512 code `HNSW::MinimaxHeap::pop_min`


mengdilin:
Hi @alexanderguzhva to help me understand the PR, do you know if there are tests that cover some of the code changes for:
- AVX-512 code for `PQDecoder8` (would it be https://github.com/facebookresearch/faiss/blob/4cfa63865f7fbf4836ff9b97a6d9495015bc39ff/tests/test_code_distance.cpp#L40 ?)
- Fix for `HNSW::upper_beam != 1` search use case
- AVX-512 code `HNSW::MinimaxHeap::pop_min` (would it be https://www.internalfb.com/code/fbsource/[38890d96d0b41924e47969458830568f5174c315]/fbcode/faiss/tests/test_hnsw.cpp?lines=49 ?)

alexanderguzhva:
Hi @mengdilin 
I did not include unit tests:
* AVX-512 code for `PQDecoder8` - this code is disabled, although I tested one in production on my end. As far as understand, the code should be tested during various `PQ` use cases among use cases. Same, there's no dedicated test code for AVX2 in my PR. 
* Fix for `HNSW::upper_beam != 1` - tbh, I don't see any `upper_beam` usage anywhere, so I am kinda waiting for comments from Matthijs (issue #3610) whether this feature is alive at all
* AVX-512 code `HNSW::MinimaxHeap::pop_min()` - same, fine on my end in production environment, and I expected this to be tested in regular AVX512 unit tests, so there's dedicated test for AVX512 in my PR.

fyi, I think that `HNSW::MinimaxHeap` requires a dedicated unit test in a separate PR (which may go before my PR), because the problem where caused a recall HUGE regression in the past, See #2771). Please let me know what you think.

mdouze:
Historically, upper_beam was used to test if it helps to have several entry points in the level-0 HNSW index. 
In my experiments, it did not improve, so we could remove the code. 
In general, I am trying to remove "near-dead" code from HNSW because there are additional levels of complexity added to the object so it's useful if we don't clutter the code with useless options.

alexanderguzhva:
@mdouze, How would you like to proceed? Should my PR completely remove `upper_beam` or would you like to have an additional PR?

alexanderguzhva:
@mdouze I've completely removed `HNSW::upper_beam` field, while preserving the format for index_load/index_save. Please let me know if this is what was expected.

mdouze:
Excellent, thanks

alexanderguzhva:
rebased to master

faiss/impl/HNSW.h
```diff
@@ -142,9 +142,6 @@ struct HNSW {
     /// enough?
     bool check_relative_distance = true;
 
-    /// number of entry points in levels > 0.
-    int upper_beam = 1;
```
mnorris11, line 146: Is it OK to just remove this and document in the next release? Is that typical for breaking changes thus far? FWIW I think I've seen some other folks setting beam in posted issues.

faiss/impl/HNSW.cpp
```diff
@@ -641,6 +650,7 @@ int search_from_candidates(
                 if (dis < threshold) {
                     if (res.add_result(dis, idx)) {
                         threshold = res.threshold;
+                        nres += 1;
```
mnorris11, line 653: This seems like a big change because it actually changes the return value, was it a bug? I don't see unit tests around the return value of this function, should we add them? Maybe it could also be a follow up.

alexanderguzhva: I think that this code fixes the bug. Please check the old baseline version of this function, https://github.com/facebookresearch/faiss/blob/691e01575e3a54c1c9094660f83415ab302d19d1/faiss/impl/HNSW.cpp#L610-L629 . It expects `nres` to grow

PR #3699, opened 2024-07-29
add test for external modules

mdouze:
Summary:
External modules should work fine as plug-ins to Faiss in the following cases:

* additional objects that can be passed in as callbacks to Faiss

* functions that use `faiss.swig_ptr` to pass in arrays.

The `swig_ptr` functionality does not always work well (also depending on the platform). Therefore this diff adds a small external swig file to test that everything works smoothly on that end.

Differential Revision: D60379753

### DIFF
--- a/faiss/python/faiss_example_external_module.swig
+++ b/faiss/python/faiss_example_external_module.swig
@@ -1,0 +1,133 @@
+
+%module faiss_example_external_module;
+
+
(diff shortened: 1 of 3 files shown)


mnorris11:
Hi @mdouze should we close this one since it is merged in https://github.com/facebookresearch/faiss/pull/3981 ?

PR #3713, opened 2024-08-01
Enable ROCm in build-only mode

ramilbakhshyiev:
Summary: Enable ROCm in CI

Differential Revision: D60598458

### DIFF
--- a/.github/actions/build_cmake/action.yml
+++ b/.github/actions/build_cmake/action.yml
@@ -12,6 +12,10 @@ inputs:
     description: 'Enable RAFT support.'
     required: false
     default: OFF
+  rocm:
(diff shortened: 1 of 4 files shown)


ramilbakhshyiev:
@iotamudelta @ItsPitt this PR enables ROCm in build-only for PRs. I have some more cleanup / simplification that I need to do but ready to start on enabling the full build when you are.

iotamudelta:
@ramilbakhshyiev we'll need the token from you to configure the ROCm test runners. Can you reach out to myself and @ItsPitt via email so that we can coordinate this securely? Emails are first name . last name @ amd.com (they should also be associated in our commits). Thanks a lot!

PR #3731, opened 2024-08-07
introduce options for reducing the overhead for a clustering procedure

alexanderguzhva:
Several changes:
1. Introduce `ClusteringParameters::check_input_data_for_NaNs`, which may suppress checks for NaN values in the input data
2. Introduce `ClusteringParameters::use_faster_subsampling`, which uses a newly added SplitMix64-based rng (`SplitMix64RandomGenerator`) and also may pick duplicate points from the original input dataset.  Surprisingly, `rand_perm()` may involve noticeable non-zero costs for certain scenarios.
3. Negative values for `ClusteringParameters::seed` initialize internal clustering rng with high-resolution clock each time, making clustering procedure to pick different subsamples each time. I've decided not to use `std::random_device` in order to avoid possible negative effects.

Useful for future `ProductResidualQuantizer` improvements.


mdouze:
Can you say a bit more about when the rand_perm is too slow? It is surprising to me that the rng could be a perf bottleneck.
I assume it's when the number of k = samples needed is << n = total number of ids. 
Because the rand_perm is O(n). 
When k << n we could sample k elements and checking with an std::set that each elements hasn't been selected before.

mdouze:
Still importing...

alexanderguzhva:
I use a dataset 1048576x768 for PRQ experiments with beam_width = 16. In this case with my new candidate code these two mentioned pieces (check for Nans and `rand_perm`) may take up to 10-15% of the total training time. For `rand_perm()`, it will be required to generate a permutation of, say, 524288 points out of 16M available.

PR #3756, opened 2024-08-14
minor refactors to move logic near to the places where it's required.

pankajsingh88:
Summary:
Minor refactors:
* move `nminus1` calculation in IVF to debug block (removes redundant calculation when `verbose` is not set)
* `decode_listno` in IVFPQ only when encoding is `by_residual`
* use ternary (`?`) to avoid if/else block.

Differential Revision: D61283444


asadoughi:
Closing as stale

PR #3760, opened 2024-08-15
Add midding hipStream SWIG typedef to fix ROCm memleak in Python

ramilbakhshyiev:
Summary: This fixes the memleak and the warning received after running Python tests under ROCm since no destructor was declared and objects would remain allocated.

Differential Revision: D61357579


ramilbakhshyiev:
@ItsPitt @iotamudelta FYI

PR #3761, opened 2024-08-15
Reorder imports in torch_test_contrib_gpu

ramilbakhshyiev:
Summary: This fixes CUDA errors inside faiss in the test environment. If torch is loaded first (this change) then both torch and faiss see all GPUs available on the machine in the ROCm build. Without this change, torch sees the GPUs and faiss does not. AMD team is looking at finding the root cause but we wanted to fix this for now.

Differential Revision: D61358018


ramilbakhshyiev:
@ItsPitt @iotamudelta FYI

PR #3785, opened 2024-08-22
fix get_compile_options bug

mengdilin:
Summary: Right now when avx512 is turned on, we will only return AVX2 in options. My understanding is turning on avx512 sets both the macros `__AVX2__` and `__AVX512F__`: https://fburl.com/vgh7jg9p

Reviewed By: asadoughi

Differential Revision: D61674490

### DIFF
--- a/faiss/utils/utils.cpp
+++ b/faiss/utils/utils.cpp
@@ -117,8 +117,9 @@ std::string get_compile_options() {
 
 #ifdef __AVX2__
     options += "AVX2 ";
-#elif __AVX512F__
+#ifdef __AVX512F__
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@mengdilin technically, according to the way this function is organized, this function is expected to return a single compilation option, not to list ones. The issue is valid, but the fix should be different

```
#ifdef __AVX512F__
    options += "AVX512 ";
#elif __AVX2__
    options += "AVX2 ";
#endif
```

mengdilin:
@alexanderguzhva addressed by https://github.com/facebookresearch/faiss/pull/3798

PR #3788, opened 2024-08-23
[IGNORE] 3526: numpy 2 hacks

asadoughi:
ran with numpy==2.1.0 on OS X
* need to understand why SWIG is not converting np.float32 as float * as it was before
* numpy/numpy/18058 mentions we should be using show_config() or perhaps show_runtime() instead of the internal api


asadoughi:
3 additional failures to debug
```
FAILED tests/test_contrib.py::TestClustering::test_ivf_train_2level - AssertionError: np.int64(55) not less than 51
```

```
FAILED tests/test_index_accuracy.py::OPQRelativeAccuracy::test_OIVFPQ - assert np.float64(0.971) >= np.float64(0.972)
```

PyTorch on arm64 SVE is not compatible with numpy 2

asadoughi:
Replaced by #4179

PR #3796, opened 2024-08-26
Use weights_only for load

kit1980:
`torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.

If `weights_only=True` doesn't work for some cases, then explicit `weights_only=False` should be used.

Found with https://github.com/pytorch-labs/torchfix/

### DIFF
--- a/demos/demo_qinco.py
+++ b/demos/demo_qinco.py
@@ -36,7 +36,7 @@ import model_qinco
 
 with torch.no_grad():
 
-    qinco = torch.load("/tmp/bigann_8x8_L2.pt", weights_only=True)
+    qinco = torch.load("/tmp/bigann_8x8_L2.pt", weights_only=False)
     qinco.eval()
     # print(qinco)
     if True:
--- a/demos/demo_qinco.py
+++ b/demos/demo_qinco.py
@@ -36,7 +36,7 @@ import model_qinco
 
 with torch.no_grad():
 
-    qinco = torch.load("/tmp/bigann_8x8_L2.pt")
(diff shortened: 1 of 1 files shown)


mdouze:
Please check if the code still works

kit1980:
No, the demo doesn't work with `weights_only=True` 

```
Traceback (most recent call last):
  File "/home/sdym/repos/facebookresearch/faiss/demos/demo_qinco.py", line 39, in <module>
    qinco = torch.load("/tmp/bigann_8x8_L2.pt", weights_only=True)
  File "/home/sdym/.conda/envs/py39/lib/python3.9/site-packages/torch/serialization.py", line 1096, in load
    raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options 
	(1) Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
	WeightsUnpickler error: Unsupported global: GLOBAL model_qinco.QINCo was not an allowed global by default. Please use `torch.serialization.add_safe_globals([QINCo])` to allowlist this global if you trust this class/function.
```

I ended up specifying `weights_only=False`.
This is still valuable as in a future PyTorch release `weights_only` will default to True, so the tests will break if `weights_only` is not set explicitly.

Ideally investigation is needed if the full pickle functionality is logically required there, and if it's possible to change the code to make `weights_only=True` work.

PR #3801, opened 2024-08-27
Fix bench_fw_codec

kuarora:
Summary: It was not training even. So, changing default to run training.

Reviewed By: mengdilin

Differential Revision: D61884150


asadoughi:
D61884150 is committed.

Issue #2996, opened 2023-08-09
Memory leak: IDSelectorBatch + SearchParameters


vaaliferov:
## Summary
I'm trying to use `IDSelectorBatch` + `SearchParameters`, but it seems like there is a memory leak somewhere.
* If I use `IDSelectorBatch` sequentually without calling  `SearchParameters` after that, then everything is okay.
* If I use `IDSelectorBatch` just once, and then call `SearchParameters` a few times, then everything fine as well.
* But if I call `IDSelectorBatch` and then `SearchParameters` sequentually a few times, then memory usage increases, and I can't find a way to free that extra memory.

## Platform
OS: Ubuntu 20.04.1
Faiss version: faiss-cpu 1.7.4

## Reproduction instructions

```python
import gc
import faiss
import numpy as np

for _ in range(10):
    subset = np.arange(0, 5000000)
    sel = faiss.IDSelectorBatch(subset)
    params = faiss.SearchParameters(sel=sel)
    mem_usage = faiss.get_mem_usage_kb() / 1024 ** 2
    print(round(mem_usage, 2), end=' ')
    gc.collect()
```

## Output
```
0.34 0.57 0.79 1.02 1.24 1.47 1.69 1.92 2.14 2.37
```

mdouze:
Thanks for the clean bug report...

mdouze:
I can repro

mdouze:
This test here is an error, since the values are not classes but instances. 
https://github.com/facebookresearch/faiss/blob/main/faiss/python/class_wrappers.py#L1084C28-L1084C28
however, it does not explain the memory leak.

mdouze:
```
subset = np.arange(0, 5000000)
sel = faiss.IDSelectorBatch(subset)
sel.this.own()    # True: correct 
params = faiss.SearchParameters(sel=sel)
sel.this.own()   # False: why???
```

mdouze:
see https://github.com/facebookresearch/faiss/pull/3007

vaaliferov:
@mdouze, thank you !

dbalabka:
@mdouze , thank you a lot! What is a proximate time when you will be able to release this fix?

dshkliarenko:
[...]

PR #3810, opened 2024-08-29
Fix IDSelector memory leak with SWIG, add unit tests (#3704)

mnorris11:
Summary:

Background
--
Issue: https://github.com/facebookresearch/faiss/issues/2996

Prior attempted fix: https://github.com/facebookresearch/faiss/pull/3007/files#diff-d704f33c46a4ef2936c8cf2a66b05f1993e25e79ee5c19d4b63c4e0cf46b0a42

Partially copied from https://github.com/facebookresearch/faiss/pull/3139 with an additional unit test.

It is a confirmed and reproducible memory leak every time. There is a workaround. See the comments on https://github.com/facebookresearch/faiss/issues/2996.

Relevant SWIG docs: https://www.swig.org/Doc4.1/SWIGDocumentation.html#Typemaps_nn2

Current status
--

`buck test faiss/tests:test_search_params -- test_ownership_2`

Test prints:

without SWIG fix:
`[0.49, 0.82, 1.15, 1.39, 1.67, 1.88, 2.16, 2.36, 2.65, 2.85, 3.13, 3.34, 3.62, 3.82, 4.11, 4.31, 4.6, 4.8, 5.08, 5.28]`

with SWIG fix:
`[0.52, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.71, 0.71, 0.71, 0.71, 0.71, 0.71, 0.71]`

Note: This test is not ideal. Ideally we could enable ASAN to catch these memory leaks. But ASAN does not seem to catch these Python memory leaks. Example:
```
def test_ownership_3(self):
        d = 32
        quantizer = faiss.IndexFlatL2(d)
        quantizer.this.own(False)
```
The above test produces no ASAN error, even though the `quantizer` object leaks.

Differential Revision: D61992599

### DIFF
--- a/faiss/python/class_wrappers.py
+++ b/faiss/python/class_wrappers.py
@@ -1170,28 +1170,6 @@ def add_to_referenced_objects(self, ref):
     else:
         self.referenced_objects.append(ref)
 
-class RememberSwigOwnership:
-    """
-    SWIG's seattr transfers ownership of SWIG wrapped objects to the class
-    (btw this seems to contradict https://www.swig.org/Doc1.3/Python.html#Python_nn22
-    31.4.2)
-    This interferes with how we manage ownership: with the referenced_objects
-    table. Therefore, we reset the thisown field in this context manager.
-    """
-
-    def __init__(self, obj):
-        self.obj = obj
-
-    def __enter__(self):
-        if hasattr(self.obj, "thisown"):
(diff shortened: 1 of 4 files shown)


mnorris11:
Looking into errors.

PR #3822, opened 2024-09-03
faster hnsw CPU index training

alexanderguzhva:
This change decreases the training time for 1M x 768 dataset down to 10 minutes from 13 minutes in our experiments.

Please verify and benchmark.


mengdilin:
I microbenchmarked `IndexHNSWFlat::add` on [google benchmark](https://github.com/facebookresearch/faiss/pull/3835) on this change and compared it against the master commit. The results are looking good (we are seeing 7-11% CPU time improvement for the microbenchmarks) with a reasonably small [p-value](https://github.com/google/benchmark/blob/main/docs/tools.md#u-test) from the U tests. 

With parameters: 
`d=64, n=1000, M=32, threads=1`, we get a 7% CPU time improvement on average (P1567300587)
`d=128, n=1000, M=16, threads=1`, we get a 7% CPU time improvement on average (P1567316788)
`d=128, n=1000, M=32, threads=5`, we get a 11% CPU time improvement on average(P1567363053)
`d=64, n=2000, M=16, threads=5`, we get a 10% CPU time improvement on average (P1567491190)
`d=128, n=1000, M=16, threads=5`, we get a 5% CPU time improvement on average (P1567371095) but this result is not statistically significant enough to be counted

I think we can verify actual impact on production with @mnorris11's work on observability for the internal customers

@kuarora and @mdouze: are there other parameters we would want to try out here?

alexanderguzhva:
@mdouze done

Issue #3845, opened 2024-09-09
add unit test for fix in PR 3840


mengdilin:
TSIA: https://github.com/facebookresearch/faiss/pull/3840#issuecomment-2338073113

PR #3849, opened 2024-09-11
add hnsw unit test for PR 3840

mengdilin:
Summary:
https://github.com/facebookresearch/faiss/issues/3845

Add unit tests for helper search utilities for HNSW. These utility functions live inside an anonymous namespace and each has a reference version gated behind a const bool, I refactored them so the reference version is a flag for the function which defaults to false.

If we are concerned about the performance overhead of the extra if branching (whether to use reference version or not) inside these utility functions, I'm happy to lift out the reference versions to their own functions inside the unit test

Differential Revision: D62510014

### DIFF
--- a/faiss/impl/HNSW.cpp
+++ b/faiss/impl/HNSW.cpp
@@ -470,105 +470,6 @@ void search_neighbors_to_add(
     vt.advance();
 }
 
-/**************************************************************
- * Searching subroutines
- **************************************************************/
-
-/// greedily update a nearest vector at a given level
-HNSWStats greedy_update_nearest(
-        const HNSW& hnsw,
-        DistanceComputer& qdis,
-        int level,
-        storage_idx_t& nearest,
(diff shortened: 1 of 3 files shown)


mengdilin:
@alexanderguzhva I followed your advise and wrote a unit test for your fix. Would appreciate a look

alexanderguzhva:
@mengdilin 
1) why is it needed to put things into `hnsw_utils` namespace?
2) are there any real-world scenarios where a switch to a reference version is really useful (besides testing)?

mengdilin:
@alexanderguzhva the utility functions were in an anonymous namespace before so I have to move them to a namespace. Happy to move it to just the faiss namespace if that's the convention.

We don't have production usecase for the referenced version. The only benefit of branching is that we are keeping the referenced version close to the source location of the code. I can move them to just unit tests similar to how we have done it for HeapPopMin's unit test

PR #3853, opened 2024-09-12
Add AVX-512 implementation for the distance and scalar quantizer functions.

mulugetam:
The distance and scalar quantizer functions currently have AVX2 implementations. This patch adds the AVX-512 equivalents for each of the AVX2 implementations. 

While preparing to push this PR, I realized that you have already implemented the AVX-512 equivalent for [HNSW::MinimaxHeap::pop_min](https://github.com/facebookresearch/faiss/blob/a166e13a25b2a5fe46adce4d7d06677d5199e598/faiss/impl/HNSW.cpp#L1176-L1265), which is great.

### DIFF
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -44,7 +44,9 @@ namespace faiss {
  * that hides the template mess.
  ********************************************************************/
 
-#ifdef __AVX2__
+#if __AVX512F__ && __F16C__
+#define USE_AVX512_F16C
+#elif __AVX2__
 #ifdef __F16C__
 #define USE_F16C
 #else
@@ -79,7 +81,17 @@ struct Codec8bit {
         return (code[i] + 0.5f) / 255.0f;
     }
 
-#ifdef __AVX2__
+#ifdef __AVX512F__
+    static FAISS_ALWAYS_INLINE __m512
+    decode_16_components(const uint8_t* code, int i) {
+        const __m128i c16 = _mm_loadu_si128((__m128i*)(code + i));
+        const __m512i i32 = _mm512_cvtepu8_epi32(c16);
+        const __m512 f16 = _mm512_cvtepi32_ps(i32);
+        const __m512 half_one_255 = _mm512_set1_ps(0.5f / 255.f);
+        const __m512 one_255 = _mm512_set1_ps(1.f / 255.f);
+        return _mm512_fmadd_ps(f16, one_255, half_one_255);
+    }
+#elif __AVX2__
     static FAISS_ALWAYS_INLINE __m256
     decode_8_components(const uint8_t* code, int i) {
         const uint64_t c8 = *(uint64_t*)(code + i);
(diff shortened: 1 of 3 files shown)


akashsha1:
@mnorris11 , @junjieqi , @naveentatikonda  - could you help add necessary reviewers to this PR? Since this is our first contribution to faiss, we're not sure who are the right set of reviewers for this change. thanks.

mengdilin:
Curious have you benchmarked this change? If so, do you have any performance numbers?

It looks like there are errors for avx512's scalar quantizer test suite: https://github.com/facebookresearch/faiss/actions/runs/10838170232/job/30084638332?pr=3853 and the change broke aarch64 compilation in https://github.com/facebookresearch/faiss/actions/runs/10838170232/job/30084638104?pr=3853

Can you try fixing those?

faiss/impl/ScalarQuantizer.cpp
```diff
@@ -121,6 +135,29 @@ struct Codec4bit {
...
+        __m128i c16 = _mm_unpacklo_epi8(c8ev_vec, c8od_vec);
+        __m256i i8lo = _mm256_cvtepu8_epi32(c16);
+        __m256i i8hi = _mm256_cvtepu8_epi32(_mm_srli_si128(c16, 8));
+        __m512i i16 = _mm512_castsi256_si512(i8lo);
+        i16 = _mm512_inserti32x8(i16, i8hi, 1);
+        __m512 f16 = _mm512_cvtepi32_ps(i16);
+        __m512 half = _mm512_set1_ps(0.5f);
+        f16 = _mm512_add_ps(f16, half);
+        __m512 one_15 = _mm512_set1_ps(1.f / 15.f);
+        return _mm512_mul_ps(f16, one_15);
```
alexanderguzhva, line 157: use `_mm512_fmadd_ps()` instead

faiss/impl/ScalarQuantizer.cpp
```diff
@@ -256,6 +293,19 @@ struct Codec6bit {
 
 #endif
 
+#ifdef __AVX512F__
+    static FAISS_ALWAYS_INLINE __m512
+    decode_16_components(const uint8_t* code, int i) {
+        __m256 v0 = decode_8_components(code, i);
```
alexanderguzhva, line 299: Please implement 16 components instead of 2x8.
Alternatively, please refer to https://github.com/zilliztech/knowhere/blob/main/thirdparty/faiss/faiss/impl/ScalarQuantizerCodec_avx512.h , which has it all implemented

faiss/impl/ScalarQuantizer.cpp
mulugetam: Fixed.

faiss/impl/ScalarQuantizer.cpp
mulugetam: Fixed.

faiss/impl/ScalarQuantizer.cpp
```diff
@@ -955,7 +1147,45 @@ struct SimilarityL2<1> {
...
+
+    FAISS_ALWAYS_INLINE float result_16() {
+        // performs better than dividing into _mm256 and adding
+        return _mm512_reduce_add_ps(accu16);
+    }
+};
+
+#endif
+
 #ifdef __AVX2__
```
mengdilin, line 1187: when __AVX512__ is defined, __AVX2__ is also present in the compiler macro. This leads to unused variable errors for Meta's CI. Can you change this to

```
#elif defined(__AVX2__)
```
to silence the error? This should mirror the `if avx512 elif avx2` logic in `get_distance_computer`

faiss/impl/ScalarQuantizer.cpp
```diff
@@ -1078,6 +1308,44 @@ struct SimilarityIP<1> {
...
```
mengdilin, line 1349: Need to do the same `#elif defined(__AVX2__)` here

mengdilin (changes requested):
Thanks for working on this! Need two minor changes to fix Meta's compilation error from unused variables.

mengdilin:
Running microbenchmarking for the distance computation across different SQ types for dimension 128 in a single-threaded environment with input size of 2000, we see a average of 40% cpu time improvement :D

faiss/impl/ScalarQuantizer.cpp
mulugetam: Fixed.

faiss/impl/ScalarQuantizer.cpp
mulugetam: Fixed.

mengdilin:
@alexanderguzhva let me know if there is any other concern with the PR. I cross-referenced the changes here with https://github.com/zilliztech/knowhere/blob/main/thirdparty/faiss/faiss/impl/ScalarQuantizerCodec_avx512.h  and they match. The changes in `faiss/utils/distances_simd.cpp` have existing tests e.g. `test_ivfpq_indexing` and `test_distances_simd`. If not, I will merge it

PR #3860, opened 2024-09-14
Add explicit instanciations of `search_dispatch_implem`

JAicewizard:
Since commit 32f0e8cf92cd2275b60364517bb1cce67aa29a55 (https://github.com/facebookresearch/faiss/pull/3190) faiss fails to build on gcc7.

For some reason gcc7 needs these explicit instanciations (or rather, for some reason later versions are OK without them). This commit partially revers that commit, adding back the explicit instanciations.

### DIFF
--- a/faiss/IndexFastScan.cpp
+++ b/faiss/IndexFastScan.cpp
@@ -549,6 +549,22 @@ void IndexFastScan::search_implem_14(
     }
 }
 
+template void IndexFastScan::search_dispatch_implem<true>(
+        idx_t n,
+        const float* x,
+        idx_t k,
+        float* distances,
+        idx_t* labels,
+        const NormTableScaler* scaler) const;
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@JAicewizard Could you please share your use case, because gcc7 is quite old? Thanks.

JAicewizard:
GCC7 might be quite old, however build systems that target old versions of glibc often use very old compilers. Many build environments that support old systems are also build on old systems, one I am working with still uses GCC4 (although it should be updated soon). Previous versions of faiss worked on GCC4 even (maybe even earlier, but that is the latest I tested).

Since this is just a small change without a big maintenance burden, I hope it would be accepted.

mdouze:
Agreed. We should accommodate as old compilers as possible.

PR #3868, opened 2024-09-17
FIx a bug for a non-simdlib code of ResidualQuantizer

alexanderguzhva:
This causes an access violation error. 

The reason why this was not caught in unit tests for AVX/NEON is that this code branch is unlikely to be used. 

The reason why this was not caught in unit tests for a plain non-SIMD binary is unclear.

More ResidualQuantizer patches to follow.


mnorris11:
Thanks for finding it!

Issue #3883, opened 2024-09-24
Installation of _swigfaiss_avx2.so seems to depend on installation of C++ headers and libraries


Janosch:
# Summary

I am compiling FAISS while building an alpine docker image. If I follow the installation steps from INSTALL.md, `swigfaiss_avx2.py` and `_swigfaiss_avx2.so` are missing in pythons `site-packages/faiss-1.8.0-py3.12.egg/faiss`.

If I install the C++ headers and libraries before installing the python package, the AVX2-enabled python module and shared library are copied to `site-packages/faiss-1.8.0-py3.12.egg/faiss` as expected.

If there is a dependency between step 3 and step 4, it should be reflected in the INSTALL.md.

# Platform

OS: alpine:3.20 linux/amd64

Faiss version: 1f42e81

Installed from: Compiled

Faiss compilation options: 
FAISS_ENABLE_GPU=OFF
BUILD_TESTING=OFF
CMAKE_BUILD_TYPE=Release
FAISS_OPT_LEVEL=avx2

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [x] C++
- [x] Python

# Reproduction instructions

1. Create docker image using this [Dockerfile](https://gist.github.com/Janosch/9be25b3537409ebd7509ca79b4685ff7)
2. `docker build --platform linux/amd64 -t missing-avx2 .`
3. `docker run -it missing-avx2 sh`
4. `ls -lah /usr/lib/python3.12/site-packages/faiss-1.8.0-py3.12.egg/faiss`
5. Verify that `swigfaiss_avx2.py` is missing

For comparison, see this [Dockerfile](https://gist.github.com/Janosch/912d06bdc79bb5d2ed9cf1a52140b89d) which reverses steps 3 and 4 from the installation instructions and therefore works.

alexanderguzhva:
@Janosch this was the default behavior all the time, as far as I remember. You compile `faiss` and then `swigfaiss`

Janosch:
@alexanderguzhva The example of the failing Dockerfile was badly constructed, my bad. Of course, you need to build `faiss` before building `swigfaiss` bindings. My point is, that if you follow the steps in the INSTALL.md you end up without AVX2 support. I created a new [Dockerfile](https://gist.github.com/Janosch/885d19caec40af3782b2397036b2c6f3) as example.

What does _not_ work:
1. Build faiss
2. Build swigfaiss
3. Install faiss

This, however, is working:
1. Build faiss
[...]

PR #3892, opened 2024-09-26
Update INSTALL.md to ensure AVX2 and AV512 support for python package

lranon1:
Following the current documentation creates the python package without AVX2 or AV512 support. Updated documentation notes that corresponding faiss version must be built before swigfaiss.

fixes facebookresearch/faiss#3883

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -177,6 +177,22 @@ $ (cd build/faiss/python && python setup.py install)
 The first command builds the python bindings for Faiss, while the second one
 generates and installs the python package.
 
+If making use of optimization options, build the correct target before swigfaiss.
+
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -177,6 +177,22 @@ $ (cd build/faiss/python && python setup.py install)
 The first command builds the python bindings for Faiss, while the second one
 generates and installs the python package.
 
+If making use of optimization options, build the correct target before swigfaiss.
```
asadoughi, line 180: Please move this to above step 3 to get the ordering correct.

asadoughi (changes requested):
Thanks!

PR #3921, opened 2024-10-07
Update Pinned RAFT version

tarang-jain:


ramilbakhshyiev:
@tarang-jain Is this PR an attempt to fix the CMake segfault with RAFT?

tarang-jain:
@ramilbakhshyiev yes this is an attempt to fix it. Not pinning the version would install 24.12 which is the latest nightly version.

ramilbakhshyiev:
@tarang-jain Questions:
1. Looks like the change fixes RAFT build issues. Is it ready to be imported?
2. Is it possible to avoid having to specific the version in 2 separate spots?

tarang-jain:
@ramilbakhshyiev yes, ready to be imported.
I don't think we can avoid specifying versions in 2 different spots. The `RAPIDS_VERSION` variable specifies which version of `rapids-cmake` to use, whereas the other one is in github actions which tells CI how to setup the conda environment. It doesn't look like the former is available to the latter.

ramilbakhshyiev:
I'm importing as we speak for the time being. However, since we pin the import to 24.06, is it possible to auto set RAPIDS_VERSION to the latest somehow?

tarang-jain:
We would have to set `RAPIDS_VERSION` to 24.06 and not the latest because we want to use `rapids-cmake` from that version. The second pin is only for github CI and would not affect building the library otherwise.

ramilbakhshyiev:
Thanks @tarang-jain for fixing this!

PR #3549, opened 2024-06-25
Migrate from RAFT to CUVS

tarang-jain:
Remove the dependency on `raft::compiled` and modify GPU implementations to use cuVS backend in place of RAFT.

A deeper insight into the dependency:
FAISS gets the ANN algorithm implementations such as IVF-Flat and IVF-PQ from cuVS. RAFT is meant to be a lightweight C++ header-only template library that cuVS relies on for the more fundamental / low-level utilities. Some examples of these are RAFT's device mdarray and mdspan objects; the RAFT resource object (`raft::resource`) that takes care of the stream ordering of device functions; linear algebra functions such as mapping, reduction, BLAS routines etc. A lot of the cuVS functions take the RAFT mdspan objects as arguments (for example `raft::device_matrix_view`). Therefore FAISS relies on both cuVS and RAFT. FAISS gets RAFT headers through cuVS and uses them to create the function arguments that can be consumed by cuVS. Note that we are not explicitly linking FAISS against `raft::raft` or `raft::compiled`. Only the required headers are included and compiled rather than compiling the whole RAFT shared library. This is the reason we still see mentions of `raft` in FAISS.

PR #3925, opened 2024-10-08
[DO NOT MERGE] Bisect CMake version that resulted in regression

KyleFromNVIDIA:
CMake maintainer here, diagnosing the CMake crash from https://github.com/facebookresearch/faiss/pull/3549. I'm going to run `git bisect` on the CMake git history to find the exact commit that caused the regression, and use this PR to run the test.


tarang-jain:
@asadoughi can we have an approval so that all checks can run for this PR?

.github/workflows/build.yml
```diff
@@ -216,8 +216,13 @@ jobs:
         env:
           ANACONDA_API_TOKEN: ${{ secrets.ANACONDA_API_TOKEN }}
         with:
+<<<<<<< HEAD
+          label: staging
+          cuvs: "ON"
+=======
```
asadoughi, line 222: This has to be fixed for CI to run

asadoughi:
Unfortunately because @KyleFromNVIDIA is a first-time contributor I have to manually click Approve on every commit. Appreciate your patience.

tarang-jain:
@asadoughi, Kyle has to test some more things. Can you again approve so that the CI runs on this PR?

asadoughi:
@KyleFromNVIDIA @tarang-jain  It looks like the segmentation faults are occurring again in our RAFT CI builds on the HEAD of main. For example [here](https://github.com/facebookresearch/faiss/actions/runs/11407220502/job/31743416914?pr=3971#step:3:2931) for a PR that is just a comment change. Do we have an ETA on when we can roll forward with cuVS & CMake? We may have to disable the RAFT CI in the interim to unblock PRs.

KyleFromNVIDIA:
What's interesting is that now that we're building CMake 3.30 from source, rather than using Conda's binary, CMake itself runs fine, but it still runs into the `__libc_single_threaded` issue. I'm wondering if there's some sort of libc mismatch between the Conda environment and the libc that's present on the system.

ramilbakhshyiev:
> What's interesting is that now that we're building CMake 3.30 from source, rather than using Conda's binary, CMake itself runs fine, but it still runs into the `__libc_single_threaded` issue. I'm wondering if there's some sort of libc mismatch between the Conda environment and the libc that's present on the system.

I will let you answer @asadoughi's question on the ETA but I wondered if you tried dropping the [symlink step](https://github.com/facebookresearch/faiss/pull/3925/files#diff-c5dee0272afe1826fd7f0936f07026a8c97865bd18ae5234490d22c4b0800800R120-R127) for cuVS/RAFT that we needed to use as a workaround? You can try dropping the whole step or parts of it.

KyleFromNVIDIA:
I've opened https://github.com/facebookresearch/faiss/pull/3973 to get myself out of first-time contributor status. If that gets merged, will I then be able to run workflows without approval? That would allow me to iterate more quickly.

Issue #2884, opened 2023-05-31
Supporting ARM SVE, the newer extended vector instruction set for aarch64


vorj:
# Summary

Dear @mdouze and all,

ARM SVE is a newer extended vector instruction set than NEON and is supported on CPUs like AWS Graviton3 and Fujitsu A64fx.
I've added SVE support and some functions implemented with SVE to faiss, then compared their execution times.
It seems that my implementation improves the performance on some environment.
This is just first implementation to show the ability of SVE, and we plan to implemnent SVE version of other functions currently not ported to SVE.

It might be unable to check on Circle CI currently, however would you mind if I submit this as PR?

# Platform

OS: Ubuntu 22.04

Faiss version: a3296f42adee7a0159b7ac09d7642e862edb142f, and mine

Installed from: compiled by myself

Faiss compilation options: `cmake -B build -DFAISS_ENABLE_GPU=OFF -DPython_EXECUTABLE=$(which python3) -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DFAISS_OPT_LEVEL=sve`
    - `-DFAISS_OPT_LEVEL=sve` is new optlevel introduced by my changes.

Running on:
- [x] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [x] Python

# Reproduction instructions

I only post the results to search SIFT1M. If you need more detailed information, please let me know.

![benchmark result](https://github.com/facebookresearch/faiss/assets/40021161/ac34630a-7470-42e0-945a-baa0bea012ff)

- Evaluated on an AWS EC2 c7g.large instance, run faiss on 
- `original` is the current (a3296f42adee7a0159b7ac09d7642e862edb142f) implementation
- `SVE` is the result of my implementation supporting ARM SVE

![image](https://github.com/facebookresearch/faiss/assets/40021161/8edfc585-a4dd-4a43-93f5-217c9418e9ba)

The above image illustrates the ratio of speed up.

- In the best case, `SVE` is approx. 2.26x faster than `original` (IndexIVFPQ + IndexHNSWFlat, M: 32  nprove: 16)
    - `original` : 0.618 ms
    - `SVE` : 0.274 ms

mdouze:
Thanks for looking into this! 
Do I understand correctly that this is with a 512-bit SIMD width? 
Indeed we should have a way to integrate code for hardware that is not supported by CircleCI (AVX512 being the other example). 
So we welcome a PR for this functionality.

[...]

PR #3933, opened 2024-10-09
Add some SVE implementations

vorj:
related: #2884 

I added some SVE implementations of:

- `code_distance`
    - `distance_single_code`
    - `distance_four_codes`
- `exhaustive_L2sqr_blas_cmax_sve`
- `fvec_inner_products_ny`
- `fvec_madd`

## Evaluation result

I evaluated the search for SIFT1M dataset on AWS EC2 c7g.large and r8g.large instances.
`main` is the current (2e6551ffa3f6fbdb1ba814c2c531fb399b00d4e3) implementation.

### c7g.large (Graviton 3)

g3_sift1m

g3_ivfpq

On Graviton 3, `IndexIVFPQ` has been improved particularly. In the best case (IndexIVFPQ + IndexFlatL2, M: 32), this PR is approx. 2.38-~~2.50~~**2.44**x faster than `main` .

- nprobe: 1, 0.069ms/query → 0.029ms/query
- nprobe: 4, 0.181ms/query → ~~0.074~~**0.075**ms/query
- nprobe: 16, 0.613ms/query → ~~0.245~~**0.251**ms/query

### r8g.large (Graviton 4)

g4_sift1m

g4_ivfpq

On Graviton 4, especially `IndexIVFPQ` for tiny `nprobe` has been improved. In the best case (IndexIVFPQ + IndexFlatL2, M: 8, nprobe: 1), this PR is approx. 1.33x faster than `main` (0.016ms/query → 0.012ms/query).

### DIFF
--- a/faiss/impl/code_distance/code_distance-sve.h
+++ b/faiss/impl/code_distance/code_distance-sve.h
@@ -1,0 +1,356 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#ifdef __ARM_FEATURE_SVE
+
+#include <arm_sve.h>
+
+#include <tuple>
+#include <type_traits>
+
+#include <faiss/impl/code_distance/code_distance-generic.h>
+
+namespace faiss {
+
+template <typename PQDecoderT>
+std::enable_if_t<!std::is_same_v<PQDecoderT, PQDecoder8>, float> inline distance_single_code_sve(
+        // the product quantizer
+        const size_t M,
+        // number of bits per quantization index
+        const size_t nbits,
+        // precomputed distances, layout (M, ksub)
+        const float* sim_table,
+        const uint8_t* code) {
+    // default implementation
+    return distance_single_code_generic<PQDecoderT>(M, nbits, sim_table, code);
+}
+
+static inline void distance_codes_kernel(
+        svbool_t pg,
+        svuint32_t idx1,
+        svuint32_t offsets_0,
+        const float* tab,
+        svfloat32_t& partialSum) {
+    // add offset
+    const auto indices_to_read_from = svadd_u32_x(pg, idx1, offsets_0);
+
+    // gather values, similar to some operations of tab[index]
+    const auto collected =
+            svld1_gather_u32index_f32(pg, tab, indices_to_read_from);
+
+    // collect partial sum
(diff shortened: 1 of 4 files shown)


vorj:
Eww, well, usually I'm developing SVE version of faiss on Graviton 3, and the tests pass on that so I'd missed the test failure on Graviton 4. I fixed it so current CI will pass. I will update a performance report in PR description later.

alexanderguzhva:
@vorj lgtm. Two comments:
1. would you like to consider using `svprfw_gather()` as well?
2. x86 implementation contains two special cases for `M=4` and `M=8`. This is because ones are very typical. Would you like to consider adding these two as well? (M=16 as well, maybe)

vorj:
@alexanderguzhva 

1. Actually, I will lose the AWS account currently I use on Friday, Oct 11th, at 10 am UTC since I'm leaving my current position. I will try to implement it, but it may not make in time. I think that finding an effective prefetching needs some attempts.
2. 
    - Even current implementation runs `distance_codes_kernel` minimally [using `break`](https://github.com/facebookresearch/faiss/pull/3933/files#diff-929a8ac6000671698b7dbcb143e6243ef0ff097d9e7983a2d789f3d226c37028R251):
        - on the CPU with 128bit SVE register,
            - `M=4` : once
            - `M=8` : twice
            - `M=16` : four times
        - on the CPU with 256bit SVE register
            - `M=4` : once
            - `M=8` : once
            - `M=16` : twice
    - And when we use NEON for this, we can't use vector load.
        - note: ARM CPUs are not necessarily little endian, so we shouldn't make a vector register like `vreinterpretq_u8_u32(vdupq_n_u32(*reinterpret_cast<uint32_t*>(code)))` but can make it using some scalar codes.
    - Of course I understand that `M` is not so big for common `distance_code` usage, and specialized implementation will be make it faster. I will also try to investigation the performance of specialized implementation, but sorry if I can't make it in time.

vorj:
@alexanderguzhva

> Maybe, neon is sufficient for these

I'd also forgotten last night that, but neon doesn't have gather/scatter.
Thus even if we will implement specialized `code_distance` for `M=4/8(/16)` , we still need to use SVE.

vorj:
@alexanderguzhva I've implemented specialized one for `M <= svcntw()` . On 128bit SVE environment, specialized one runs when `M=4` . On 256bit SVE environment, specialized one runs when `M=4/8` . It seems better performance than generic one, approx. 2x for Graviton 3 and approx. 1.3x for Graviton 4 where `M=4` .

alexanderguzhva:
@vorj lgtm. Thanks for your SVE implementation!

PR #3962, opened 2024-10-16
Removing Manual Hipify Build Step

ItsPitt:
- Called the hipify script at CMAKE configure time removing the need for the user to run it.
- Now removes any .hip files left over when running the hipify script.
- Cleaned up the hipify script to remove redundancy.


mnorris11:
For some reason `Build / Linux x86_64 GPU w/ RAFT` is failing both on this PR and some others due to a segfault in python tests. Strange that it did not fail during the nightly build though. Looking into it

PR #3970, opened 2024-10-17
Add VectorTransform read from filename to the C API

makosten:
This adds read_VectorTransform to the C API. This is helpful for independently loading a vector transform rather than as an IndexPreTransform.

### DIFF
--- a/c_api/index_io_c.cpp
+++ b/c_api/index_io_c.cpp
@@ -15,6 +15,7 @@
 
 using faiss::Index;
 using faiss::IndexBinary;
+using faiss::VectorTransform;
 
 int faiss_write_index(const FaissIndex* idx, FILE* f) {
     try {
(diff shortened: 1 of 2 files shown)


Enet4 (approved):
Yup, that seems to be in order.

Enet4 (approved):

c_api/index_io_c.cpp
```diff
@@ -83,3 +84,13 @@ int faiss_read_index_binary_fname(
     }
     CATCH_AND_HANDLE
 }
+
+int faiss_read_VectorTransform_fname(
```
junjieqi, line 88: Could we make the change to the function name to all lower cases like faiss_read_vector_transform()

c_api/index_io_c.h
```diff
@@ -71,6 +72,13 @@ int faiss_read_index_binary_fname(
         const char* fname,
         int io_flags,
         FaissIndexBinary** p_out);
+
+/** Read vector transform from a file.
+ * This is equivalent to `faiss:read_VectorTransform` when a file path is given.
+ */
+int faiss_read_VectorTransform_fname(
```
junjieqi, line 79: ditto

junjieqi (commented):
In addition, could you provide an unit test for it?

c_api/index_io_c.cpp
Enet4: This is merely following the convention already applied for the rest of the C API. As the underlying object being created is of type `VectorTransform`, the function signature retains that name to let users understand what it maps to in the C++ API.
IMO it would be too disruptive to start changing this convention now.

I also don't think we have any unit tests for the C API at the moment, so we'd need to kickstart that.

PR #3982, opened 2024-10-23
Prefilter with attributes before Search

VinhASOKEN:
A search with a prefilter on one or two attributes of type float, within the range [lower, upper], before executing the main search.

```python
d = 512
nb = 50000
nq = 3

xb = np.random.rand(nb, d).astype(np.float32)
ids = np.linspace(1, nb, nb).astype(np.int64)
xtime = np.linspace(1, nb, nb).reshape(nb, 1).astype(np.float32)
xq = np.random.rand(nq, d).astype(np.float32)

index = faiss.index_factory(d, "IVF1024,SQfp16")
index.set_is_include_one_attribute()
print("USING ONE ATTRIBUTE IVF1024, SQfp16 ? :", index.get_is_include_one_attribute())

index.train(xb)
index.add_with_ids_with_one_attribute(xb, xtime, ids)

diss, ids, times = index.search_with_one_attribute(xq, 1000, 10000, 5)
```


bshethmeta:
@VinhASOKEN Have you looked at the IDSelector API? Might be doing similar things

VinhASOKEN:
> @VinhASOKEN Have you looked at the IDSelector API? Might be doing similar things

VinhASOKEN:
@bshethmeta I have looked into it, and IDSelector can help you search by the IDs you have selected by checking which IDs meet the conditions. Here, I am working on allowing the indexes to add new attributes, and you can also prefilter (if desired) based on those new attributes before starting the search. Typically, Faiss only helps us with searching, and then we apply the filter afterward. It is very likely that the top-K results may not include those that fall within the range we expect.

VinhASOKEN:
> @VinhASOKEN Have you looked at the IDSelector API? Might be doing similar things

@bshethmeta I have looked into it, and IDSelector can help you search by the IDs you have selected by checking which IDs meet the conditions. Here, I am working on allowing the indexes to add new attributes, and you can also prefilter (if desired) based on those new attributes before starting the search. Typically, Faiss only helps us with searching, and then we apply the filter afterward. It is very likely that the top-K results may not include those that fall within the range we expect.

asadoughi:
Thanks for submitting the PR for consideration.

IDSelector can, in fact, be used for pre-filtering, but it does require the user to maintain their own metadata index outside of the vector index. 

For any API change, we'll need to deliberate the interface design options and their tradeoffs. In this case, for how users could combine the metadata and vector indexes into a single index - it's not evident to me the API proposed is the best approach without considering alternatives.

For now, I would consider not moving forward with this approach and revisiting this concept in the future. Feel free to open a Github issue to facilitate the design discussion in the meantime.

PR #3988, opened 2024-10-23
add all wrapped indexes to the index_read

mnorris11:
Summary: These were not added in the refactoring, so adding them now.

Differential Revision: D64702824


asadoughi:
Merged in efa18da

Issue #3862, opened 2024-09-16
bfloat16 (bf16) support in faiss


mlomeli1:
Many LLMs are trained with bf16, if we want to use the hidden states of LLMs for retrieval, those vectors will be in bf16 dtype. It would be helpful to support bf16 in Faiss so that we can use LLMs as retriever or embedding model.

mdouze:
Note that we cannot pass through the numpy wrapper because numpy does not support bf16. 
Adapting `gpu_knn` code for pytorch should be easy 
https://github.com/facebookresearch/faiss/blob/dc55e11874e4e32d0f04a9d62156ebe2ca25192d/contrib/torch_utils.py#L497

mdouze:
+ add bfloat16 support here 

https://github.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuDistance.h#L76

mlomeli1:
Example of how we currently would use a PQ codec to encode/decode pytorch bf16 tensors:
````
torch.from_numpy( codec.sa_decode(codec.sa_encode(x.to(device='cpu', dtype=torch.float32).numpy()) )
````
this piece of code showcases all the cpu moves + up casting + converting to numpy array. Successively, at decoding, we need to convert back to a tensor. Ideally, we could avoid some of these if this was supported for pytorch tensors.

alexanderguzhva:
just in case, there is a ScalarQuantizer implementation for bf16, maybe portions of it can be reused

PR #4014, opened 2024-11-05
Faiss GPU: bfloat16 brute-force kNN support

mdouze:
Summary:
This diff adds support for bfloat16 vector/query data types with the GPU brute-force k-nearest neighbor function (`bfKnn`).

The change is largely just plumbing the new data type through the template hierarchy (so distances can be computed in bfloat16).

Of note, by design, all final distance results are produced in float32 regardless of input data type (float32, float16, bfloat16). This is because the true nearest neighbors in many data sets can often differ by only ~1000 float32 ULPs in terms of distance which will result in possible false equivalency. This seems to be one area where lossy compression/quantization thoughout does not work as well (and is also why `CUBLAS_MATH_DISALLOW_REDUCED_PRECISION_REDUCTION` is set in `StandardGpuResources.cpp`. However, given that there is native bf16 x bf16 = fp32 tensor core support on Ampere+ architectures, the matrix multiplication itself should 

WARNING: The one thing this diff does not yet handle properly is header inclusion / compilation for GPUs older than Ampere. This will need to be fixed before landing (so that compiling with an older CUDA SDK or compiling for the Volta architecture will simply error out at runtime properly with lack of support, instead of failing to compile (?)

Differential Revision: D65459723

### DIFF
--- a/faiss/gpu/GpuDistance.cu
+++ b/faiss/gpu/GpuDistance.cu
@@ -242,7 +242,7 @@ void bfKnn(GpuResourcesProvider* prov, const GpuDistanceParams& args) {
     FAISS_THROW_IF_NOT_MSG(
             args.vectorType == args.queryType,
             "limitation: both vectorType and queryType must currently "
-            "be the same (F32 or F16");
+            "be the same (F32 / F16 / BF16");
 
 #if defined USE_NVIDIA_RAFT
     // Note: For now, RAFT bfknn requires queries and vectors to be same layout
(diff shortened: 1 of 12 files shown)


mdouze:
Related to  https://github.com/facebookresearch/faiss/issues/3862

wickedfoo:
> For ROCm support, `hip/hip_bf16.h` needs to be included. This'll provide the hipified `__hip_bfloat16` data type and operations as per [amd_hip_bf16.h](https://github.com/ROCm/clr/blob/amd-staging/hipamd/include/hip/amd_detail/amd_hip_bf16.h) as well as packed `__hip_bfloat162`. Hope this helps.

I thought referring to HIP specific datatypes was the entire point of HIP cross-compilation? I see coverage for functions/CUDA types in https://rocm.docs.amd.com/projects/HIPIFY/en/latest/tables/CUDA_Device_API_supported_by_HIP.html but why is there no support for `__nv_bfloat16` / `__nv_bfloat162`? (FYI that documentation indicates that there is no analogue for `__nv_bfloat162` but your `amd_hip_bf16.h` header does define one).

It's kind of annoying to have to `#ifdef` HIP cases just for this datatype.

PR #3266, opened 2024-02-27
Enable Intel®-AMX/oneDNN to accelerate IndexFlatIP search

guangzegu:
### Description
Intel® AMX, which is an AI acceleration engine deeply embedded into every core of our 4th/5th Gen Intel® Xeon® Scalable processor.

### How to use
When invoking Cmake ,  add an option as follows:

- `-DFAISS_ENABLE_DNNL=OFF` Enable support for oneDNN to accelerate IndexFlatIP search(possible values are `ON` and `OFF`)

When you want to use Intel®-AMX/oneDNN to accelerate the search of indexFlatIP, set `FAISS_ENABLE_DNNL` to ON and run on 4th/5th Gen Intel® Xeon® Scalable processor, the exhaustive_inner_product_seq method will be accelerated.

PR #4020, opened 2024-11-06
Use _mm512_popcnt_epi64 to speedup hamming distance evaluation.

mulugetam:
The `_mm512_popcnt_epi64` intrinsic is used to accelerate Hamming distance calculations in `HammingComputerDefault` and `HammingComputer64`. 

Benchmarking with [bench_hamming_computer](https://github.com/facebookresearch/faiss/blob/main/benchs/bench_hamming_computer.cpp) on AWS [r7i](https://aws.amazon.com/ec2/instance-types/r7i/) instance shows a performance improvement of up to 30% compared to AVX-2.

This PR depends on [PR#4025](https://github.com/facebookresearch/faiss/pull/4025)

### DIFF
--- a/faiss/utils/hamming_distance/avx512-inl.h
+++ b/faiss/utils/hamming_distance/avx512-inl.h
@@ -9,6 +9,11 @@
 #define HAMMING_AVX512_INL_H
 
 // AVX512 version
+// The _mm512_popcnt_epi64 intrinsic is used to accelerate Hamming distance
+// calculations in HammingComputerDefault and HammingComputer64. This intrinsic
+// is not available in the default FAISS avx512 build mode but is only
+// available in the avx512_spr build mode, which targets Intel(R) Sapphire
+// Rapids.
 
 #include <cassert>
 #include <cstddef>
--- a/faiss/utils/hamming_distance/avx512-inl.h
+++ b/faiss/utils/hamming_distance/avx512-inl.h
@@ -1,0 +1,485 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#ifndef HAMMING_AVX512_INL_H
+#define HAMMING_AVX512_INL_H
+
+// AVX512 version
+
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+
+#include <faiss/impl/platform_macros.h>
+
+#include <immintrin.h>
+
+namespace faiss {
+
+/* Elementary Hamming distance computation: unoptimized  */
+template <size_t nbits, typename T>
+inline T hamming(const uint8_t* bs1, const uint8_t* bs2) {
+    const size_t nbytes = nbits / 8;
+    size_t i;
+    T h = 0;
+    for (i = 0; i < nbytes; i++) {
+        h += (T)hamdis_tab_ham_bytes[bs1[i] ^ bs2[i]];
+    }
+    return h;
+}
+
+/* Hamming distances for multiples of 64 bits */
+template <size_t nbits>
+inline hamdis_t hamming(const uint64_t* bs1, const uint64_t* bs2) {
+    const size_t nwords = nbits / 64;
+    size_t i;
+    hamdis_t h = 0;
+    for (i = 0; i < nwords; i++) {
+        h += popcount64(bs1[i] ^ bs2[i]);
+    }
+    return h;
+}
+
+/* specialized (optimized) functions */
+template <>
+inline hamdis_t hamming<64>(const uint64_t* pa, const uint64_t* pb) {
+    return popcount64(pa[0] ^ pb[0]);
+}
+
+template <>
+inline hamdis_t hamming<128>(const uint64_t* pa, const uint64_t* pb) {
+    return popcount64(pa[0] ^ pb[0]) + popcount64(pa[1] ^ pb[1]);
+}
+
+template <>
+inline hamdis_t hamming<256>(const uint64_t* pa, const uint64_t* pb) {
+    return popcount64(pa[0] ^ pb[0]) + popcount64(pa[1] ^ pb[1]) +
+            popcount64(pa[2] ^ pb[2]) + popcount64(pa[3] ^ pb[3]);
+}
+
+/* Hamming distances for multiple of 64 bits */
+inline hamdis_t hamming(
+        const uint64_t* bs1,
+        const uint64_t* bs2,
+        size_t nwords) {
+    hamdis_t h = 0;
+    for (size_t i = 0; i < nwords; i++) {
+        h += popcount64(bs1[i] ^ bs2[i]);
+    }
+    return h;
+}
+
+/******************************************************************
+ * The HammingComputer series of classes compares a single code of
+ * size 4 to 32 to incoming codes. They are intended for use as a
+ * template class where it would be inefficient to switch on the code
+ * size in the inner loop. Hopefully the compiler will inline the
+ * hamming() functions and put the a0, a1, ... in registers.
+ ******************************************************************/
+
+struct HammingComputer4 {
+    uint32_t a0;
+
+    HammingComputer4() {}
+
+    HammingComputer4(const uint8_t* a, int code_size) {
+        set(a, code_size);
+    }
+
+    void set(const uint8_t* a, int code_size) {
+        assert(code_size == 4);
+        a0 = *(uint32_t*)a;
+    }
+
+    inline int hamming(const uint8_t* b) const {
+        return popcount64(*(uint32_t*)b ^ a0);
+    }
+
+    inline static constexpr int get_code_size() {
+        return 4;
+    }
+};
+
+struct HammingComputer8 {
+    uint64_t a0;
+
+    HammingComputer8() {}
+
+    HammingComputer8(const uint8_t* a, int code_size) {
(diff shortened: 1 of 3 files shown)


alexanderguzhva:
@mulugetam Sure, `_mm512_popcnt_epi64()` requires VPOPCNTDQ (at least Intel IceLake CPUs), but Faiss AVX512 for Linux builds defaults to F, CD, VL, DQ, BW extensions (https://github.com/facebookresearch/faiss/blob/main/faiss/CMakeLists.txt). So, the performance gain from this PR may be seen only if someone builds Faiss by himself using an MSVC compiler.

mengdilin:
@mulugetam FYI we are also working with other Intel collaborators (@guangzegu and @xtangxtang) to incorporate AMX to FAISS from https://github.com/facebookresearch/faiss/pull/3266 @asadoughi suggested perhaps we should have a meeting and sync up on the collaboration with Intel

mengdilin:
Thanks for your contributions :) Question: What is the performance for Hamming distance calculations when `VPOPCNTDQ` is not supported compared to AVX2?

mulugetam:
@alexanderguzhva 
> VPOPCNTDQ

Any downside in using the scheme below? I check if the compiler and the underlying machine support VPOPCNTDQ and then add the `-mavx512vpopcntdq` flag to the `target_compile_options` based on that.

```
function(check_avx512vpopcntdq_support result_var)
    set(TEST_PROGRAM "${CMAKE_BINARY_DIR}/check_avx512vpopcntdq.cpp")

    file(WRITE ${TEST_PROGRAM} "
    int main() {
        #ifdef __has_builtin
            #if __has_builtin(__builtin_cpu_supports)
                if (__builtin_cpu_supports(\"avx512vpopcntdq\")) {
                    return 0;
                }
            #endif
        #endif
        return 1;
    }
    ")

    try_run(RUN_RESULT COMPILE_RESULT
            ${CMAKE_BINARY_DIR}
            ${TEST_PROGRAM})

    if(RUN_RESULT EQUAL 0)
        set(${result_var} TRUE PARENT_SCOPE)
    else()
        set(${result_var} FALSE PARENT_SCOPE)
    endif()
endfunction()

include(CheckCXXCompilerFlag)
check_cxx_compiler_flag("-mavx512vpopcntdq" COMPILER_SUPPORTS_AVX512VPOPCNTDQ)

if(COMPILER_SUPPORTS_AVX512VPOPCNTDQ)
        check_avx512vpopcntdq_support(AVX512VPOPCNTDQ_SUPPORT)
        add_definitions(-DHAS_AVX512VPOPCNTDQ_SUPPORT)
endif()

// Line 259 gets modified
if(HAS_AVX512VPOPCNTDQ_SUPPORT)
    target_compile_options(faiss_avx512 PRIVATE
        $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mpopcnt -mavx512vpopcntdq>)
else()
    target_compile_options(faiss_avx512 PRIVATE
        $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mpopcnt>)
endif()

```

mulugetam:
@alexanderguzhva

VPOPCNTDQ

Any downside in using the scheme below? I check if the compiler and the underlying machine support VPOPCNTDQ and then add the -mavx512vpopcntdq flag to the target_compile_options based on that.

```
function(check_avx512vpopcntdq_support result_var)
    set(TEST_PROGRAM "${CMAKE_BINARY_DIR}/check_avx512vpopcntdq.cpp")

    file(WRITE ${TEST_PROGRAM} "
    int main() {
        #ifdef __has_builtin
            #if __has_builtin(__builtin_cpu_supports)
                if (__builtin_cpu_supports(\"avx512vpopcntdq\")) {
                    return 0;
                }
            #endif
        #endif
        return 1;
    }
    ")

    try_run(RUN_RESULT COMPILE_RESULT
            ${CMAKE_BINARY_DIR}
            ${TEST_PROGRAM})

    if(RUN_RESULT EQUAL 0)
        set(${result_var} TRUE PARENT_SCOPE)
    else()
        set(${result_var} FALSE PARENT_SCOPE)
    endif()
endfunction()

include(CheckCXXCompilerFlag)
check_cxx_compiler_flag("-mavx512vpopcntdq" COMPILER_SUPPORTS_AVX512VPOPCNTDQ)

if(COMPILER_SUPPORTS_AVX512VPOPCNTDQ)
        check_avx512vpopcntdq_support(AVX512VPOPCNTDQ_SUPPORT)
        add_definitions(-DHAS_AVX512VPOPCNTDQ_SUPPORT)
endif()

// Line 259 gets modified
if(HAS_AVX512VPOPCNTDQ_SUPPORT)
    target_compile_options(faiss_avx512 PRIVATE
        $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mpopcnt -mavx512vpopcntdq>)
else()
    target_compile_options(faiss_avx512 PRIVATE
        $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mpopcnt>)
endif()

```

mulugetam:
> Thanks for your contributions :) Question: What is the performance for Hamming distance calculations when `VPOPCNTDQ` is not supported compared to AVX2?

@mengdilin If I benchmark the AVX-512 build without VPOPCNTDQ and compare it to the AVX2 build, I’d see about a 1% - 5% speedup (depending on the code_size) for AVX-512.

mulugetam:
> @mulugetam FYI we are also working with other Intel collaborators (@guangzegu and @xtangxtang) to incorporate AMX to FAISS from #3266 @asadoughi suggested perhaps we should have a meeting and sync up on the collaboration with Intel

@mengdilin 
That's nice. I'm approaching this from an [OpenSearch](https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/) perspective, where FAISS is used as one of the engines. We've observed significant performance improvements in vector indexing and search with our [previous PR](https://github.com/facebookresearch/faiss/pull/3853), which added AVX-512 support for ScalarQuantizer.

alexanderguzhva:
@mulugetam Technically, cmake-based check for vpopcntdq may lead to the following situation. Say, a build is performed on a most recent CPU get, and then the built python package is put into conda / pip repository. This package gets spread across the world and leads to problems for all ppl who has Intel Cascade Lake CPUs.
More preferrable solution at this stage would be to add a new compilation mode (there are 'generic', 'avx2' and 'avx512'), it is up to Faiss team to decide whether to accept another arch.
Also, it is possible to have dynamic kernel selection via function pointers, but the kernel is tiny (function pointer overhead will be huge), and Faiss does not currently support such an approach.

alexanderguzhva:
@mulugetam `FP16` means `avx512fp16` or `f16c`?

mulugetam:
@alexanderguzhva avx512fp16. 

https://networkbuilders.intel.com/solutionslibrary/intel-avx-512-fp16-instruction-set-for-intel-xeon-processor-based-products-technology-guide

alexanderguzhva:
@mulugetam Well, there are two more problems then.
1. fp16 is not supported on AMD, so this means that the proposed arch should be called something like `avx512-sr`. Which is beneficial in case if FAISS finally gets AMX support, bcz AMX also requires SR as a minimum.
2. fp16 might lack the needed precision range. As of now, FP16 version of Faiss `ScalarQuantizer` assumes that the data is stored in fp16, but computations are performed in fp32. Performing pure computations in fp16 (say, via `_mm512_fmadd_ph()`) might produce out-of-range problems for the final result. So, I'd recommend adding some specific flag to a `ScalarQuantizer` class that lets it use pure fp16-based computations.

Nevertheless, please write and benchmark the code, and then we'll decide how to integrate it into Faiss properly. :) Thanks

mulugetam:
@alexanderguzhva @mengdilin I have created new [PR#4025](https://github.com/facebookresearch/faiss/pull/4025) that adds 'avx512-sr' architecture mode and marked this PR to be dependent on it.

mengdilin:
@mulugetam the dependent PR was just merged. Do you mind rebasing this PR on top to kick off the new CI?

mulugetam:
Thank you @mengdilin

faiss/utils/hamming_distance/avx512-inl.h
```diff
@@ -0,0 +1,485 @@
...
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#ifndef HAMMING_AVX512_INL_H
+#define HAMMING_AVX512_INL_H
+
+// AVX512 version
```
mengdilin, line 11: Can you add a comment here summarizing the discussions of the two PRs:
```
The _mm512_popcnt_epi64 intrinsic is used to accelerate Hamming distance calculations in HammingComputerDefault and HammingComputer64. This is not available in default FAISS avx512 built mode and is only available on Intel Sapphire Rapids
```

mengdilin (approved):

faiss/utils/hamming_distance/avx512-inl.h
mulugetam: Added.

PR #4021, opened 2024-11-07
Upgrade CUVS to 24.12

tarang-jain:


tarang-jain:
@asadoughi @cjnolet after a lot of trial and error there are always two conflicts with cuda-12.4:
1. Using the `nvidia` channel for installing CUDA related deps for cuda versions before 12.5 is not advisable. The most accurate conda pkgs pre cuda12.4 are in the conda-forge channel
2. Using `conda-forge` to install those deps creates some mixup between the `defaults` channel and `conda-forge`. Furthermore, the CPU check itself does not pass currently if I remove the defaults channel and use `conda-forge` everywhere.
cuda12.5 onwards the nvidia channel is updated with the most accurate cuda related pkgs and can be used instead of conda-forge. So we will have to wait for faiss to start using cuda12.5 before merging this PR in.

PR #4022, opened 2024-11-08
add range_search() to IndexRefine

alexanderguzhva:
This is very convenient to have `range_seach()` in `IndexRefine`. Unlike the plain `search()` method, `range_search()` just reevaluates the computed distances from the baseline index. The labels are not re-sorted according to new distances, because this is not listed as a requirement in a method description 
https://github.com/facebookresearch/faiss/blob/adb188411a98c3af5b7295c7016e5f46fee9eb07/faiss/Index.h#L150-L161 
https://github.com/facebookresearch/faiss/blob/adb188411a98c3af5b7295c7016e5f46fee9eb07/faiss/impl/AuxIndexStructures.h#L35

### DIFF
--- a/faiss/IndexRefine.cpp
+++ b/faiss/IndexRefine.cpp
@@ -166,6 +166,45 @@ void IndexRefine::search(
     }
 }
 
+void IndexRefine::range_search(
+        idx_t n,
+        const float* x,
+        float radius,
+        RangeSearchResult* result,
+        const SearchParameters* params_in) const {
+    const IndexRefineSearchParameters* params = nullptr;
+    if (params_in) {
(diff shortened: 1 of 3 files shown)


gtwang01:
The unsuccessful check doesn't seem to be caused by this PR.
Can you add some unit tests to ensure reliability?

alexanderguzhva:
@gtwang01 will do

gtwang01:
@alexanderguzhva Following up on this

alexanderguzhva:
@gtwang01 I've added a unit test. Please let me know if it is sufficient. Thanks.

gtwang01 (approved):
Thanks for working on this

PR #4025, opened 2024-11-12
Add a new architecture mode: 'avx512_spr'.

mulugetam:
This PR adds a new architecture mode to support the new extensions to AVX512, namely [AVX512-FP16](https://networkbuilders.intel.com/solutionslibrary/intel-avx-512-fp16-instruction-set-for-intel-xeon-processor-based-products-technology-guide), which have been available since Intel® Sapphire Rapids.

This PR is a prerequisite for [PR#4020](https://github.com/facebookresearch/faiss/pull/4020) that speeds up hamming distance evaluations.


mengdilin:
Hmm weird the CIs are not running on the PR. Do you mind pushing a new commit and see if the CIs start?

mulugetam:
@mengdilin Pushed a new commit but CI not starting. Is this possibly because I updated .github/workflows/build.yml?

mengdilin:
Yea there is a syntax error in the build file: see https://github.com/facebookresearch/faiss/actions/runs/11804384709

.github/workflows/build.yml
```diff
@@ -67,6 +67,17 @@ jobs:
         uses: ./.github/actions/build_cmake
         with:
           opt_level: avx512
+  linux-x86_64-AVX512-cmake:
```
mengdilin, line 70: linux-x86_64-AVX512-advanced-cmake

mulugetam:
@mengdilin CI is picking g++ version 11. But AVX512-FP16 (-mavx512fp16) requires version 12+.

conda/faiss/build-lib.sh
```diff
@@ -7,18 +7,18 @@
...
-      -DFAISS_OPT_LEVEL=avx512 \
+      -DFAISS_OPT_LEVEL=avx512-sr \
       -DFAISS_ENABLE_GPU=OFF \
       -DFAISS_ENABLE_PYTHON=OFF \
       -DBLA_VENDOR=Intel10_64lp \
       -DCMAKE_INSTALL_LIBDIR=lib \
       -DCMAKE_BUILD_TYPE=Release .
 
-make -C _build -j$(nproc) faiss faiss_avx2 faiss_avx512
+make -C _build -j$(nproc) faiss faiss_avx2 faiss_avx512 faiss_avx512_sr
```
mengdilin, line 21: This is used for faiss's conda packaging upload. I don't think we want to expose this build mode yet in conda officially. Can you omit this for now?

mengdilin:
Ah yea the conda publication CI is using the older compiler version. We should investigate on our side; however, I don't think we want to publish this architecture mode to conda right now, can you omit it?

Looks like CI failure is coming from a unit test failure from https://github.com/facebookresearch/faiss/blob/0fb56d94b7f409cd47f970b6e721deb1a837d8aa/tests/test_contrib.py#L552

Try commenting out that test and see if anything else fails?
I'm going on PTO, handing this over to the next performance oncall @kuarora

conda/faiss/build-lib.sh
mulugetam: Removed.

mulugetam:
Thanks @mengdilin. @kuarora Could you please review?

alexanderguzhva:
@mulugetam I would use `-march=sapphirerapids -mtune=sapphirerapids` for the compiler flags, because SR supports many other AVX512 instruction extensions that are not currently listed among compiler flags

faiss/CMakeLists.txt
```diff
@@ -263,6 +263,20 @@ else()
   add_compile_options(/bigobj)
 endif()
 
+add_library(faiss_avx512_sr ${FAISS_SRC})
+if(NOT FAISS_OPT_LEVEL STREQUAL "avx512-sr")
+  set_target_properties(faiss_avx512_sr PROPERTIES EXCLUDE_FROM_ALL TRUE)
+endif()
+if(NOT WIN32)
+  # Architecture mode to support AVX512 extensions available since Intel(R) Sapphire Rapids.
+  # Ref: https://networkbuilders.intel.com/solutionslibrary/intel-avx-512-fp16-instruction-set-for-intel-xeon-processor-based-products-technology-guide
```
mengdilin, line 272: thanks for the ref!

tests/test_contrib.py
```diff
@@ -568,7 +569,7 @@ def test_ivf_train_2level(self):
         # normally 47 / 200 differences
         ndiff = (Iref != Inew).sum()
         self.assertLess(ndiff, 51)
```
mengdilin, line 571: is there away to relax the threshold such that this test passes in SR mode as well?

mengdilin (commented):
Back from PTO. LGTM, can you resolve the conflict so I can import the PR. For the unit test, can you relax the threshold instead of commenting it out? If somehow you cannot relax this threshold, you can skip this unittest when in SR mode similar to https://github.com/facebookresearch/faiss/blob/697b6ddf558ef4ecb60e72e828c25a69723639c1/faiss/gpu/test/test_cagra.py#L13 (ideally we don't have to do this)

tests/test_contrib.py
mulugetam: @mengdilin The test case passes in my machine with a `ndiff` value of `50` (not sure why it fails here). I will do another push that resolves the conflict. I have also updated the name from `avx512-sr` to `avx512_spr` to be consistent with what numpy returns.

mulugetam:
@mengdilin Only timeout and anaconda_telemetry errors.

mengdilin:
@mulugetam there were some CI issues yesterday that we resolved. Can you rebase to latest commit and retry? The anaconda issues should definitely be resolved on the latest.

mengdilin:
@mulugetam `faiss/python/swigfaiss_avx512_spr.swig` shouldn't be checked in. Looks like it's part of .gitignore but it got accidentally tracked. Can you revert this?

mulugetam:
@mengdilin untracked.

PR #4026, opened 2024-11-13
Add remove ids for indexrefine

Tickdack:


faiss/IndexRefine.cpp
```diff
@@ -54,6 +54,14 @@ void IndexRefine::add(idx_t n, const float* x) {
     ntotal = refine_index->ntotal;
 }
 
+size_t IndexRefine::remove_ids(const IDSelector& sel) {
+    FAISS_THROW_IF_NOT(is_trained);
+    base_index->remove_ids(sel);
+    refine_index->remove_ids(sel);
+    ntotal = refine_index->ntotal;
+    return -1;
```
gtwang01, line 62: I believe the return value is supposed to be the number of vectors removed. You should be able to do this by getting the ntotal of the refine_index (or base_index - they should be the same) before applying the remove_ids operation.

gtwang01 (changes requested):

faiss/IndexRefine.cpp
Tickdack: sorry, I will fix it

faiss/IndexRefine.cpp
```diff
@@ -54,6 +54,14 @@ void IndexRefine::add(idx_t n, const float* x) {
     ntotal = refine_index->ntotal;
 }
 
+size_t IndexRefine::remove_ids(const IDSelector& sel) {
+    FAISS_THROW_IF_NOT(is_trained);
+    base_index->remove_ids(sel);
+    refine_index->remove_ids(sel);
+    ntotal = refine_index->ntotal;
+    return ntotal;
```
gtwang01, line 62: Sorry, let me clarify. The return value should be the number of vectors removed, not the number of vectors remaining.
```suggestion
    ntotal_before = refine_index->ntotal
    base_index->remove_ids(sel);
    refine_index->remove_ids(sel);
    ntotal_after = refine_index->ntotal;
    return ntotal_after - ntotal_before;
```

gtwang01:
Can you add some unit tests to ensure reliability?

Tickdack:
> Can you add some unit tests to ensure reliability?

ok,i will add unit tests

Tickdack:
> It seems like from this implementation that if the base index's remove_ids works but the refine index's throws, then we end up in a state where the base index and refine index are no longer in sync and are no longer useful. Is there any way to prevent this? At the very least, to notify the user if such a thing has happened and what it means? I think the user would essentially have to recreate the index if this happens.

I think some index in faiss not support remove, and use try catch to remind users. so i think it is reasonable

gtwang01:
What's the status on unit testing?

gtwang01:
Closing this PR for now - please reopen if you update it.

PR #4039, opened 2024-11-21
ROCm support for bfloat16

ItsPitt:
Updated the hipify script to handle bfloat16 conversion and unblocked disabling for ROCm.


ItsPitt:
Is the windows build working correctly? It seems to be failing somewhere in setup.

asadoughi:
> Is the windows build working correctly? It seems to be failing somewhere in setup.

Yes, there's a bug in the latest conda-build: addressing in #4040

PR #4048, opened 2024-12-03
Added bucketization feature to IndexLSH.

chenyuanTKCY:
I have improved the sa_encode method within the IndexLSH class to enhance its functionality. With this modification, users are now able to clearly determine which hash bucket the index data has been assigned to. Specifically, after the sa_encode method processes the input data and computes the hash values, the method now stores the resulting hash bucket mappings in a 2D vector, providing a clear association between each data point and its corresponding hash bucket. This makes it easier for users to track and manage the distribution of index data across hash buckets during the Locality-Sensitive Hashing (LSH) process.

### DIFF
--- a/faiss/IndexLSH.h
+++ b/faiss/IndexLSH.h
@@ -66,6 +66,7 @@ struct IndexLSH : IndexFlatCodes {
     void sa_encode(idx_t n, const float* x, uint8_t* bytes) const override;
 
     void sa_decode(idx_t n, const uint8_t* bytes, float* x) const override;
+
 };
 
 } // namespace faiss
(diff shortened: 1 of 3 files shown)


asadoughi:
Hi @chenyuanTKCY - thanks for the pull request. I think there are a few issues with the code: one being the location and the other being the functionality. 
1) On functionality, `bucket_mapping[i] = bucket_number` appears to be sufficient so we only need a 1-D vector, not a 2-D vector.
2) In the diff, I see you have added a new class. I would consider filing `hash_code` helper method under `faiss/cppcontrib` and graduating to IndexLSH in a future iteration if it appeared more useful to other LSH users. Could you move `hash_code(IndexLSH* lsh_index, idx_t n, const float* x, uint8_t* bytes, int bucket_count, std::vector<uint64_t>& bucket_mapping)` to something like `faiss/cppcontrib/lsh_helper.h`?

asadoughi:
Closing this PR as stale for now. If you can make the suggested updates, please file a new PR.

PR #4058, opened 2024-12-05
Unit tests for distances_simd.cpp

gtwang01:
Summary: distances_simd.cpp had insufficient test coverage. While there may be some codepaths untested due to compiler flags, this should cover most of it.

Differential Revision: D66792601


asadoughi:
Closed in 80a8ff5

PR #4074, opened 2024-12-07
Fix ScalarQuantizer to use full bucket range

ddrcoder:
Summary:
Scalar quantizers have an off-by-one bug. Even when the quantizer is trained to cover the full trained data range, it only ends up encoding the maximum quantized value for exact matches on the maximum value instead of having the upper bound correspond to the upper bound of the last bucket. This means a 4-bit quantizer often only uses 15 / 16 possible values, effectively 3.9 bits. 

Existing tests show significant movements for 4-bit quantization results; +9% recall in one of the unit tests.

The fix is to use `2^n - eps` everywhere `2^n - 1` is used.

This would break existing stored indices, though, so a backwards compatibility fix is included; ranges are adjusted at deserialization time to simulate old behavior.

Differential Revision: D66909688

### DIFF
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -11,6 +11,7 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <numeric>
 
 #include <faiss/impl/platform_macros.h>
 #include <omp.h>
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@ddrcoder this causes multiple backward compatibility problems. Add a new quantization type?

mdouze:
> @ddrcoder this causes multiple backward compatibility problems. Add a new quantization type?

I agree that the current bucket allocation is suboptimal. 

The bucket training uses parameters `RangeStat` here
https://github.com/facebookresearch/faiss/blob/main/faiss/impl/ScalarQuantizer.h#L54

Could you define a new `RangeStat` entry rather than a new quantization `QuantizationType`?

ddrcoder:
> @ddrcoder this causes multiple backward compatibility problems. Add a new quantization type?

That's addressed; the ranges are adapted to produce the same values as before. I do still need to prove that with a unit test, though.

mdouze:
Please address my comment above. 
Only the training code should be need to be adapted, there is no reason to change the file format or the distance computation code.

PR #4084, opened 2024-12-10
Link cuVS Docs

tarang-jain:
Small updates to the ReadMe files. More detailed description in a follow up PR for the wiki.

Remove the cuvs conda CI checks


INSTALL.md
```diff
@@ -37,8 +37,8 @@ $ conda install -c pytorch/label/nightly faiss-cpu
 # GPU(+CPU) version
 $ conda install -c pytorch/label/nightly -c nvidia faiss-gpu=1.9.0
 
-# GPU(+CPU) version with NVIDIA RAFT
-conda install -c pytorch -c nvidia -c rapidsai -c conda-forge faiss-gpu-raft=1.9.0 pytorch pytorch-cuda numpy
```
asadoughi, line 41: `conda install -c pytorch -c rapidsai -c conda-forge pytorch/label/nightly::faiss-gpu-cuvs pytorch pytorch-cuda numpy` works for me.

asadoughi (commented):
Thanks!

It looks like CUDA12 is working. You may want to take a look at the CUDA11 [job](https://github.com/facebookresearch/faiss/actions/runs/12270790747/job/34236594382#step:3:749) when you have a chance.

INSTALL.md
```diff
@@ -6,7 +6,7 @@ pre-release nightly builds.
 
 - The CPU-only faiss-cpu conda package is currently available on Linux (x86-64 and aarch64), OSX (arm64 only), and Windows (x86-64)
 - faiss-gpu, containing both CPU and GPU indices, is available on Linux (x86-64 only) for CUDA 11.4 and 12.1
-- faiss-gpu-raft containing both CPU and GPU indices provided by NVIDIA RAFT, is available on Linux (x86-64 only) for CUDA 11.8 and 12.1.
+- faiss-gpu-cuvs containing GPU indices provided by NVIDIA cuVS, is available on Linux (x86-64 only) for CUDA 11.8 and 12.4.
```
cjnolet, line 9: Perhaps we should make "NVIDIA cuVS" a link to the cuVS github repository here.

tarang-jain:
@asadoughi The cuda11.8 package is now fixed. This PR can be merged.

INSTALL.md
tarang-jain: The raft nightly would no longer exist. So I've removed this. Instead we have install instructions for the cuvs nightly.

tarang-jain:
I have relaxed the `cuda_constraints` in the run environment. Also, I am avoiding the use of `nvidia` channel for pkg building where it can be avoided (CUDA 12.4). In general, `conda-forge` is more robust than `nvidia` pre-CUDA 12.5 for CUDA related dependencies. (As I had indicated earlier, there might be test failures caused by conflicts because `pytorch-cuda` still uses the `nvidia` channel. If that happens I will remove the conditional removal of the `nvidia` channel. AFAIR though that was not the case with `libcuvs=24.08`. We might have to re-visit this when upgrading to cuvs=24.12).

asadoughi (changes requested):
Thanks for fixing the cuda11 build!   I left one comment inline to address the installation command is working for users as soon as this is merged.

INSTALL.md
```diff
@@ -37,8 +37,8 @@ $ conda install -c pytorch/label/nightly faiss-cpu
 # GPU(+CPU) version
 $ conda install -c pytorch/label/nightly -c nvidia faiss-gpu=1.9.0
 
-# GPU(+CPU) version with NVIDIA RAFT
-conda install -c pytorch -c nvidia -c rapidsai -c conda-forge faiss-gpu-raft=1.9.0 pytorch pytorch-cuda numpy
+# GPU(+CPU) version with NVIDIA cuVS
+conda install -c pytorch -c rapidsai -c conda-forge -c nvidia faiss-gpu-cuvs pytorch pytorch-cuda numpy
```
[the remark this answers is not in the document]
tarang-jain: @asadoughi I have made this change. I'd also like to point out that on the RAPIDS side, we use the cuda-version package to identify the cuda version of the nightly (if we dont do this, it'll install the cuda12 package by default). Another thing to note is that the run-time dependency on cuda-version is relaxed for the cuvs packages (hence the range for cuda-version and not a specific version. What I am trying to say (though this is yet to be tested) is that if I as a user have CUDA12.2 installed, I can still use the conda package built with CUDA 12.4.

INSTALL.md
asadoughi: The proposed command in your diff installs an older version of the nightly for me compared to the command I provided in my previous comment.

asadoughi (commented):
Thanks!

faiss/gpu/impl/CuvsIVFPQ.cu
```diff
@@ -229,6 +229,7 @@ void CuvsIVFPQ::searchPreassigned(
         Tensor<idx_t, 2, true>& outIndices,
         bool storePairs) {
     // TODO: Fill this in!
```
cjnolet, line 231: Can you create an issue and reference the issue number here just so we don't lose sight of this?

INSTALL.md
tarang-jain: It seems to be installing libcuvs=24.12 for me. Prefixing the package and using the same channel label as a channel should be equivalent. Perhaps we are doing something differently. I am using this command:
`conda install -c pytorch/label/nightly -c rapidsai -c conda-forge -c nvidia faiss-gpu-cuvs 'cuda-version>=12.0,<=12.5'`

faiss/gpu/impl/CuvsIVFPQ.cu
tarang-jain: Done!

INSTALL.md
tarang-jain: Updated this!

PR #4094, opened 2024-12-16
Conditionally compile extras like benchmarks and demos

artem:
While embedding faiss as a subproject, these targets are quite useless. This PR adss a new `FAISS_ENABLE_EXTRAS` CMake option to exclude them, default is ON to preserve current behaviour.


gtwang01:
Can you rebase and retry? We recently fixed an issue that caused tests to fail

artem:
Rebased

PR #4095, opened 2024-12-16
Fix IndexIVFFastScan reconstruct_from_offset method

### LINKED ISSUE #4089: IndexIVFPQFastScan crashes with certain nlist values
# IndexIVFPQFastScan crashes with certain nlist values

The `IndexIVFPQFastScan` index in the Faiss similarity search library crashes when the `nlist` parameter (number of Voronoi cells) is not byte-aligned (e.g. 256, 65536). This occurs when calling either `index.reconstruct(i)` or `index.reconstruct_batch(ids)` after adding vectors.

## Platform
- OS: MacOS 
- Installed from: Anaconda
- Running on: CPU
- Interface: Python

## Reproduction Instructions
```python
import numpy as np
import faiss

faiss.omp_set_num_threads(16)

dim = 64  # Dimensionality of vectors 
train_vectors = np.random.random((100000, dim)).astype("float32")

def run_example(nlist: int):
    # Create an IVF index
    index = faiss.IndexIVFPQFastScan(faiss.IndexFlatL2(dim), dim, nlist, dim // 2, 4, faiss.METRIC_L2) 
    index.make_direct_map(True)
    index.train(train_vectors)
    index.add(train_vectors)
    
    print(index.reconstruct(24545))
    print(index.reconstruct_batch(np.arange(10)))

run_example(256) # This works fine
run_example(100) # Crashes with error below whenever `nlist` is not byte-aligned
```

## Error Message
```
libc++abi: terminating due to uncaught exception of type faiss::FaissException: Error in faiss::idx_t faiss::Level1Quantizer::decode_listno(const uint8_t *) const at 
/Users/runner/miniconda3/conda-bld/faiss-pkg_1728491206806/work/faiss/IndexIVF.cpp:149: 
Error: 'list_no >= 0 && list_no < nlist' failed
```

Note: If you set `by_residual=True` in the `IndexIVFPQFastScan` constructor, the error only occurs for `reconstruct_batch()`, not `reconstruct()`.

## Potential Reasons

The issue seems to stem from the values in `codes` in `IndexIVFPQFastScan::sa_decode()`: [here](https://github.com/facebookresearch/faiss/blob/87a760e63173590fcb55ab2fa2cf3c82ca1c90f4/faiss/IndexIVFPQFastScan.cpp#L285). 

The potential reason is that these codes are unpacked incorrectly. Which makes the `list_no` value read from `code` invalid when `nlist` is not byte-aligned, causing the bounds check `'list_no >= 0 && list_no < nlist'` to fail.

Additional note: The call to `decode_listno` is unnecessary when `by_residual` is false.

alisafaya:
Resolves issue #4089 - IndexIVFPQFastScan crashes with certain nlist values

The `reconstruct_from_offset` method in `IndexIVFFastScan` was incorrectly reconstructing vectors, causing crashes when the `nlist` parameter was not byte-aligned (e.g. 100 instead of 256).

The root cause was that the `list_no` (Voronoi cell number) was not being properly encoded into the `code` vector before passing it to the `sa_decode` function. This resulted in invalid `list_no` values being read in `sa_decode`, triggering the assertion failure `'list_no >= 0 && list_no < nlist'` when `nlist` in some cases.

This PR fixes the issue with the following changes to `reconstruct_from_offset`:

1. Encode the `list_no` into the beginning of the `code` vector using the existing `encode_listno` method
2. Start the `BitstringWriter` after the coarse code portion of `code` (shifted by `coarse_code_size()` bytes)
3. Remove the residual centroid addition logic, as it is already handled in `sa_decode`

After these changes:
- Crashes no longer occur for any `nlist` value 
- Reconstruction is now correct, matching the output of `IndexIVFPQ`

Fixes #4089

Please review and let me know if any changes are needed. Thanks!

### DIFF
--- a/benchs/bench_fw/index.py
+++ b/benchs/bench_fw/index.py
@@ -677,6 +677,13 @@ class Index(IndexBase):
                 lambda: add_preassigned(index_ivf, xbt, QI.ravel()),
                 once=True,
             )
+        elif isinstance(index, faiss.IndexIDMap):
+            _, t, _ = timer(
+                "add_with_ids",
+                lambda: index.add_with_ids(
+                    xb, np.arange(len(xb), dtype='int32')),
+                once=True,
+            )
         else:
             _, t, _ = timer(
                 "add",
--- a/faiss/IndexIVFFastScan.cpp
+++ b/faiss/IndexIVFFastScan.cpp
@@ -1353,10 +1353,11 @@ void IndexIVFFastScan::reconstruct_from_offset(
         int64_t offset,
         float* recons) const {
     // unpack codes
-    InvertedLists::ScopedCodes list_codes(invlists, list_no);
-    std::vector<uint8_t> code(code_size, 0);
+    size_t coarse_size = coarse_code_size();
+    std::vector<uint8_t> code(coarse_size + code_size, 0);
     encode_listno(list_no, code.data());
-    BitstringWriter bsw(code.data() + coarse_code_size(), code_size);
+    InvertedLists::ScopedCodes list_codes(invlists, list_no);
+    BitstringWriter bsw(code.data() + coarse_size, code_size);
 
     for (size_t m = 0; m < M; m++) {
         uint8_t c =
--- a/faiss/IndexIVFFastScan.cpp
+++ b/faiss/IndexIVFFastScan.cpp
@@ -1372,10 +1372,11 @@ void IndexIVFFastScan::reconstruct_orig_invlists() {
     FAISS_THROW_IF_NOT(orig_invlists != nullptr);
     FAISS_THROW_IF_NOT(orig_invlists->list_size(0) == 0);
 
+#pragma omp parallel for if (nlist > 100)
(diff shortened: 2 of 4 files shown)


asadoughi:
Thanks for the contribution! Can you add unit tests, in Python should be sufficient, for the two cases you referenced? 
- Crashes no longer occur for any nlist value
- Reconstruction is now correct, matching the output of IndexIVFPQ

alisafaya:
I added unit tests, and fixed another bug in [`IndexIVFFastScan::reconstruct_orig_invlists()`](https://github.com/facebookresearch/faiss/blob/1e4586a5a0d7ceb8a538c4878d08b0623a676a08/faiss/IndexIVFFastScan.cpp#L1241).

The bug involves incorrect usage of invlists vs orig_invlists when getting the list size. The original code was using orig_invlists->list_size() which is incorrect since we want to read from the source invlists when getting codes and sizes. The fix changes it to use invlists->list_size() to be consistent with where we're reading the codes and IDs from.

mdouze:
LGTM

PR #4122, opened 2025-01-10
Remove unused(and wrong) io macro

chasingegg:
`READSTRING` is not used and also the function's meaning is actually WRITESTRING, just remove it


mnorris11:
Merged, thanks for the contribution!

PR #4123, opened 2025-01-10
Add `ngpu` default argument to `knn_ground_truth`

Di-Is:
This pull request introduces a new default argument, `ngpu=-1`, to the `knn_ground_truth` function in the `faiss.contrib`.

### Bug Fix

In the current implementation, running tests under the tests directory (CPU tests) in an environment with faiss-gpu installed would inadvertently use the GPU and cause unintended behavior.
This pull request prevents the GPU from being used during CPU-only tests by explicitly controlling GPU allocation via the ngpu parameter.

### API Consistency

Other functions that call `faiss.get_num_gpus` in `faiss.contrib`, such as `range_search_max_results` and `range_ground_truth`, already include the `ngpu` argument.
Adding this parameter to `knn_ground_truth` will ensure consistency across the API, reduce potential confusion, and improve ease of use.

### DIFF
--- a/contrib/exhaustive_search.py
+++ b/contrib/exhaustive_search.py
@@ -11,7 +11,7 @@ import logging
 
 LOG = logging.getLogger(__name__)
 
-def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2, ngpu=-1):
+def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2, shard=False, ngpu=-1):
     """Computes the exact KNN search results for a dataset that possibly
     does not fit in RAM but for which we have an iterator that
     returns it block by block.
@@ -28,7 +28,9 @@ def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2, ngpu=-1):
 
     if ngpu:
         LOG.info('running on %d GPUs' % ngpu)
-        index = faiss.index_cpu_to_all_gpus(index)
+        co = faiss.GpuMultipleClonerOptions()
+        co.shard = shard
+        index = faiss.index_cpu_to_all_gpus(index, co=co, ngpu=ngpu)
 
     # compute ground-truth by blocks, and add to heaps
     i0 = 0
--- a/contrib/exhaustive_search.py
+++ b/contrib/exhaustive_search.py
@@ -11,7 +11,7 @@ import logging
 
 LOG = logging.getLogger(__name__)
 
-def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2):
+def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2, ngpu=-1):
     """Computes the exact KNN search results for a dataset that possibly
     does not fit in RAM but for which we have an iterator that
     returns it block by block.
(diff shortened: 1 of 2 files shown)


contrib/exhaustive_search.py
```diff
@@ -23,8 +23,11 @@ def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2):
     rh = faiss.ResultHeap(nq, k, keep_max=keep_max)
 
     index = faiss.IndexFlat(d, metric_type)
-    if faiss.get_num_gpus():
-        LOG.info('running on %d GPUs' % faiss.get_num_gpus())
+    if ngpu == -1:
```
junjieqi, line 26: If I understand it correctly, you are trying to find a method to enable gpu based some parameter. I'd suggest that you could use a bool because right now, we either not use  gpu or we use all.

Di-Is: Thanks for your review.

Since the other functions `range_search_max_results` and `range_ground_truth` control the number of GPUs with the argument `ngpu` argument, I thought it would be good to add `ngpu` for API uniformity.

junjieqi: then probably you need to pass the ngpu as part of parameters in the index_cpu_to_all_gpus, like
```
index_gpu = faiss.index_cpu_to_all_gpus(index, co=co, ngpu=ngpu)
```

https://github.com/facebookresearch/faiss/blob/b9fe1dcdf71602f5d733dbd78adce06bba20d615/contrib/exhaustive_search.py#L164

It will use ngpu to control the number of GPUs

contrib/exhaustive_search.py
```diff
@@ -23,8 +23,11 @@ def knn_ground_truth(xq, db_iterator, k, metric_type=faiss.METRIC_L2):
...
+        ngpu = faiss.get_num_gpus()
+
+    if ngpu:
+        LOG.info('running on %d GPUs' % ngpu)
         index = faiss.index_cpu_to_all_gpus(index)
```
junjieqi, line 31: index_gpu = faiss.index_cpu_to_all_gpus(index, co=co, ngpu=ngpu)

junjieqi:
In addition, I saw many of builds failed, could you also help fix them? Thanks

contrib/exhaustive_search.py
Di-Is: I modified it, but left the variable name as “index” for now.

junjieqi:
@Di-Is  looks like there is still a falied CI. Would you mind to fix it? Thanks

Di-Is:
> @Di-Is looks like there is still a falied CI. Would you mind to fix it? Thanks

My understanding is that there is an error in the C++ test.
Sorry, but it is difficult to deal with errors in code that has not been changed in this PR.

junjieqi:
@Di-Is If I checked the history of CI https://github.com/facebookresearch/faiss/actions/runs/12777190078, I don't think any of them has been broken recently. The failure for your CI should be related with your change. You could double check by removing the some of your code in the test file to see if anything will break.

Di-Is:
I checked locally and could not reproduce...
Could you please re-run the failed jobs only?
(I'm thinking that the failing test is a memory leak test and the test itself may be unstable.)

PR #4128, opened 2025-01-14
more fast-scan reconstruction

mdouze:
Summary: Fix reconstruction code for the fast-scan and IVF fast-scan indices.

Differential Revision: D68159014


mdouze:
This is actually a follow up on https://github.com/facebookresearch/faiss/pull/4095

PR #4136, opened 2025-01-23
repro CUDA 11.4.4 failure in test_gpu_basics (version matthijs)

mdouze:
- **repro CUDA 11.4.4 failure in test_gpu_basics (#4135)**
- **repro error**


mdouze:
Made a separate PR so that I can push to it

PR #4137, opened 2025-01-24
fix: gpu tests link failure with static lib

simshi:
# The bug
build gpu test with static lib failed undefined reference from gpu source code to cpu source, e.g.
`GpuAutoTune.cpp:(.text+0x3bc): undefined reference to 'typeinfo for faiss::Index'`

# build config:
`cmake -S . -B build -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=ON -DFAISS_ENABLE_GPU=ON`
the link command would be like: `c++ -o faiss/gpu/test/TestCodePacking faiss/libfaiss.a -Wl,--push-state,--whole-archive  faiss/gpu/libfaiss_gpu.a  -Wl,--pop-state ...`
key point is linking to faiss_gpu after faiss which is a classic link order issue of "undefined reference"

# my fix:
make faiss_gpu an `OBJECT` library.

# Bouns
TestXXX files get small because '--whole-archive' propagated to test link command.

### DIFF
--- a/faiss/gpu/CMakeLists.txt
+++ b/faiss/gpu/CMakeLists.txt
@@ -258,12 +258,12 @@ if(FAISS_ENABLE_CUVS)
           utils/CuvsUtils.cu)
 endif()
 
-add_library(faiss_gpu STATIC ${FAISS_GPU_SRC})
-set_target_properties(faiss_gpu PROPERTIES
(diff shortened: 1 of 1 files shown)


gtwang01:
Can you give us some more context on the bug? How do we reproduce it and how does this fix it?

PR #4152, opened 2025-01-29
Update CAGRA docs

tarang-jain:
Update CAGRA train and add docs to reflect current capabilities.


tarang-jain:
@asadoughi this is a very simple docs update PR to note what CAGRA train and add funcs are used for.

divyegala (approved):

PR #4154, opened 2025-01-29
Fix bug in PCA balanced bins eigenvalues accumulator

pankajsingh88:
Summary: PCA balanced bins should be product of eigenvalues not sum of eigenvalues

Differential Revision: D68839548


alibeklfc:
Closing this PR after discussion with the author

PR #4155, opened 2025-01-29
Correct capitalization of FAISS to Faiss

mnorris11:
Summary: Changed everything except the CHANGELOG.md which I assume should not be changed after it lands.

Differential Revision: D68842649


mnorris11:
Merged in 1334d169b9279a8c581835ef6025f4752b246ce8

PR #4159, opened 2025-01-31
Fix the order of parameters in bench_scalar_quantizer_distance.

mulugetam:
I believe the order of the parameters was not intended to be that way. This PR fixes it.


junjieqi:
@mulugetam Hi, looks like there is a test failure, would you like to fix it?

mulugetam:
@junjieqi not failing on my local machine and the change should not cause the fail (you may review the change). Perhaps you re-run CI? I can also push a dummy commit.

PR #4199, opened 2025-02-19
Memory-mapping and Zero-copy deserializers

alexanderguzhva:
This PR introduces a backport of a combination of https://github.com/zilliztech/knowhere/pull/996 and https://github.com/zilliztech/knowhere/pull/1032 that allow to have memory-mapped and zerocopy indces.

The root underlying idea is that we replace certain `std::vector<>` containers with a custom `faiss::MaybeOwnedVector<>` container, which may behave either as `std::vector<>`, or as a view of a certain pointer / descriptor. We don't replace all the instances of `std::vector<>`, but the largest ones.

This change affects `IndexFlatCodes`-based and `IndexHNSW` CPU indices.

(done) alter IVF lists as well.
(done) alter binary indices as well.

Memory-mapped index works like this:
```C++
std::unique_ptr<faiss::Index> index_mm(
            faiss::read_index(filenamename.c_str(), faiss::IO_FLAG_MMAP_IFC));
```
In theory, it should be ready to be used from Python. All the descriptor management should be working.

Zero-copy index works like this:
```C++
#include <faiss/impl/zerocopy_io.h>

faiss::ZeroCopyIOReader reader(buffer.data(), buffer.size());
std::unique_ptr<faiss::Index> index_zc(faiss::read_index(&reader));
```
All the pointer management for `faiss::ZeroCopyIOReader` should be handled manually.
I'm not sure how to plug this into Python yet, maybe, some ref-counting is required.

(done) some refactoring

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -69,12 +69,12 @@ set(FAISS_SRC
   impl/io.cpp
   impl/kmeans1d.cpp
   impl/lattice_Zn.cpp
+  impl/mapped_io.cpp
   impl/pq4_fast_scan.cpp
   impl/pq4_fast_scan_search_1.cpp
   impl/pq4_fast_scan_search_qbs.cpp
   impl/residual_quantizer_encode_steps.cpp
-  impl/io.cpp
-  impl/lattice_Zn.cpp
+  impl/zerocopy_io.cpp
   impl/NNDescent.cpp
   invlists/BlockInvertedLists.cpp
   invlists/DirectMap.cpp
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -53,7 +53,166 @@
 #include <faiss/IndexBinaryHash.h>
 #include <faiss/IndexBinaryIVF.h>
 
+// mmap-ing and viewing facilities
+#include <faiss/impl/maybe_owned_vector.h>
+
+#include <faiss/impl/mapped_io.h>
+#include <faiss/impl/zerocopy_io.h>
+
 namespace faiss {
+
+/*************************************************************
+ * Mmap-ing and viewing facilities
+ **************************************************************/
+
+template <typename VectorT>
+void read_vector_with_size(VectorT& target, IOReader* f, size_t size) {
+    ZeroCopyIOReader* zr = dynamic_cast<ZeroCopyIOReader*>(f);
+    if (zr != nullptr) {
+        if constexpr (is_maybe_owned_vector_v<VectorT>) {
+            // create a view
+            char* address = nullptr;
+            size_t nread = zr->get_data_view(
+                    (void**)&address,
+                    sizeof(typename VectorT::value_type),
+                    size);
+
+            FAISS_THROW_IF_NOT_FMT(
+                    nread == (size),
+                    "read error in %s: %zd != %zd (%s)",
+                    f->name.c_str(),
+                    nread,
+                    size_t(size),
+                    strerror(errno));
+
+            VectorT view = VectorT::create_view(address, nread);
+            target = std::move(view);
+
+            return;
+        }
+    }
+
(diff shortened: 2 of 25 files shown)


alexanderguzhva:
@mdouze Would please be able to suggest a solution for the following problem. Because I changed the container type for `IndexFlatCodes::codes` field, it triggered some errors in python unit tests, which are related to these `.codes` (bcz the field type got changed). What would be the most meaningful solution for dealing with it? Basically, I'm not sure how to minimize the amount of backward-compatibility problems, as I am not an expert swig user

faiss/CMakeLists.txt
```diff
@@ -69,12 +69,12 @@ set(FAISS_SRC
   impl/io.cpp
   impl/kmeans1d.cpp
   impl/lattice_Zn.cpp
+  impl/mapped_io.cpp
   impl/pq4_fast_scan.cpp
   impl/pq4_fast_scan_search_1.cpp
   impl/pq4_fast_scan_search_qbs.cpp
   impl/residual_quantizer_encode_steps.cpp
-  impl/io.cpp
-  impl/lattice_Zn.cpp
```
mdouze, line 77: why do these two file disappear?

faiss/impl/index_read.cpp
```diff
@@ -53,8 +53,167 @@
...
+
+            return;
+        }
+    }
+
+    // the default case
+    READVECTOR(target);
+}
+
+template <typename VectorT>
```
mdouze, line 157: possible to factorize with previous function?

faiss/impl/index_read.cpp
```diff
@@ -53,8 +53,167 @@
...
+            return;
+        }
+    }
+
+    target.resize(size);
+    READANDCHECK(target.data(), size);
+}
+
+template <typename VectorT>
+void read_vector(VectorT& target, IOReader* f) {
```
mdouze, line 100: Would it be possible to enable this via an IO flag ? The IVF memory mapped files work on regular `FileIOReader`

https://github.com/facebookresearch/faiss/blob/main/faiss/invlists/OnDiskInvertedLists.cpp#L771

mdouze (commented):
Thanks for tackling this. See a few inline reviews. 
As long as the interface of MaybeOwnedVector is sufficiently close of std::vector, it should work seamlessly in Python (duck typing). I'll see if some adaptations to the swig interface are required. 
It is not clear to me why the `MappedFileIOReader` is needed rather than the regular `FileIOReader`. Is it to maintain the ownership of the reader within the `MaybeOwnedVector` and close the file when deallocated?

mdouze:
This patch should fix the python test: 
https://gist.github.com/mdouze/83eff60f99ae1a3cba524c6b8320dd8f
could you integrate it?

faiss/CMakeLists.txt
alexanderguzhva: duplicates

faiss/impl/index_read.cpp
alexanderguzhva: [> OnDiskInvertedLists is unsupported on Windows.](https://github.com/facebookresearch/faiss/blob/55d022fbb055c522f7738d709105de7aae6926cd/tests/test_contrib.py#L716-L719)

my code allows mmap functionality on Windows as well. Let me take a look

alexanderguzhva:
> It is not clear to me why the MappedFileIOReader is needed rather than the regular FileIOReader.
in order to have a correct support for Windows

alexanderguzhva:
@mdouze I'm going to add IVFs and maybe Binary indices today as well, so please don't merge it yet :)

alexanderguzhva:
@mdouze please let me know which other unit tests are useful to have
I've added the code for `ArrayInvertedLists` and binary vectors as well.

alexanderguzhva:
the PR is ready to be reviewed

PR #4235, opened 2025-03-11
RaBitQ implementation

alexanderguzhva:
This is a reference implementation of the https://arxiv.org/pdf/2405.12497
> Jianyang Gao, Cheng Long, "RaBitQ: Quantizing High-Dimensional Vectors with a Theoretical Error Bound for Approximate Nearest Neighbor Search".

The goal is to correctly set up the internals using Faiss.

The following comments for the implementation:
* The code does not include the computations for the symmetric distance, because it is absent in the original article. This can be added later, though.
* The original `RaBitQ` includes random matrix rotation as a part of it, but I've decided to rely on external `faiss::IndexPreTransform` and `faiss::RandomRotationMatrix` facilities. 
* Certain features required internal changes in `faiss::IndexIVF`, but I did that as least invasive as possible, without breaking the backward compatibility.
* Not sure about naming convensions, maybe certain classes and structures need to be renamed
* `METRIC_INNER_PRODUCT` is supported as well
* More unit tests are needed?
* I did not bring any hardware-specific optimizations, bcz this is a reference implementation. Certain `simdlib` facilities may be added later, if needed

Here's how to use IndexRaBitQ
```Python
        ds = datasets.SyntheticDataset(...)

        index_rbq = faiss.IndexRaBitQ(ds.d, faiss.METRIC_L2)
        index_rbq.qb = 8

        # wrap with random rotations
        rrot = faiss.RandomRotationMatrix(ds.d, ds.d)
        rrot.init(rrot_seed)

        index_cand = faiss.IndexPreTransform(rrot, index_rbq)
        index_cand.train(ds.get_train())
        index_cand.add(ds.get_database())
```

Here's how to use IndexIVFRaBitQ
```Python
        ds = datasets.SyntheticDataset(...)

        index_flat = faiss.IndexFlat(ds.d, faiss.METRIC_L2)
        index_rbq = faiss.IndexIVFRaBitQ(index_flat, ds.d, nlist, faiss.METRIC_L2)
        index_rbq.qb = 8

        # wrap with random rotations
        rrot = faiss.RandomRotationMatrix(ds.d, ds.d)
        rrot.init(rrot_seed)

        index_cand = faiss.IndexPreTransform(rrot, index_rbq)
        index_cand.train(ds.get_train())
        index_cand.add(ds.get_database())
```

### DIFF
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -475,7 +475,7 @@ float RaBitQuantizer::RaBitDistanceComputer::symmetric_dis(idx_t i, idx_t j) {
     FAISS_THROW_MSG("Not implemented");
 }
 
-float symmetric_dis_core(
+float RaBitQuantizer::symmetric_dis_core(
         const uint8_t* code_i,
         const float* centroid_i,
         const uint8_t* code_j,
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -215,7 +215,7 @@ float RaBitDistanceComputerNotQ::distance_to_code_L2(
     //
     // compute <q,o> using floats
     float dot_qo = 0;
-    // It was a willful decision (after the discussion) to not to pre-cache 
+    // It was a willful decision (after the discussion) to not to pre-cache
     //   the sum of all bits, just in order to reduce the overhead per vector.
     uint64_t sum_q = 0;
     for (size_t i = 0; i < d; i++) {
@@ -345,7 +345,7 @@ float RaBitDistanceComputerQ::distance_to_code_L2(const uint8_t* code) const {
         dot_qo += (count_dot << j);
     }
 
-    // It was a willful decision (after the discussion) to not to pre-cache 
+    // It was a willful decision (after the discussion) to not to pre-cache
     //   the sum of all bits, just in order to reduce the overhead per vector.
     uint64_t sum_q = 0;
     {
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -64,7 +64,7 @@ void RaBitQuantizer::compute_codes_core(
 
     // compute codes
 #pragma omp parallel for if (n > 1000)
-    for (size_t i = 0; i < n; i++) {
+    for (int64_t i = 0; i < n; i++) {
         // ||or - c||^2
         float norm_L2sqr = 0;
         // ||or||^2, which is equal to ||P(or)||^2 and ||P^(-1)(or)||^2
@@ -137,7 +137,7 @@ void RaBitQuantizer::decode_core(
     const float inv_d_sqrt = (d == 0) ? 1.0f : (1.0f / std::sqrt((float)d));
 
 #pragma omp parallel for if (n > 1000)
-    for (size_t i = 0; i < n; i++) {
+    for (int64_t i = 0; i < n; i++) {
         const uint8_t* code = codes + i * code_size;
 
         // split the code into parts
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -1,0 +1,487 @@
+#include <faiss/impl/RaBitQuantizer.h>
+
+#include <algorithm>
+#include <cmath>
+#include <cstring>
+#include <limits>
+#include <memory>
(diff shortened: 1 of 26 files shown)


faiss/impl/RaBitQuantizer.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+#include <faiss/impl/FaissAssert.h>
+#include <faiss/impl/platform_macros.h>
+#include <faiss/utils/distances.h>
+
+namespace faiss {
+
+struct FactorsData {
+    // ||or - c||
+    float factor_0 = 0;
```
mdouze, line 18: why not give the fields proper names? Otherwise you might as well use `float factors[4]`

faiss/impl/RaBitQuantizer.cpp
```diff
@@ -0,0 +1,487 @@
...
+    // this is the scheme for popcount
+    const size_t di_8b = (d + 7) / 8;
+    const size_t di_64b = (di_8b / 8) * 8;
+
+    unsigned long long dot_qo = 0;
+    for (size_t j = 0; j < qb; j++) {
+        const uint8_t* query_j = rearranged_rotated_qq.data() + j * di_8b;
+
+        // process 64-bit popcounts
+        unsigned long long count = 0;
```
mdouze, line 331: please use explicity sized integer types (eg uint64_t)

faiss/impl/RaBitQuantizer.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+struct FactorsData {
+    // ||or - c||
+    float factor_0 = 0;
+    // sum_xb
+    float factor_1 = 0;
+    float factor_2 = 0;
+    // ||or||^2
+    float factor_3 = 0;
+};
```
mdouze, line 24: In my implementation there are only 2 floats per database vector. Do you store additional ones for efficiency?

mdouze (commented):
Looks good! Thanks for deferring the SIMD optimization to later.
I left a few comments.

faiss/impl/RaBitQuantizer.cpp
alexanderguzhva: will do. As far as I remember, I've used `unsigned long long`, bcz there were problems with compilations on `MacOS` for these `__builtin_popcount` functions

alexanderguzhva:
@mdouze two more comments after the discussion

First. `IndexIVF::get_InvertedListScanner()` signature is made into
```C++
    virtual InvertedListScanner* get_InvertedListScanner(
            bool store_pairs = false,
            const IDSelector* sel = nullptr,
            const IVFSearchParameters* params = nullptr) const;
```
because of the following logic that can override `sel`
```C++
void IndexIVF::search_preassigned(...) const {
    ...
    IDSelector* sel = params ? params->sel : nullptr;
    const IDSelectorRange* selr = dynamic_cast<const IDSelectorRange*>(sel);
    if (selr) {
        if (selr->assume_sorted) {
            sel = nullptr; // use special IDSelectorRange processing
        } else {
            selr = nullptr; // use generic processing
        }
    }
    ....
}
```
Please let me know your thoughts.

Second. RaBitQ uses 3 factors 
```C++
struct FactorsData {
    float or_minus_c_l2sqr = 0;
    float dp_multiplier = 0;
    // this is needed to support BOTH L2 and IP on the same data
    float or_l2sqr = 0;
};
```
The third one `or_l2sqr` is needed to support both L2 and IP, similar to how PQ / SQ can use the same data for different metrics. So, these three numbers per vector are independent from a chosen metric. These three factors can be reduced to two, if we make a decision to make factors dependent from the chosen metric. 
Please let me know if you'd like to have 2 or 3 factors per vector. Just double-checking.

mdouze:
Thanks for the generously commented code! 
Sorry I forgot to add I'd prefer to have 2 factors (separate metrics for L2 and IP), since the metric is known at index build time. In the long run, it is important to have a compact representation.
I can live with your change to the `get_InvertedListScanner`. 
If you can implement the 2-factor version, I'll accept the diff.

faiss/impl/platform_macros.h
```diff
@@ -76,6 +76,7 @@ inline int __builtin_clzll(uint64_t x) {
 
 #define __builtin_popcount __popcnt
```
mdouze, line 77: I'd prefer to call it popcount32 and popcount64 to have explicit bit sizes

alexanderguzhva:
updated to 2-factor version, added more unit tests

PR #4236, opened 2025-03-12
Hide fields desc_name and flat_desc_name that cache the descriptive names generated for the index and the flat index

satymish:
Summary: Hide fields desc_name and flat_desc_name that cache the descriptive names generated for the index and the flat index

Differential Revision: D71015246


mdouze:
@algoriddle an opinion?

PR #4239, opened 2025-03-14
Fix bug with metric_arg in IndexHNSW

### LINKED ISSUE #4224: METRIC_Lp issues with IndexHNSWFlat and IndexFlat
Distances are not calculated correctly for IndexHNSWFlat using METRIC_Lp.

```
#include <faiss/IndexHNSW.h>
#include <iostream>

int main() {
    faiss::IndexHNSWFlat index(1, 32, faiss::METRIC_Lp);
    index.metric_arg = 3;

    float data[1] = {0.0};
    index.add(1, data);

    float query[1] = {2.0};
    float distance;
    faiss::idx_t label;
    index.search(1, query, 1, &distance, &label);

    std::cout << "Distance: " << distance << std::endl;
    std::cout << "Label: " << label << std::endl;

    return 0;
}
```

Running the above outputs the following.

```
Distance: 1
Label: 0
```

I would expect to see a distance of (2 ^ 3) ^ 1/3 = 2.

For a simpler example, replace the index in the above code with `faiss::IndexFlat index(1, faiss::METRIC_Lp);`. This outputs the following.

```
Distance: 8
Label: 0
```

The output seems to skip the final exponent of 1/3.

I am curious if either behavior is reproducible on other machines. If so, is this intended functionality?

rohildshah:
Fix #4224.

The issue is that `IndexHNSW`'s internal `Index* storage` doesn't inherit `metric_arg`. One solution is to set `metric_arg` in `IndexHNSW::add`, which is what I did. Not sure what the best place to do this would be.

### DIFF
--- a/faiss/IndexHNSW.cpp
+++ b/faiss/IndexHNSW.cpp
@@ -210,7 +210,9 @@ IndexHNSW::IndexHNSW(int d, int M, MetricType metric)
         : Index(d, metric), hnsw(M) {}
 
 IndexHNSW::IndexHNSW(Index* storage, int M)
-        : Index(storage->d, storage->metric_type), hnsw(M), storage(storage) {}
+        : Index(storage->d, storage->metric_type), hnsw(M), storage(storage) {
+    metric_arg = storage->metric_arg;
+}
 
 IndexHNSW::~IndexHNSW() {
     if (own_fields) {
@@ -332,7 +334,6 @@ void IndexHNSW::add(idx_t n, const float* x) {
             storage,
             "Please use IndexHNSWFlat (or variants) instead of IndexHNSW directly");
     FAISS_THROW_IF_NOT(is_trained);
-    storage->metric_arg = metric_arg;
     int n0 = ntotal;
     storage->add(n, x);
     ntotal = storage->ntotal;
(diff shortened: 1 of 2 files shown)


gtwang01:
Please fix the CI error.

mdouze:
I think the proper way of doing this is to copy the metric_arg in the `IndexHNSW` constructor here 

https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L212

and use the constructor with explicit storage, ie. build an IndexFlat, set the metric_arg on the IndexFlat and call the IndexHNSW constructor with the IndexFlat parameter.

rohildshah:
Sounds good, I've made the change. Since my initial issue (#4224) was about `IndexHNSWFlat`, I'll add that someone wanting to use `METRIC_Lp` with `IndexHNSWFlat` would have the following two options:

```
faiss::IndexHNSWFlat index(d, M, faiss::METRIC_Lp);
index.storage->metric_arg = p;
```
or 
```
faiss::IndexFlat storage_index(d, faiss::METRIC_Lp);
storage_index.metric_arg = p;

faiss::IndexHNSW index(&storage_index, M);
```

I've only added a test case for the second method since the first method has worked from the start.

mdouze:
Thanks for noticing the bug and the fix!

PR #4250, opened 2025-03-21
re-land mmap diff

mdouze:
Summary:
This is an attempt to re-land the diff stack D69972250  D70982449

It was reverted because the bottom of the stack did not pass the tests.

The original code comes from Alexandr Guzhva's  https://github.com/facebookresearch/faiss/pull/4199

Differential Revision: D71614511

### DIFF
--- a/demos/demo_residual_quantizer.cpp
+++ b/demos/demo_residual_quantizer.cpp
@@ -103,7 +103,7 @@ int main() {
         index.is_trained = true;
 
         // override vectors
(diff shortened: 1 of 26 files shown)


mdouze:
Thanks for flagging, will check :-|

mdouze:
@alexanderguzhva  should be better now.

alexanderguzhva:
@mnorris11 yes, this is something that I've missed. I'll add a fix

PR #4270, opened 2025-04-01
relax input params for IndexIVFRaBitQ::get_InvertedListScanner()

alexanderguzhva:


junjieqi:
@alexanderguzhva would you mind giving more context for this one and why do we remove the FAISS_THROW_IF_NOT_MSG

PR #4291, opened 2025-04-14
moved IndexIVFPQ and IndexPQ to dynamic dispatch

mdouze:
Summary:
moved IndexIVFPQ and IndexPQ to dynamic dispatch.
Since the code was already quite modular, this boils down to make independent cpp files for the different SIMD versions.

Differential Revision: D72937709


mdouze:
This export is not functional. It is mainly for external visibility, on how dynamic dispatching will work.

mdouze:
The dynamic dispatching requires:
1. several implementations of a function / class to co-exist. The SIMDLevel object can be used as a template parameter to differentiate these functions 
2. the code to be in several compile units that use different compilation flags (ie. -mavx2, etc.)

faiss/utils/simd_levels.cpp
```diff
@@ -0,0 +1,61 @@
...
+                i != sizeof(level_names),
+                "FAISS_SIMD_LEVEL %s unknown",
+                env_var);
+        return;
+    }
+
+#ifdef __x86_64__
+    {
+        unsigned int eax, ebx, ecx, edx;
+        asm volatile("cpuid"
```
alexanderguzhva, line 41: just in case https://github.com/zilliztech/knowhere/blob/main/src/simd/instruction_set.h

mdouze: Sounds good! How portable is the cpuid.h header?

mdouze: So you would recommend to make all the distance functions pointers ? 
That's also a possibility but wondering about the performance impact...

PR #4294, opened 2025-04-14
Fix Type Error in Conditional Logic

mjs2600:
The old logic was asking for `D`'s shape, which indicated that it should be a NumPy array. However, if you check the truth value of a NumPy array, you get an error.

The code is checking for the existence of `D`. The most direct way to handle that is to make sure that it isn't `None`.

### DIFF
--- a/faiss/python/class_wrappers.py
+++ b/faiss/python/class_wrappers.py
@@ -1038,7 +1038,7 @@ def handle_VectorTransform(the_class):
 
 def handle_AutoTuneCriterion(the_class):
     def replacement_set_groundtruth(self, D, I):
-        if D:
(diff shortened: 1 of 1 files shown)


mjs2600:
Can someone let me know what the linting issue is? The results of the linter are internal.

mnorris11:
> Can someone let me know what the linting issue is? The results of the linter are internal.

Right now there's no issue. There are different internal lints/tests with levels below Error. Right now I see no errors. It should finish in an hour or two, then if there are no Errors, we can merge. If there are, I will leave a comment. The UI in how it shows in Github could be improved: lower level signals than Error should not really show as a red signal here on Github.

PR #4296, opened 2025-04-15
Split ScalarQuantizer code into independent parts

mdouze:
Summary: Splits the ScalarQuantizer code into parts so that the AVX2 and AVX512 can be compiled independently.

Differential Revision: D73037185


alexanderguzhva:
`https://github.com/facebookresearch/faiss/pull/4296/files#diff-6b8e2e1bd2570626453d39cccd202853ebf5306d63aa04bf0ab5ed1d4ab54e9c` should not be included

alexanderguzhva:
it is worth having an API that controls `SIMDConfig::level`, at least to evaluate the performance of various SIMD implementations in benchmarks

mdouze:
superseded by #4309

PR #4298, opened 2025-04-16
Added option to build GPU libraries separately

stemann:
In some cases, it might be preferable to have GPU support as an optional add-on feature, e.g., in Julia. This PR adds a cmake option to enable the GPU library, and the GPU C library, to be built as separate libraries.

This enables a consumer to `dlopen` (`libfaiss_c`, and) `libfaiss`, and if, e.g., CUDA is available to also load the CUDA (`libfaiss_gpu_c`, and) `libfaiss_gpu`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,7 @@ option(FAISS_ENABLE_MKL "Enable MKL." ON)
 option(FAISS_ENABLE_PYTHON "Build Python extension." ON)
 option(FAISS_ENABLE_C_API "Build C API." OFF)
 option(FAISS_ENABLE_EXTRAS "Build extras like benchmarks and demos" ON)
+option(FAISS_BUILD_GPU_SHARED_LIBS "Build GPU shared libraries." OFF)
 option(FAISS_USE_LTO "Enable Link-Time optimization" OFF)
 
 if(FAISS_ENABLE_GPU)
--- a/c_api/gpu/CMakeLists.txt
+++ b/c_api/gpu/CMakeLists.txt
@@ -3,7 +3,15 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-target_sources(faiss_c PRIVATE
+if(FAISS_BUILD_GPU_SHARED_LIBS)
+  add_library(faiss_gpu_c)
+  target_link_libraries(faiss_gpu_c PRIVATE faiss_gpu)
+  set(FAISS_GPU_C faiss_gpu_c)
+else()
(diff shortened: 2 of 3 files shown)


stemann:
Current PR is based on 1.10 - should of course be rebased.

stemann:
It might make more sense to name the libraries `libfaiss_cuda` and `libfaiss_rocm` than `libfaiss_gpu`.

mnorris11:
Sorry, I didn't quite get this part:

> With the current build set-up, where the GPU objects are statically linked into libfaiss, a consumer might dlopen the non-CUDA libfaiss_c, and in turn the non-CUDA libfaiss. Then when loading the CUDA libfaiss_c, the dynamic linker will re-use the non-CUDA libfaiss and fail to resolve CUDA/GPU symbols in libfaiss:

Why can't the user just load the CUDA libfaiss_c if they are using GPU, i.e. skip the first step?

stemann:
The reason the user can't just load the CUDA libfaiss_c is a combination of:

* It is preferable to provide access to Faiss from Julia with CUDA as add-on functionality.
* Julia is a dynamic environment, so there is not necessarily an ahead-of-time decision to *not* load the non-GPU `libfaiss_c`.

In addition, the preferred method for adding CUDA functionality is via Julia [package extensions](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)), where the Faiss (Julia) package may be extended, at run-time, with additional functionality, e.g., CUDA.

I.e., it should be enough to add the `Faiss` Julia package to the user environment, and then if the user also adds the `CUDA` (or `CUDA_Runtime_jll`) Julia package, the CUDA functionality will automatically be made available.

mnorris11:
Hi @stemann sorry for the delay, was still discussing internally. But I see you have moved the PR to draft mode, so I will follow up after you publish again

stemann:
No worries. Please do comment on the general idea. I only moved the PR to draft as it is a bit of a rough draft.

stemann:
Ah - that sounds good.

On a related note, I've also been toying around (also in https://github.com/JuliaPackaging/Yggdrasil/pull/11057) with building the AVX2, and AVX512 versions, where it would actually be preferable to get a library with the same name, e.g. `libfaiss.so`, and just have the content be different depending on the targeted instruction set (e.g., plain `x86_64`, or `avx2`, `avx512`, ...). But no need to complicate this request/PR with that - that's a separate concern.

JAicewizard:
with #4422 you could run the CUDA version (as if no GPUs are present) without the target needing to have CUDA present. However, it would be very cool to to load cuda/rocm on demand.

PR #4299, opened 2025-04-17
fix: algorithm of spreading vectors over shards

simshi:
#3374 has a typo, my local repo has the correct fixing (find it when sync from recent v1.10), the intended fixing is:

for example, if we have `n = 10`, and `nshards = 8`

- the original buggy code
  ```c++
  size_t shard_size = (n + nshards - 1) / nshards;
  // i.e.  shard_size = (10 + 8 - 1) / 8 = 2
  size_t i0 = idx * shard_size;
  // all i0: [0, 2, 4, 6, 8, 10, 12, 14], wrong!!! 14>n
  size_t ni = std::min(shard_size, n - i0);
  // all ni: [2, 2, 2, 2, 2, 0, -2, -4], wrong!!!
  ```

- should be:
  ```c++
  size_t base_shard_size = n / nshards;
  // i.e.  shard_size = 10 / 8 = 1
  size_t i0 = idx * base_shard_size + std::min(size_t(idx), n % nshards);
  // all i0: [0, 2, 4, 5, 6, 7, 8, 9]
  size_t ni = base_shard_size;
  if (idx < n % nshards) {
      ++ni;
  }
  // all ni: [2, 2, 1, 1, 1, 1, 1, 1]
  ```


mnorris11:
Hi @simshi do you want to add a unit test for all the input values you care about for sharding? Then we know it passes them all in the future

simshi:
> Hi @simshi do you want to add a unit test for all the input values you care about for sharding? Then we know it passes them all in the future

Agreed, I tried to use mock, changed some code in IcmEncoder and GpuIcmEncoder, looks ugly. no good solution, now. :face_exhaling:

mnorris11:
> > Hi @simshi do you want to add a unit test for all the input values you care about for sharding? Then we know it passes them all in the future
> 
> Agreed, I tried to use mock, changed some code in IcmEncoder and GpuIcmEncoder, looks ugly. no good solution, now. 😮‍💨

Sorry, I am missing the issue, can we just test the GpuIcmEncoder::encode?

We should be able to create it like this right?
https://github.com/facebookresearch/faiss/blob/a3cd63fb6aa5cb910cfa7ddb388ce6559867c4af/faiss/gpu/GpuIcmEncoder.cu#L114-L123

We can have the test init an LSQ, create an GpuIcmEncoder with that LSQ, a new StandardGpuResources(), and device 0. Then test the encode?

simshi:
Added a test, no detailed checks without mock, but it can crash with old buggy implementation

PR #4301, opened 2025-04-19
feat: implement indexes, tests, demos and utilities for IMI-GPU

aaraujo-main:
Hi Folks!

In the last few years, I have been working on a project with some guys from the Department of Computer Science at UFMG (DCC/UFMG) and associates to implement and validate the Inverted Multi-Index for GPU (IMI-GPU). We recently published our work, "IMI-GPU: Inverted multi-index for billion-scale approximate nearest neighbor search with GPUs" in the Journal of Parallel and Distributed Computing, and now I come with the implemented code to contribute back to you. 

So, this PR contains all the changes I considered could be relevant after merging with the original code, solving the conflicts, aligning few stuff and making a squash commit. Below, I list the authors of the project, the links to the publication, the summary, architectural decisions, limitation, of the changes. 

Please let me know what are your thoughts for merging this. I have tried to make things as modular as possible without being invasive while working in parallel on a different repository, but it has been a time since the initial fork and I assume there might something to be improved and new good practices to be applied. But it would be great if can have this deployed without major changes and open new issues for future big improvements/refactoring. I am looking forward to find the best way merge this back to the original code :)

**Authors of the Project**
- Alan Araujo (Me)
- Willian Barreiros Jr.
- Jun Kong 
- Renato Ferreira
- George Teodoro

**Publication**
- Peer-Reviewed: https://doi.org/10.1016/j.jpdc.2025.105066
- Preprint: https://ssrn.com/abstract=4839994

**Summary**
This code implements a new `GpuIndexIMIPQ` class, the main C++ API for initializing and accessing the IMI-GPU. Along with it, there is also `GpuIndexIMI`, its abstract parent class, and the `GpuMultiIndex2`, the IMI coarse quantizer with two codebooks. The GPU implementation of IMI-GPU and its coarse quantizer are repesctively `IMIPQ` and `MultiIndex2`. In addition to the indexes, the multi-sequence algorithm, a central part of this solution can be found in the `faiss/gpu/utils/MultiSequence` files. Finally, the search steps of the original `IVFPQ` were adapted and new variations were implemented in theirs respective files for `IMIPQ`. We have tests and demos for the main new indexes and algorithms in `faiss/gpu/test` and `faiss/gpu/demos`. These demos contain the main code to reproduce our main experiments and few extra features. This includes the capability to run both IMIPQ and IVFPQ in CPU and GPU with C++, with multiple index shards/replicas and processes with OpenMP and MPI. For our experiments, we used custom bash scripts with personal environment variables, but I extracted the general examples of use and put them into `faiss/gpu/demos/scripts`.

**Architectural Decisions**
As I commented, the idea was to not be invasive. In this way, we could be fair about the comparisons between IMI-GPU and IVFPQ. This resulted in some duplication for the adapted flow from IVFPQ, For example, a possible main point of concern is the code duplicated between `GpuIndexIMI` and `GpuIndex`. Since `x` is always a continuous list of d-dimensional vectors, and IMI-GPU must work on `x` split into two sets of (d/2)-dimension sub-vectors, we decided to put put the split step into the `search` function and replicated it. But this could be improved later as by something as calling a `searchPreImpl_` that would call the current `searchImpl_` function. Anyway, we could wait the review for more details. Furthermore, you can find the distance computation in the inverted list phase duplicated for IMIPQ. Because it wasn’t so trivial to abstract the Multi-Index as a coarse quantizer for IVFPQ since we expect different output from it, and a different way of managing the precomputed codes when iterating over the inverted lists.

**Limitations**
- This implements only the C++ API, it still lacks the Python API.
- IMIPQ is currently indexing up to 2^32 inverted list, because the it is assuming a `short` (16 bits) ids for each centroid in the codebooks of the coarse quantizer, or `short2` (32 bits) pair id when considering the id of the virtual centroid. But the coarse quantizer already supports working with `idx_t` (or`int` after replicating existing code changed at some point to work only with `idx_t`) and `int2` inputs and outputs. For finishing the overall support, the multi-sequence algorithm must be reviewed, and the distance computation functions for the inverted lists must be adapted.
- Only the L2 metric and 32-bit floating point distances are supported for IMIPQ.

Thanks, and Kind Regards.


mdouze:
Hi @AlaoPrado 

Thanks for this remarkable effort!

Historically, we did not prioritize porting the IMI to GPU because a HNSW coarse quantizer offers better speed-accuracy tradeoffs (on CPU). Therefore, the  options in the current Faiss are: 
1. use a flat coarse quantizer 
2. decrease the number of centroids, which slows down the scanning phase 
3. use a CPU HNSW
4. use the Cagra graph-based implementation (with CuVS enabled)
so it could be that your GPU implementation reaches operating points that are not available before. 

I skimmed through the paper that has a comparison between IVFPQ with flat coarse quantizer and your approach. However it lacks (1) an evaluation with several numbers of centroids (2) a comparison with options 2-4 above. 

One question: on CPU the most efficient operating points for IMI are obtained by limiting the number of distances to compute (max_codes) rather than the number of centroids to visit (nprobe). This is because with IMI the inverted lists are severely imbalanced. Did you also observe this on GPU? 

Regarding the code itself, at first glance it seems that changing/adding 100 files is too much. The demos and benchmarks would be much shorter if written in Python. There is code that seems generic across different index types (eg. IMIAppend), so would need a stronger integration in Faiss.

AlaoPrado:
Hi,

Sorry for the delay. I had to pause for a while and focus on other personal priorities. But I am going to check if there is a refactor that could be done for decreasing the number of changes. In the meanwhile, I noticed the action was failing due to a missing MPI package in the runner and I expect to have it fixed now. Additionally, I removed the `scripts` folder from the `faiss/gpu/demos` for decreasing the number of files changed. I hope it helps.

The action was failing, because the runner was missing a MPI library required by a new demo for IVFPQ and IMIPQ on multiple GPUs. I just added a conditional, so the demo isn't compiled if the library isn't found.

Thanks, and Kind Regards,
Alan

mnorris11:
Hi @AlaoPrado  we're still trying to figure out what to do; meanwhile, would you be able to resolve the merge conflicts? Then I will re-run the tests to confirm MPI issue is gone.

aaraujo-main:
> Hi @AlaoPrado we're still trying to figure out what to do; meanwhile, would you be able to resolve the merge conflicts? Then I will re-run the tests to confirm MPI issue is gone.

Hi @mnorris11, OK. That would be great. Thanks for the notice. Yes, sure. I expect to have the conflicts solved by tomorrow.

aaraujo-main:
@mnorris11 The conflict was just about few headers, and I just had the time to review. Now it should be fine. Thank you.

aaraujo-main:
Hi @mnorris11, Could you please approve again? I might have fixed the problems by the following changes:

- I replaced the MPI search with specific search for MPI_CXX and make CMake to check MPI_CXX_FOUND before compiling the demo. So, even if the machine doesn't have MPI, the remaining code can be compiled.
- I updated the formatting of the code pointed by clang-format-18.

Thanks, and Kind Regards.

aaraujo-main:
Hi @mnorris11,
The GitHub Actions are not passing for a different reasons:
1) **Linux x86_64 GPU**: At some point, `TestGpuIndexIMIPQ.testAdd` started to fail. After recompiling it locally, I can reproduce for debugging. There are other 4 stests with `Subprocess aborted`, but they are passing locally.
2) **Windows x86_64 (conda)**: few error on `vecs_storage.cpp` for template code with fixed-sized buffers that are declared in the stack (e.g.: `TLoad buffer[currentDimension];`).
3) **Linux x86_64 GPU w/ cuVS**: error in `faiss/gpu/impl/CuvsIVFPQ.cu(55)`. It doesn't seem related to this branch.
4) **Linux x86_64 GPU w/ ROCm**: skipped.

I'm going to address the `TestGpuIndexIMIPQ.testAdd` and the `vecs_storage.cpp`.
Kind Regards.

aaraujo-main:
Hi @mnorris11,

The problems in TestGpuIndexIMIPQ test cases were caused by a recent change on `GpuIndex`. Since `GpuIndexIMI` needs to split its input data for adding and searching, it couldn't use the new `add` method as it is today, with the new `_ex` udpate. So, I just reimplemented the old `add` for `GpuIndexIMI` and made the regression to pass. In the future, I was thinking the split functionality (CPU or even GPU) for input vectors could be added as a configurable setting for `GpuIndexIMI` and `GpuMultiIndex2`, so they could implement a virtual method like `splitInputVectorsEnabled` to parametrized a new version of both `add` and `search` methods in `GpuIndex`. I also fixed the problem in `vecs_storage.cpp` by replacing the stack buffers with dynamically allocated versions.

Now the GitHub Actions should pass.

Thanks, and Kind Regards.

aaraujo-main:
@mnorris11 What does it mean "**imported** this pull request"? Should I do something? I just fixed the formatting issues by running `clang-format -i` on my last modified files, which I honestly forgot. I'm not sure why the GPU step failed in the last run. Thanks.

aaraujo-main:
Hi @mnorris11,
Could you please give me some feedback on why the GPU steps failed in the last GitHub action, and what are the next steps? Please let me know if there is something I could do. Thanks.

mnorris11:
Hi @AlaoPrado , very sorry for the delay on this.

Currently a main concern is about maintainability. We have never deprecated an index in Faiss, so when we merge, we commit to maintaining probably forever, so the team has to understand the benefits of the addition. The change is also fairly large to review.

Some questions to help us understand the operating points:

1. I saw you mention the "maximum inverted list length" is the same between the configurations. But how is the imbalance? You could check with a quick histogram comparing the two approaches.
2. From the graphs, it seems like it is most effective vs IVFPQ at recall@1, does that match your interpretation?
3. Were you ever able to do a comparison to `options 2-4` that Matthijs mentioned above? cuVS CAGRA is a graph based index so it will use much more memory than say IVFPQ, but it will be very accurate and fast. (You mentioned setting equivalent or lower memory usage for IMIPQ, so it is something to take into account.)

(By the way, "Importing" just means taking it into the internal system to run on the internal CI and check signals.)

aaraujo-main:
Hi @mnorris11 — no problem, and thanks for the answer. I hope we can find the best way to proceed and merge the contribution. If you prefer a more direct approach, we can schedule a meeting to go over each point.

You make a fair point: parts of the code could definitely be improved. When I started implementing this more than a year ago, I decided to duplicate some parts of the flow so we could achieve two important goals:

a. Compare the new and old approaches fairly without changing the original code.  
b. Continue developing the new implementation on a separate branch even if FAISS later changed its code architecture significantly.

I agree there are items to address. Please let me know your main concerns so we can decide the best architectural choices together. My worry was investing in a complex architecture that might be discarded later if a different approach were adopted. Note I also took the opportunity to improve a few aspects of the existing code that aren't necessarily core functionality (could be separated into different PRs if needed), for example:

I. Added more allocation types to make debugging easier.  
II. Implemented fixed-size memory for the index to avoid fragmentation (and related memory loss) when reallocating inverted lists during append.  
III. Added a static function to calculate the expected memory size for indexes.  
IV. Implemented *vecs_read and *vecs_write functions for reading the SIF1B dataset format in C++.

Although the tests involve many calculations, I created tests for all the entry points in the new indexes if I well remember (or at least for the main functionalities), so they can catch regressions and help maintainability. I followed up with the principle of delivering tested code. There are multiple scenarios for the same tests so I could exercise corner cases. Therefore, I think the main issue to address is code duplication. Again, please tell me what you think.

Since I opened the PR my coding style has matured. This is even more noticeable given the time since I started the project. Also, It is now easier for me to make larger changes with Copilot's latest features. When I suggested splitting the changes into several PRs earlier, I was trying to get some help (and possibly experiment with Copilot after initial contribution) because I'm the only developer on this and I no longer have access to the university server with high-power GPUs to replicate the experiments. I only have a personal GPU (8 GB), and I no longer have the indexes from the paper to recreate the same metrics. That's why I tried to avoid additional experiments for now. If it helps, we added a paragraph in the publication after the preprint commenting on GGNN and BANG (recent graph-based approaches). We concluded that our lower memory usage and fewer data transfers are advantages. Our throughput remained competitive compared to what I remember from those works.

Now, my responses to your points:

1) Sorry — since I don't have the original indexes, I would need to recreate them. I can prepare that if you think more experiments are necessary.  
2) The effect is strongest on Recall@1, but IMIPQ is also best for Recall@16 (note the graph is log-scaled). It is more effective in both memory and speedup, as described in the publication:

image

"For Recall@1 of about 0.14, Multi-D-ADC is 3.2× faster than IVFADC. For Recall@16, Multi-D-ADC is at least 1.66× and at most 1.9× faster than IVFADC. Also, Multi-D-ADC attains higher recall for a time-limited (1 second) search compared to IVFADC (0.62 vs 0.54). For Recall@128, Multi-D-ADC is about 1.7× faster than IVFADC for Recall@128 values below 0.5, after which the difference is reduced."

3) Yes — from what I understand, I was able to make the comparison described in point 2. In the new graphs I compare IVFPQ using 2× and 4× fewer centroids, and IMIPQ is still significantly better for Recall@1 and Recall@16, and better in quick searches for Recall@128. I don't see where the CPU HNSW option (point 3) fits in the comparison. The paper focuses on IVFPQ vs IMIPQ on GPU; IMIPQ search is fully implemented on GPU. The CPU comparison was only to justify porting IMIPQ to GPU.  Finally, I didn't pursue the experiments in point 4 because I no longer have access to the large-GPU server (my personal GPU is 8 GB), so it's not immediate. I was waiting for your feedback before setting up everything. As I mentioned, we also considered graph approaches after the preprint, if that helps.

Thanks again. We can follow up by call, or by email if you prefer to dive into the details. For example, I could navigate with you through the code changes to clarify any point.

Kind Regards,  
Alan

mnorris11:
Sorry for the long delay on this. We will not be making much change to the GPU Faiss implementation in the near term. We are partnering with cuVS folks for GPU based vector search, which is available through Faiss. They can provide a better answer about where the IMI-GPU impl could potentially fit. I have sent a message to them to let them know.

PR #4316, opened 2025-04-26
FreeBSD compatibility patch

yurivict:


mnorris11:
Thanks for this contribution!

We can import and merge this as well as future compatibility patches that you volunteer, but we will likely not be able to support FreeBSD-specific enhancements to the code ourselves.

PR #4325, opened 2025-05-01
Increase top-k limit on GPU

mnorris11:
Differential Revision: D74007435

We cannot just use ifdef USE_NVIDIA_CUVS #define GPU_MAX_SELECTION_K 16384, because we may compile such that USE_NVIDIA_CUVS but we are not presently using a cuVS index.
- Distance.cu is part of Faiss GPU code, so it hard codes to false.
- TestGpuIndexBinaryFlat also just hard codes to false, because it has some interaction with CPU BinaryFlat index. It doesn't use cuVS currently, so no need to expand that test (I think)
- We don't need to deal with GpuAutoTune for cuVS indexes for now, so it hard codes to false. This can be updated later if we want to. We would need to move the config out of protected in GpuIndex.h.
- Other call sites set it to the value of the config.use_cuvs.

### DIFF
--- a/faiss/gpu/GpuAutoTune.cpp
+++ b/faiss/gpu/GpuAutoTune.cpp
@@ -54,8 +54,10 @@ void GpuParameterSpace::initialize(const Index* index) {
         ParameterRange& pr = add_range("nprobe");
         for (int i = 0; i < 12; i++) {
             size_t nprobe = 1 << i;
-            if (nprobe >= ix->getNumLists() || nprobe > getMaxKSelection())
+            if (nprobe >= ix->getNumLists() ||
+                nprobe > getMaxKSelection(false)) {
                 break;
+            }
             pr.values.push_back(nprobe);
         }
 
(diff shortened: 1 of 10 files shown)


alexanderguzhva:
@mnorris11 are you sure that it is going to work? O_o

mnorris11:
> @mnorris11 are you sure that it is going to work? O_o

Absolutely not, which is why I wanted to run it on all the platforms at once and get some signals. Exporting from Phabricator puts the PR into non-draft mode automatically, converted it back now.

faiss/gpu/impl/L2Select.cu
```diff
@@ -256,7 +256,15 @@ void runL2SelectMin(
         } else if (k <= 1024) {
             RUN_L2_SELECT(128, 1024, 8);
 
-#if GPU_MAX_SELECTION_K >= 2048
+#if defined USE_NVIDIA_CUVS
+
+#if GPU_MAX_SELECTION_K >= 16384
```
wickedfoo: this code is Faiss classic only (it is a fused k-select + L2 distance computation)

PR #4336, opened 2025-05-08
[BUG] Copy IVF Centroids to Host for CPU Quantizer

tarang-jain:
For cuVS IVF indexes, identify if the coarse quantizer is a CPU index before adding the IVF centroids to it.


junjieqi:
@tarang-jain I'm wondering if we could add some tests for the change

Issue #4323, opened 2025-04-29
Gower distance?


dsweet99:
Gower distance [https://en.wikipedia.org/wiki/Gower%27s_distance] would be very useful, as it would make it possible to use mixed variable types (categorical, ordinal, continuous) directly (e.g., without creating embeddings beforehand).

Gower distance was mentioned previous in the comments to https://github.com/facebookresearch/faiss/issues/1299

mnorris11:
Similar to how Jaccard distance was added, we don't have plans to work on it, but if you submit a PR, we are happy to review.

Issue #1299, opened 2020-07-20
Jaccard similarity 


SidharthChandra:
Running on:
- CPU

Interface: 
- Python
 
How can I find Jaccard similarity

mdouze:
I assume this is 

jac(a, b) = sum_i min(a_i, b_i) / sum_i max(a_i, b_i)

It is not supported. I'll mark it as an enhancement. We already support other weird similarity measures.

mdouze:
TODO check if this is a similarity or dis-similarity 
(most of the code assumes all distances except INNER_PRODUCT are dis-similarities)

NinaCalvi:
has this been added?

edgBR:
Ey, this would be a good feature to have.

Even gower distance would nice to have:

https://ruivieira.dev/gower-distance.html

lvoursl:
Implemented and committed to the main branch, closing this issue :)

mansourshams:
I use langchain with jaccard similarity, it returns large numbers (200 range)

PR #4371, opened 2025-06-02
Dsweet/gowers distance

dsweet99:
Add Gower's distance as `faiss.METRIC_GOWER`.

See [Wikipedia](https://en.wikipedia.org/wiki/Gower%27s_distance#:~:text=In%20statistics%2C%20Gower's%20distance%20between,%2C%20ordinal%2C%20or%20continuous%20variables.) for a description.

Since Gower's distance combines continuous, ordinal, and categorical variables into a single distance, we need a way to represent them. This implementation wants:
- continuous: Scale your variables to [0,1], so x_i = x_i / (ub_i - lb_i), where ub_i and lb_i are the upper and lower bounds for variable i, x_i
- ordinal: Same as continuous (even though x_i is an integer now)
- categorical: Assign each category a negative integer value. For example, the categories [red, green, blue] could map to [-1, -2, -3]. The values don't matter. It only matters that you are consistent, e.g., green is -2 in all of your vectors.

There's a unit test. 

Also, I've incorporated into local code with
```
faiss.IndexFlat(num_dim, faiss.METRIC_GOWER)
```
and have had no problems so far.

This PR addresses this issue: https://github.com/facebookresearch/faiss/issues/4323
And this comment on a different issue: https://github.com/facebookresearch/faiss/issues/1299#issuecomment-1378904784

Side note: This PR was basically developed by Cursor.


faiss/utils/extra_distances-inl.h
```diff
@@ -163,6 +164,47 @@ inline float VectorDistance<METRIC_ABS_INNER_PRODUCT>::operator()(
...
+        // Skip NaN values
+        if (std::isnan(x[i]) || std::isnan(y[i])) {
+            continue;
+        }
+        
+        // Check if this is a numeric dimension (0-1) or categorical (negative)
+        if (x[i] >= 0 && y[i] >= 0) {
+            // Check if numeric values are in [0,1] range
+            if (x[i] > 1 || y[i] > 1) {
+                std::cerr << "Warning: Numeric values must be in [0,1] range for Gower distance" << std::endl;
```
mdouze: errors are reported with FAISS_THROW_IF_NOT in Faiss

faiss/python/swigfaiss.swig
```diff
@@ -514,6 +514,15 @@ void gpu_sync_all_devices()
 
 %include  <faiss/MetricType.h>
 
+%pythoncode %{
```
mdouze: please remove, this is unrelated to the new metric

dsweet99: Ok.

faiss/utils/extra_distances-inl.h
dsweet99: We can't throw exceptions from inside the omp-parallelized code, and this kind of verbosity seems inconsistent with the rest of the metrics anyway, some I'm just going to remove the text and return NaN.

(If you prefer, I can add a validation check before the distance calculation is called. In any case, I'll need to document Gower's distance elsewhere so users know how to represent their data before calling into faiss.)

mdouze: NaN is fine

mdouze:
lgtm

wickedfoo:
Can you provide an implementation for the GPU as well?

can follow other metrics such as JaccardSimilarity here:

https://github.com/facebookresearch/faiss/blob/040732fb8f059c2cdb2635fc15eb17573373f49c/faiss/gpu/impl/DistanceUtils.cuh#L254

and add the implementation in this file (can search for JaccardSimilarity or other metrics to see where they are instantiated in this file, 2 different places):

https://github.com/facebookresearch/faiss/blob/040732fb8f059c2cdb2635fc15eb17573373f49c/faiss/gpu/impl/Distance.cuh

dsweet99:
> Can you provide an implementation for the GPU as well?
> 
> can follow other metrics such as JaccardSimilarity here:
> 
> https://github.com/facebookresearch/faiss/blob/040732fb8f059c2cdb2635fc15eb17573373f49c/faiss/gpu/impl/DistanceUtils.cuh#L254
> 
> and add the implementation in this file (can search for JaccardSimilarity or other metrics to see where they are instantiated in this file, 2 different places):
> 
> https://github.com/facebookresearch/faiss/blob/040732fb8f059c2cdb2635fc15eb17573373f49c/faiss/gpu/impl/Distance.cuh

"Can" is key here. I've never written GPU code before (.cuh ? .huh?) and I don't own a GPU.

"Could", given time, yes, but "can" -- not really.

dsweet99:
I spoke too soon!

While *I* don't know GPU programming, LLM's do. And I apparently I have an on-board CUDA-compatible GPU.
The code is pushed. 

I'm not sure whether it matters but I used
- Quadro T1000 Mobile
- CUDA v12.5.39

dsweet99:
Is a merge a possibility?

mnorris11:
Hi @dsweet99 , it seems like the cuVS test fails. Can we add a skip if cuVS is not enabled? Example:
```
@unittest.skipIf(
    "CUVS" not in faiss.get_compile_options(),
    "only if CUVS is compiled in")
```
From https://github.com/facebookresearch/faiss/blob/3ad3eec575f6866c309a3270c9000db1f0822bb0/faiss/gpu/test/test_cuvs.py#L13-L15

dsweet99:
>  it seems like the cuVS test fails.
Which test?

(I compiled faiss w/GPU support but not cuVS support and run all of the Python tests. I get either PASSED or SKIPPED.)

mnorris11:
> > it seems like the cuVS test fails.
> > Which test?
> 
> (I compiled faiss w/GPU support but not cuVS support and run all of the Python tests. I get either PASSED or SKIPPED.)

do_test_gower that you added

Are the logs viewable for you here? https://github.com/facebookresearch/faiss/actions/runs/15834176641/job/44634893463

I think skipping it if cuVS is enabled is enough; we don't need to worry about supporting it for cuVS, just have to make sure tests pass.

dsweet99:
I see the logs now. I added the decorator:
```
@unittest.skipIf(
    "CUVS" in faiss.get_compile_options(),
    "only if CUVS is not compiled in")
 ```

PR #4378, opened 2025-06-05
Add cuVS filter conversion utility

lowener:
cuVS is using a different filter type than FAISS. This PR will add conversion function from `FAISS::IDSelector` to `cuvs::core::bitset_view` to allow users to do pre-filtering on their searches.

### DIFF
--- a/faiss/gpu/utils/CuvsFilterConvert.cu
+++ b/faiss/gpu/utils/CuvsFilterConvert.cu
@@ -32,14 +32,7 @@
 
 namespace faiss::gpu {
 
-/**
- * @brief CUDA kernel to set a range of bits in a bitset to true
- *
- * @param bitset_data Pointer to the bitset data
- * @param imin Starting index
- * @param imax Ending index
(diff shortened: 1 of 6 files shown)


faiss/gpu/utils/CuvsFilterConvert.cu
```diff
@@ -0,0 +1,257 @@
...
+#include <omp.h>
+#include <raft/core/host_mdarray.hpp>
+#include <thrust/for_each.h>
+#include <raft/core/bitset.cuh>
+#include <raft/core/copy.cuh>
+
+namespace faiss::gpu {
+
+/**
+ * @brief CUDA kernel to set a range of bits in a bitset to true
```
mnorris11: (nit) the existing Faiss function comments leave out the `@brief` and just start the comment like this, can we switch to that? Otherwise LGTM
```
/** CUDA kernel to...
```

mnorris11:
Thanks @lowener for your contribution! 

Could you add a small description of what problem this solves to help reviewers understand?

PR #4384, opened 2025-06-10
cuVS Cagra FP16 support

jinsolp:
Supporting fp16 for cuVS cagra, and introducing new extended APIs for this.
Discussions related to this issue: https://github.com/facebookresearch/faiss/issues/4324

Added tests in `faiss/gpu/test/TestGpuIndexCagra.cu` and `faiss/gpu/test/test_cagra.py` for example usage.

PR #4188, opened 2025-02-13
Add the support for IndexIDMap with Cagra index

navneet1v:
## Description
Add the support for adding vectors with ids when IndexIDMap is used with Cagra Index.

Resolves issue: https://github.com/facebookresearch/faiss/issues/4107

PR #4411, opened 2025-06-25
Add support for IndexIDMap with Cagra fp16

jinsolp:
This PR does 2 things
- Enable support fot `IndexIDMap` with Cagra fp16 (original support introduced in https://github.com/facebookresearch/faiss/pull/4188)
    - Added tests in `test_cagra.py`
- Reflecting feedback about python API from https://github.com/facebookresearch/faiss/pull/4384#discussion_r2167013214

### DIFF
--- a/faiss/python/class_wrappers.py
+++ b/faiss/python/class_wrappers.py
@@ -42,6 +42,14 @@ def _check_dtype_uint8(codes):
                         " uint8, but found %s" % ("codes", codes.dtype))
     return np.ascontiguousarray(codes)
 
+def _np_type_to_faiss_numeric(np_type):
+    if np_type == np.float32:
+        return faiss.Float32
+    elif np_type == np.float16:
+        return faiss.Float16
+    else:
+        raise TypeError("Input data must be ndarray of dtype "
+                        " float32 or float16, but found %s" % (np_type))
 
 def replace_method(the_class, name, replacement, ignore_missing=False):
     """ Replaces a method in a class with another version. The old method
@@ -211,7 +219,7 @@ def handle_NSG(the_class):
 
 def handle_Index(the_class):
 
-    def replacement_add(self, x, numeric_type = faiss.Float32):
+    def replacement_add(self, x):
         """Adds vectors to the index.
         The index must be trained before vectors can be added to it.
         The vectors are implicitly numbered in sequence. When `n` vectors are
@@ -226,13 +234,11 @@ def handle_Index(the_class):
 
         n, d = x.shape
         assert d == self.d
-        if numeric_type == faiss.Float32:
-            x = np.ascontiguousarray(x, dtype='float32')
-        else:
-            x = np.ascontiguousarray(x, dtype='float16')
+        numeric_type = _np_type_to_faiss_numeric(x.dtype)
+        x = np.ascontiguousarray(x, dtype=x.dtype)
         self.add_c(n, swig_ptr(x), numeric_type)
 
-    def replacement_add_with_ids(self, x, ids, numeric_type = faiss.Float32):
+    def replacement_add_with_ids(self, x, ids):
         """Adds vectors with arbitrary ids to the index (not all indexes support this).
         The index must be trained before vectors can be added to it.
         Vector `i` is stored in `x[i]` and has id `ids[i]`.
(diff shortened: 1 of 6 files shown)


jinsolp:
One of the issues that I've noticed is that the current extended API approach results in a lot of warnings from all child structs at compile time. This is because we have an overload of a function (e.g. `add(n, x)` and `add(n, x, faiss.Float16)`) in the base struct (`Index.h`), but the child structs only override one of these overloaded functions (because most of them don't bother about the latter `add`).

The straightforward solution is to add `using Index::add;` to every single child struct, but this seems very tedious... (and we will have to do it for every extended-api function)
Any suggestions to tackling this??

faiss/python/class_wrappers.py
```diff
@@ -223,14 +228,16 @@ def replacement_add(self, x, numeric_type = faiss.Float32):
...
+        numeric_type = _np_type_to_faiss_numeric(x.dtype)
+        
         if numeric_type == faiss.Float32:
             x = np.ascontiguousarray(x, dtype='float32')
-        else:
+            self.add_c(n, swig_ptr(x))
+        else:   # fp16
             x = np.ascontiguousarray(x, dtype='float16')
-        self.add_c(n, swig_ptr(x))
+            self.add_c(n, swig_ptr(x), faiss.Float16)
```
mdouze: The problem is that if we try to call the add with a float16 array on an index that does not support it, we'll get a RuntimeError.
Workarounds for this (ordered from most to least desireable): 
- do the float16 -> float32 conversion in the C++ add with type tags 
-  leave as it (need to specify the numeric_type in python)
- catch the exception and retry converting to float32
- add some flag to the class that announces whether  float16 is supported.

mdouze (changes requested):
For this change to work without user visible degradation, we should implement implicit conversion fp16->fp32 in C++ for add/search/train. There are a few fp16 routines for CPU here: https://github.com/facebookresearch/faiss/blob/main/faiss/utils/fp16.h

faiss/python/class_wrappers.py
jinsolp: > The problem is that if we try to call the add with a float16 array on an index that does not support it, we'll get a RuntimeError.

Isn't this the desired behavior though? To raise a runtime error saying that the type isn't supported if the user attempts to use it?

mdouze: This is acceptable. It is just a bit counter-intuitive that the fp16 flag needs to be specified explicitly when the input array is fp16.

jinsolp:
@mdouze Thanks for your feedback! I have made a few changes : )
- Renamed all extended APIs with the `Ex` postfix because of cpp compile warnings related to not overriding overloaded functions.
- Added the `numeric_type` tag support for `torch_utils.py`

mdouze:
I don't understand the rationale for the `Ex` suffix. Where does the cpp compiler issue warnings?

jinsolp:
@mdouze In the current CI runs for cuVS build, it can be seen that it raises a warning for every single index https://github.com/facebookresearch/faiss/actions/runs/16378195325/job/46284306200

e.g. `overloaded virtual function "faiss::Index::train" is only partially overridden in class "faiss::IndexPQ"`

I've stated the reason for this in the comment above

> One of the issues that I've noticed is that the current extended API approach results in a lot of warnings from all child structs at compile time. This is because we have an overload of a function (e.g. add(n, x) and add(n, x, faiss.Float16)) in the base struct (Index.h), but the child structs only override one of these overloaded functions (because most of them don't bother about the latter add).
The straightforward solution is to add using Index::add; to every single child struct, but this seems very tedious... (and we will have to do it for every extended-api function)

TLDR; warning happen because most child structs don't override the overloaded functions (i.e. most child structs don't override the extended API).

There are two ways to tackle this cpp warning
1. add `using` in every single index struct so that it can see the base `Index` struct's overloaded functions with numeric types (the extended functions)
- pro: we have the same name: the original `add` and the `add` with `void*` and `numeric_type`.
- con: we have to change every single index in FAISS
2. change the name of the extended functions so that they are not an "overload". i.e. instead of having two `add` functions with different signatures, just have `add` and `addEx`
- pro: we only need to change a few files where we are using the extended API
- con: we have the `*Ex` postfix for extended API

mnorris11:
Sorry for delayed responses. At least to me, it does not seem like a problem to annotate this 1 line on all the indexes that require it, if the API stays as `add()`, etc. It looks cleaner and keeps with the current pattern.

jinsolp:
Got it, thanks @mnorris11 , I'll fix this today and push : )

jinsolp:
@mnorris11 Now the diff is really big, but we are 1) keeping the previous API (without the `Ex`) and 2) compiling without warnings.

swig wasn't able to process macros or work with `using`. So what I've done is manually override functions if the child struct is already overriding its original counterpart. (i.e. if the child class is already overriding `add()` I also overrode `add(void*, numeric_type)`. 
All overrode functions are just forwarded to its base struct's function to keep the logic consistent as before.

Issue #3936, opened 2024-10-10
`faiss` fails to import on python==3.12 & `aarch64-linux` because of deprecated `numpy.distutils`


GaetanLepage:
# Summary

To know whether the system supports SVE, [faiss uses deprecated `numpy.distutils.cpuinfo`](https://github.com/facebookresearch/faiss/blob/d243e628880676332263347817b3fe7f474b8b5b/faiss/python/loader.py#L43-L44).
This has been removed and crashes on Python 3.12 (on `aarch64-linux` systems) with:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<string>", line 1, in <lambda>
  File "/nix/store/6iq3nhgdyp8a5wzwf097zf2mn4zyqxr6-python3-3.12.5/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/nix/store/x3g3x2s64qwsrrrg6cpfc84zycc4hlbh-python3.12-faiss-1.9.0/lib/python3.12/site-packages/faiss/__init__.py", line 16, in <module>
    from .loader import *
  File "/nix/store/x3g3x2s64qwsrrrg6cpfc84zycc4hlbh-python3.12-faiss-1.9.0/lib/python3.12/site-packages/faiss/loader.py", line 88, in <module>
    instruction_sets = supported_instruction_sets()
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/x3g3x2s64qwsrrrg6cpfc84zycc4hlbh-python3.12-faiss-1.9.0/lib/python3.12/site-packages/faiss/loader.py", line 53, in supported_instruction_sets
    if is_sve_supported():
       ^^^^^^^^^^^^^^^^^^
[...]

PR #4416, opened 2025-07-01
Add workaround to check SVE support when numpy.distutils is not available

Kh4L:
Fix for https://github.com/facebookresearch/faiss/issues/3936

### DIFF
--- a/faiss/python/loader.py
+++ b/faiss/python/loader.py
@@ -42,7 +42,7 @@ def supported_instruction_sets():
         try:
             import numpy.distutils.cpuinfo
             return "sve" in numpy.distutils.cpuinfo.cpu.info[0].get('Features', "").split()
-        except (ModuleNotFoundError, ImportError):
+        except ImportError:
             # check if SVE is supported by checking the auxval
             # using values defined as:
             # #define AT_HWCAP 16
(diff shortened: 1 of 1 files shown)


mnorris11:
Thanks for the contribution!

We are looking into the windows fix, then will re-run the CI.

faiss/python/loader.py
```diff
@@ -39,8 +39,15 @@ def is_sve_supported():
         if Version(numpy.__version__) >= Version("2.0"):
             return False
         # platform-dependent legacy fallback using numpy.distutils.cpuinfo
-        import numpy.distutils.cpuinfo
-        return "sve" in numpy.distutils.cpuinfo.cpu.info[0].get('Features', "").split()
+        try:
+            import numpy.distutils.cpuinfo
+            return "sve" in numpy.distutils.cpuinfo.cpu.info[0].get('Features', "").split()
+        except:
```
mnorris11: Can you make this specific to `ModuleNotFoundError`? (our linter is flagging the bare `except`)

Kh4L: sure, updated the PR

mnorris11: @Kh4L , sorry for the back and forth: now it says:
```
Redundant exception types in except (ModuleNotFoundError, ImportError):.  Write except ImportError:, which catches exactly the same exceptions.
```
After changing to that, should be good. (I would do it myself but we cannot import a PR then make changes)

Kh4L: just updated the PR, hopefully it passes the linter now

PR #4421, opened 2025-07-05
Use c++11 only in headers

JAicewizard:
This is a small patch, I understand that you would rather use c++17 as in the rest of the codebase.

I am currently in a situation where I need to include c++11 headers, and c++17 headers (faiss). Since this is just a small patch I am hoping you would accept this! However, if you would rather keep using c++17 features in the headers, I can keep this as downstream patch for myself.

### DIFF
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -78,7 +78,7 @@ bool read_vector_base(
         const std::optional<size_t> beforeknown_size,
         const std::optional<size_t> size_multiplier) {
     // check if the use case is right
-    if constexpr (is_maybe_owned_vector_v<VectorT>) {
+    if constexpr (is_maybe_owned_vector_v(VectorT)) {
         // is it a mmap-enabled reader?
         MappedFileIOReader* mf = dynamic_cast<MappedFileIOReader*>(f);
         if (mf != nullptr) {
--- a/faiss/impl/maybe_owned_vector.h
+++ b/faiss/impl/maybe_owned_vector.h
@@ -295,8 +295,7 @@ struct is_maybe_owned_vector : std::false_type {};
 template <typename T>
 struct is_maybe_owned_vector<MaybeOwnedVector<T>> : std::true_type {};
 
-template <typename T>
-inline constexpr bool is_maybe_owned_vector_v = is_maybe_owned_vector<T>::value;
+#define is_maybe_owned_vector_v(T) is_maybe_owned_vector<T>::value
 
 template <typename T>
 bool operator==(


ramilbakhshyiev:
Hi @JAicewizard, thank you for the contribution, I do believe it would be better to keep it consistent and use C++17 features across the board. However, I want to ensure this is not a wide-spread issue. Do you have to stick to C++11 because of your codebase / development environment? Thank you!

JAicewizard:
Hello! Yes I have to keep using c++11, as I have to integrate it into a c++11 environment. Long-term I think I can split the project up into a c++11 and 17 side, with a c api in between, and for the short term just keep this patch myself downstream. So this is not a deal-breaker for me, just wanted to post it in case you were interested.

ramilbakhshyiev:
> Hello! Yes I have to keep using c++11, as I have to integrate it into a c++11 environment. Long-term I think I can split the project up into a c++11 and 17 side, with a c api in between, and for the short term just keep this patch myself downstream. So this is not a deal-breaker for me, just wanted to post it in case you were interested.

Got it, thank you @JAicewizard for the explanation and for caring to offer this!

ramilbakhshyiev:
Hi @JAicewizard, @mdouze recommended to consider importing this change. I rebased this PR and imported to see how if it succeeds internally or breaks things. We can re-assess based on those results. If we decide to import it in the end, I think it would be great to add a comment describe the need for this change.

ramilbakhshyiev:
Rebased and reimported to see if the updated version succeeds.

alexanderguzhva:
@JAicewizard 
Please change the code in a header into something like the following:

For `faiss/impl/maybe_owned_vector.h`:
```C++
...

template <typename T>
struct is_maybe_owned_vector : std::false_type {};
template <typename T>
struct is_maybe_owned_vector<MaybeOwnedVector<T>> : std::true_type {};

// guard with c++-17 (maybe, it is available somewhere in faiss/impl/platform.h?)
#if ((defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L)

template <typename T>
inline constexpr bool is_maybe_owned_vector_v = is_maybe_owned_vector<T>::value;

#endif

...
```

if C++11 is required for `faiss/impl/index_read.cpp`, you may also use `if (is_maybe_owned_vector<VectorT>::value)` and get rid of `constexpr` check using a traditional `SFINAE` mechanism

alexanderguzhva:
@ramilbakhshyiev fyi

ramilbakhshyiev:
> @JAicewizard Please change the code in a header into something like the following:
> 
> For `faiss/impl/maybe_owned_vector.h`:
> 
> ```c++
> ...
> 
> template <typename T>
> struct is_maybe_owned_vector : std::false_type {};
> template <typename T>
> struct is_maybe_owned_vector<MaybeOwnedVector<T>> : std::true_type {};
> 
> // guard with c++-17 (maybe, it is available somewhere in faiss/impl/platform.h?)
> #if ((defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L)
> 
> template <typename T>
> inline constexpr bool is_maybe_owned_vector_v = is_maybe_owned_vector<T>::value;
> 
> #endif
> 
> ...
> ```
> 
> if C++11 is required for `faiss/impl/index_read.cpp`, you may also use `if (is_maybe_owned_vector<VectorT>::value)` and get rid of `constexpr` check using a traditional `SFINAE` mechanism

Thanks @alexanderguzhva! @JAicewizard, please lmk if you're willing to make and test these changes? I can re-import and test a revision.

JAicewizard:
Hello @ramilbakhshyiev, sorry vacation and other activities got in the way, thank you for considering this! Having c++17 code in the c files is not a problem since that will not be included in the c++11 code.

JAicewizard:
Not sure why CI is failing, seems to be unrelated to this PR?

ramilbakhshyiev:
I rebased it and retried the tests, let's see if it succeeds

PR #4422, opened 2025-07-05
Add option to link cuda statically

JAicewizard:
When linking to cuda dynamically, and the target does not have cuda installed, the user can not even start code with FAISS compiled with cuda. Therefore it could be useful to link the CUDA runtime/blas statically instead to avoid requiring CUDA on the users end.

### DIFF
--- a/faiss/gpu/CMakeLists.txt
+++ b/faiss/gpu/CMakeLists.txt
@@ -169,6 +169,15 @@ set(FAISS_GPU_HEADERS
   utils/warpselect/WarpSelectImpl.cuh
 )
 
+set(CUDA_LIBS CUDA::cudart CUDA::cublas)
+if(FAISS_GPU_STATIC)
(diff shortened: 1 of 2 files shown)


faiss/gpu/CMakeLists.txt
```diff
@@ -169,6 +169,15 @@ set(FAISS_GPU_HEADERS
   utils/warpselect/WarpSelectImpl.cuh
 )
 
+set(CUDA_LIBS CUDA::cudart CUDA::cublas)
+if(FAISS_GPU_STATIC)
+	if(FAISS_ENABLE_ROCM)
```
mnorris11: Our linter is flagging this as having tab literals, could you update to 2 spaces?

JAicewizard: Yess, fixed!

Issue #4430, opened 2025-07-14
CAGRA to support IndexIDMap for binary and int8 types


cjnolet:
cc @jinsolp @tarang-jain

PR #4439, opened 2025-07-17
`int8` support for cuVS cagra

jinsolp:
Closes https://github.com/facebookresearch/faiss/issues/4430 by exposing int8 for cuVS cagra.
This PR (https://github.com/facebookresearch/faiss/pull/4411) needs to be merged before this one .

### DIFF
--- a/faiss/gpu/GpuCloner.cpp
+++ b/faiss/gpu/GpuCloner.cpp
@@ -94,7 +94,7 @@ Index* ToCPUCloner::clone_Index(const Index* index) {
 #if defined USE_NVIDIA_CUVS
     else if (auto icg = dynamic_cast<const GpuIndexCagra*>(index)) {
         IndexHNSWCagra* res = new IndexHNSWCagra();
-        if (icg->get_numeric_type() == faiss::NumericType::Float16) {
+        if (icg->get_numeric_type() != faiss::NumericType::Float32) {
             res->base_level_only = true;
         }
         icg->copyTo(res);
--- a/faiss/Index.h
+++ b/faiss/Index.h
@@ -61,6 +61,8 @@ struct DistanceComputer;
 enum NumericType {
     Float32,
     Float16,
+    UInt8,
+    Int8,
 };
 
 inline size_t get_numeric_type_size(NumericType numeric_type) {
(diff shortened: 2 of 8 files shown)


faiss/gpu/GpuCloner.cpp
```diff
@@ -94,7 +94,7 @@ Index* ToCPUCloner::clone_Index(const Index* index) {
 #if defined USE_NVIDIA_CUVS
     else if (auto icg = dynamic_cast<const GpuIndexCagra*>(index)) {
         IndexHNSWCagra* res = new IndexHNSWCagra();
-        if (icg->get_numeric_type() == faiss::NumericType::Float16) {
+        if (icg->get_numeric_type() != faiss::NumericType::Float32) {
```
mnorris11, line 97: For my learning, what happens in the 16 and int8 case?

jinsolp: This is for [`copyTo`](https://github.com/facebookresearch/faiss/blob/71f775b7ea927884e02a522548cbb6c88f354502/faiss/gpu/GpuIndexCagra.cu#L431-L439) where we copy from `GpuIndexCagra` to `IndexHNSWCagra`.
If `base_level_only=false`, we end up calling [`hnsw_add_vertices`](https://github.com/facebookresearch/faiss/blob/71f775b7ea927884e02a522548cbb6c88f354502/faiss/IndexHNSW.cpp#L55), which only support fp32 computations on CPU, so we need it to be true. 
When `base_level_only=true`, we can use the `storage`(which is `IndexScalarQuantizer`)'s `add()` instead. : )

jinsolp:
Added logic for encode/decode for interop between cpu and gpu. Previously, it was directly copying the data bits causing issues between int8 and uint8.

For example, in the `copyTo` function, it was calling [`add_sa_codes`](https://github.com/jinsolp/faiss/blob/61341a3222f0909b8c9f950181fa3a026a502736/faiss/gpu/GpuIndexCagra.cu#L531), which [copies the bits as-is](https://github.com/jinsolp/faiss/blob/61341a3222f0909b8c9f950181fa3a026a502736/faiss/IndexFlatCodes.cpp#L44). 
When we call search on the copied cpu index, the [scalar quantizer encodes the vectors](https://github.com/jinsolp/faiss/blob/2f577aef3be6706a2d0527f5ca403f44b740624a/faiss/impl/ScalarQuantizer.cpp#L835-L851). Thus we end up having different representation of the trained vectors copied over from the GPU, and the query vectors. 

The latest commit accounts for this situation and does the encoding/decoding step as part of the copyTo and copyFrom functions.

PR #4486, opened 2025-07-30
Enable web environment

grf53:
## Description
This PR enables FAISS to run in a web environment.

### OpenMP
Use dummy OpenMP with functions like `int omp_get_num_threads(void) { return 1; }` etc.

### BLAS, LAPACK
BLAS and LAPACK is written in Fortran, so they also need to be built for wasm32 target.
We will use [emscripten](https://emscripten.org/) to build faiss, so the target triple would be `wasm32-unknown-emscripten`.
To do this, we need 'LLVM flang' with target `wasm32-unknown-emscripten`.

#### Flang
Flang is, simply put, the LLVM version of gfortran.
You can get flang anywhere you want. But building LLVM by yourself is a firm, consistent way to have flang.
(Usual LLVM package doesn't contain flang as default.)
Here's a build command to build LLVM with flang for the target we want.
```
# Build
cmake -G Ninja -S ${LLVM_ROOT}/llvm -B ${BUILD_PATH} \
  -DCMAKE_INSTALL_PREFIX=llvm-18.1.1 \
  -DCMAKE_BUILD_TYPE=MinSizeRel \
  -DLLVM_DEFAULT_TARGET_TRIPLE="wasm32-unknown-emscripten" \
  -DLLVM_TARGETS_TO_BUILD="WebAssembly" \
  -DLLVM_ENABLE_PROJECTS="clang;flang;mlir

# Add location of LLVM tools in the `PATH`, if you want.
export PATH=${BUILD_PATH}/bin:$PATH
```

#### Build `libblas.a`, `liblapack.a`
./build_libblas.sh    # build libblas.a
./build_liblapack.sh  # build liblapack.


INSTALL.md
```diff
@@ -316,3 +316,86 @@ and you can run
...
+Add built binaries to path
+```bash 
+export PATH=${BUILD_PATH}/bin:$PATH
+```
+
+### Build BLAS/LAPACK static libraries
+Get LLVM Source
+```bash
+git clone -b enable-web https://github.com/grf53/faiss-web.git
+cd faiss-web
```
grf53, line 357: If this PR can be merged, Let me modify these lines before that.

mnorris11:
Hi @grf53 , thanks for the contribution!

Can you help me understand what you mean by run in web environment? This is like being called from Javascript or something?

grf53:
> Hi @grf53 , thanks for the contribution!
> 
> Can you help me understand what you mean by run in web environment? This is like being called from Javascript or something?

Hi, @mnorris11!
Yes, exactly. The changes in this PR are made to enable FAISS to run via javascript in a web browser.
You can run a javascript code like this with an appropriate binding from c++(embind).
To build the binding, you need `libfaiss.a`, `libblas.a`, and `liblapack.a` built for wasm32 target.

```javascript
const { FaissIndex, VectorFloat } = Module;
const idx = new FaissIndex(3);

const addVector = (data) => {
  const vec = new VectorFloat();
  data.forEach(v => vec.push_back(v));
  idx.add(vec);
  vec.delete();
};

addVector([1,0,0]);
addVector([0,1,0]);
addVector([0,0,1]);
addVector([1,1,1]);

const k = 3;
const queryData = [1, 0.9, 0.0];

console.log(`Searching for ${k} nearest neighbors to [${queryData.join(', ')}]...`);

const query_vec = new VectorFloat();
queryData.forEach(v => query_vec.push_back(v));

const res = idx.search(query_vec, k);
```

mdouze:
Thanks for this contribution. The platform seems like an interesting addition. 
The build scripts look like elegant/acrobatic engineering. 
It would be interesting to see the scope of this addition, ie. is it mainly a POC? 
- is the runtime performance worth it, ie. is it better than a brute force loop in javascript? 
- how can we test it? is it possible to write a CI that validates that it compiles and runs correctly?
- will you be able to maintain it over the next say 2 years if the CI breaks? 

Having CI is not strictly required, since apart from the cmakefile the PR does not change much of the code, but it would definitely be a plus for us to accept the PR.

grf53:
@mdouze Hello, thank you for your interest and I am sorry for my late response.

> It would be interesting to see the scope of this addition, ie. is it mainly a POC?

I would say the contribution level of this PR is clearly closer to a PoC. By focusing on enabling, I willingly omitted multi-threading(with OpenMP) and GPU uses which are FAISS's strengths. And there were only future expansion potential, but no specific attempts at testing or sustainability.
But actually, I tried this not for just an interesting experiment, but an actually utilization. I'm working on building an abstracted vector store with an embedding model that works on the web, and this is a part of it.

> * is the runtime performance worth it, ie. is it better than a brute force loop in javascript?

Regarding performance, we can compare it to a JavaScript brute force loop in two qualitative aspects. I apologize for the lack of quantitative experiments.

1) FAISS is a proven library for applying various indexing techniques, such as IVF and HNSW. Additionally, it can optimize vector comparisons by applying PQ, which can result in higher performance than a brute force approach in the same environment.
2) Vector indexing and search algorithms, including FAISS, involve significant computation. Compiled WASM-based implementations are generally faster than implementations using pure JavaScript (or TypeScript).

Furthermore, there is the potential for further performance improvements as web execution environments evolve.
1) [Emscripten](https://emscripten.org/docs/porting/simd.html) allows the use of various SIMD operations. Since some of these are currently supported, it's worth trying soon, but even if they are implemented, it's likely to require separate changes.
2) WebGPU is a technology that was proposed in 2021 and is expected to become more widespread as it becomes a Candidate Recommendation (CR) in 2025. FAISS, which utilizes GPUs, boasts even more impressive performance, so if WebGPU support is expanded, it will be able to be effectively utilized in web-based environments.

> * how can we test it? is it possible to write a CI that validates that it compiles and runs correctly?
> * will you be able to maintain it over the next say 2 years if the CI breaks?

Since emscripten can easily run the built results via emrun, we can run the test code in CI via a headless browser using the command below.
```
$ emcc tests/*.cpp -o test.html --emrun -s ALLOW_MEMORY_GROWTH=1 -s EXIT_RUNTIME=1 -O2
$ emrun --browser=/usr/bin/google-chrome-stable --browser_args="--headless --no-sandbox --disable-dev-shm-usage" test.html
```

I could try to integrate the existing test codes written in C++ with gtest and CI runs them, but I'm not sure if it's necessary to write separate tests specifically for the web environment, other than the existing ones.

I think I could add CI and maintain it if we could effectively leverage the existing C++ test code. Otherwise, it would be a big challenge for me to add separated tests for web environment that cover all the current level of coverage. I think I need some time to check it out.

If we perform that process in practice, should it be included in this PR?

> Having CI is not strictly required, since apart from the cmakefile the PR does not change much of the code, but it would definitely be a plus for us to accept the PR.

If we can proceed without CI for this PR, then let me know what kind of things I need to do more to tidy up these changes to be ready to be in.

Thanks.

mnorris11:
Hi @grf53, apologies for the slow response. After discussion, we decided to not merge this one right now since it is more of a POC. However, the PR will be around still, and if other users require this functionality, they can be pointed to this PR for a framework of what to do.

PR #4515, opened 2025-08-07
SIMD optimization RaBitQ

alibeklfc:
Summary:
This diff introduces a new file rabitq_simd.h with multiple SIMD-optimized implementations of the dot product calculation using population count (popcnt) operations:

 1. AVX-512 implementation with AVX512VPOPCNTDQ: Processes data in 512-bit (64-byte) chunks using dedicated AVX-512 popcnt instructions, with fallbacks to smaller vector sizes for remaining data.
2. AVX-512 fallback implementation without AVX512VPOPCNTDQ: Uses AVX512F instructions with a lookup-based popcount method for 512-bit vectors, falling back to smaller vectors for remaining data.
    
3.  AVX2 implementation: Uses a lookup-based popcount method with 256-bit (32-byte) AVX2 instructions, handling leftovers with 128-bit SSE operations and scalar processing.
    
4.  Scalar fallback: Processes data in 64-bit chunks with builtin popcount operations for systems without SIMD support.

Differential Revision: D79301607

### DIFF
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -7,15 +7,15 @@
 
 #include <faiss/impl/RaBitQuantizer.h>
 
+#include <faiss/impl/FaissAssert.h>
+#include <faiss/utils/distances.h>
+#include <faiss/utils/rabitq_simd.h>
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


alexanderguzhva:
looks good

PR #4523, opened 2025-08-08
Upgrade Faiss OSS side to numpy2

mnorris11:
Differential Revision: D79900113

This upgrades OSS Faiss to numpy 2.3.3. I tried to fix issues based on https://numpy.org/devdocs/numpy_2_0_migration_guide.html. 

Overall, numpy2 causes some test issues on various platforms/optimization levels.

Test changes that are OK:
- test_contrib.py: test_float: changing to use item(): this is just the new way of getting the value, like recommended in the migration guide.

Test changes that indicate some problem:
- test_contrib.py: test_ivf_train_2level: This fails on many platforms (avx2, conda installs). I could not find what the issue was. I had to increase the bounds slightly.
- test_index_accuracy.py: test_OIVFPQ: This fails just on AVX512. I had to increase the iterations from 10 to 12.
- test_local_search_quantizer.py and test_standalone_codec.py: Some of these break on Windows. Others needed lighter tolerances.

Debugging attempts: Especially for test_ivf_train_2level, I tried to unsuccessfully track down what might have changed. I suspected the type promotion change from numpy1 to 2. 
```
- np.float32(3) + 3. now returns a float32 when it previously returned a float64.
- np.array([3], dtype=np.float32) + np.float64(3) will now return a float64 array. (The higher precision of the scalar is not ignored.)
```
I changed all datasets to be explicitly float64s, and checked for any constants in the code path. No luck.

### DIFF
--- a/.github/actions/build_cmake/action.yml
+++ b/.github/actions/build_cmake/action.yml
@@ -36,10 +36,10 @@ runs:
         conda config --set solver libmamba
         # Ensure starting packages are from conda-forge.
         conda list --show-channel-urls
-        conda update -y -q conda
+        conda install -y -q "conda<=25.07"
         echo "$CONDA/bin" >> $GITHUB_PATH
 
-        conda install -y -q python=3.11 cmake=3.30.4 make=4.2 swig=4.0 "numpy<2" scipy=1.14 pytest=7.4 gflags=2.2
+        conda install -y -q python=3.11 cmake=3.30.4 make=4.2 swig=4.0 "numpy>=2.0,<3.0" scipy=1.16 pytest=7.4 gflags=2.2
 
         # install base packages for ARM64
         if [ "${{ runner.arch }}" = "ARM64" ]; then
--- a/conda/faiss-gpu-cuvs/meta.yaml
+++ b/conda/faiss-gpu-cuvs/meta.yaml
@@ -101,20 +101,20 @@ outputs:
         - mkl =2023.0  # [x86_64]
         - _openmp_mutex =4.5=2_kmp_llvm  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - {{ pin_subpackage('libfaiss', exact=True) }}
       run:
         - mkl =2023.0  # [x86_64]
         - _openmp_mutex =4.5=2_kmp_llvm  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - packaging
         - {{ pin_subpackage('libfaiss', exact=True) }}
     test:
       requires:
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - scipy
-        - pytorch <2.5
+        - pytorch >=2.7
         - pytorch-cuda {{ cuda_constraints }}
       commands:
         - python -X faulthandler -m unittest discover -v -s tests/ -p "test_*"
--- a/conda/faiss-gpu/meta.yaml
+++ b/conda/faiss-gpu/meta.yaml
@@ -96,20 +96,20 @@ outputs:
       host:
         - mkl =2023.0  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - _openmp_mutex =4.5=2_kmp_llvm  # [x86_64 and not win]
         - {{ pin_subpackage('libfaiss', exact=True) }}
       run:
         - mkl =2023.0  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - packaging
         - {{ pin_subpackage('libfaiss', exact=True) }}
     test:
       requires:
-        - numpy >=1.19,<2
+        - numpy >=2.0,<3.0
         - scipy
-        - pytorch <2.5
+        - pytorch >=2.7
         - pytorch-cuda {{ cuda_constraints }}
       commands:
         - python -X faulthandler -m unittest discover -v -s tests/ -p "test_*"
--- a/.github/actions/build_conda/action.yml
+++ b/.github/actions/build_conda/action.yml
@@ -43,7 +43,7 @@ runs:
       run: |
         # Ensure starting packages are from conda-forge.
         conda list --show-channel-urls
-        conda install -y -q "conda!=24.11.0"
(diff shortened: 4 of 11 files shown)


conda/faiss-gpu-cuvs/meta.yaml
```diff
@@ -101,20 +101,20 @@ outputs:
         - mkl =2023.0  # [x86_64]
         - _openmp_mutex =4.5=2_kmp_llvm  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy =2.3.3
```
jakirkham: This can be relaxed to just ensure NumPy 2 is picked up

```suggestion
        - numpy >=2,<3.0a0
```

Though if you need a particular NumPy version, the lower bound could be updated

```suggestion
        - numpy >=2.3,<3.0a0
```

conda/faiss-gpu-cuvs/meta.yaml
```diff
@@ -101,20 +101,20 @@ outputs:
...
         - {{ pin_subpackage('libfaiss', exact=True) }}
       run:
         - mkl =2023.0  # [x86_64]
         - _openmp_mutex =4.5=2_kmp_llvm  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy =2.3.3
```
jakirkham: The `numpy` package already specifies `run_exports` that constrains the `numpy` version used at runtime (both [in conda-forge]( https://github.com/conda-forge/numpy-feedstock/blob/b8dfd7c7cc9d7a7f5dafa2517654cbc87d5f8f72/recipe/recipe.yaml#L61-L62 ) and [in `defaults`]( https://github.com/AnacondaRecipes/numpy-feedstock/blob/d58e1380d7fe623dcb6813548ce47be30109db32/recipe/meta.yaml#L69-L70 ))

So this can just be either...

```suggestion
        - numpy
```

...or dropped entirely...


```suggestion
```

conda/faiss-gpu-cuvs/meta.yaml
```diff
@@ -101,20 +101,20 @@ outputs:
...
         - packaging
         - {{ pin_subpackage('libfaiss', exact=True) }}
     test:
       requires:
-        - numpy >=1.19,<2
+        - numpy =2.3.3
```
jakirkham: If you want to constrain to NumPy 2 in tests, this could be done like so...

```suggestion
        - numpy >=2,<3.0a0
```

If you need the lower bound, then you could add it like so

```suggestion
        - numpy >=2.3,<3.0a0
```

conda/faiss-gpu/meta.yaml
```diff
@@ -96,20 +96,20 @@ outputs:
       host:
         - mkl =2023.0  # [x86_64]
         - python {{ python }}
-        - numpy >=1.19,<2
+        - numpy =2.3.3
```
jakirkham: Would recommend the same changes in this and other files

jakirkham (commented):
Thanks Michael! 🙏

Had a few suggestions on how to build with NumPy 2, which will make the packages more flexible for users installing them

Also [these conda-forge NumPy docs]( https://conda-forge.org/docs/maintainer/knowledge_base/#building-against-numpy ) may provide additional context

jakirkham:
The NumPy changes also look reasonable to my eyes. Though cc-ing @seberg in case I've missed anything

Also NumPy has [a migration guide for NumPy 2]( https://numpy.org/doc/stable/numpy_2_0_migration_guide.html ). Additionally `ruff` has builtin support for migrating Python code to be NumPy 2 compliant, which can help automatically update code

mnorris11:
> Thanks Michael! 🙏
> 
> Had a few suggestions on how to build with NumPy 2, which will make the packages more flexible for users installing them
> 
> Also [these conda-forge NumPy docs](https://conda-forge.org/docs/maintainer/knowledge_base/#building-against-numpy) may provide additional context

Thanks @jakirkham for the review!
The flexible versions you suggest sound the best to me. This is the most user friendly.
We have experienced frequent nightly failures in the past when conda decides to pick up a different version (i.e. 2.1 instead of 2.3 etc), so by default I have been pinning exact versions. (I will change it to go with the flexible pinning for now, but we may have to go back to an exact version later.)

seberg:
Looks fine, cannot say quickly why the item calls are needed, but probably OK. Tolerance changes may mean that some internal computation previously went to float64 and now stay at float32, I guess. But that is presumably also fine.

jakirkham:
> Thanks @jakirkham for the review!
> The flexible versions you suggest sound the best to me. This is the most user friendly.
> We have experienced frequent nightly failures in the past when conda decides to pick up a different version (i.e. 2.1 instead of 2.3 etc), so by default I have been pinning exact versions. (I will change it to go with the flexible pinning for now, but we may have to go back to an exact version later.)

Happy to help 🙂

Thanks Michael! 🙏

If you run into issues, please let me know. Would be interested in taking a look at those when they come up

.github/actions/build_cmake/action.yml
```diff
@@ -36,10 +36,10 @@ runs:
         conda config --set solver libmamba
         # Ensure starting packages are from conda-forge.
         conda list --show-channel-urls
-        conda update -y -q conda
+        conda install -y -q "conda<=25.07"
```
jakirkham, line 39: Guessing this is due to an issue with `conda` version `25.09`, do you have more details on the issue seen?

mnorris11: Yes, seems like it just surfaced. Prior runs used 25.07 so I just pinned there.
https://github.com/facebookresearch/faiss/actions/runs/18175470029/job/51741940955
The last logs are:
```
Error while loading conda entry point: conda-libmamba-solver (cannot import name 'Spinner' from 'conda.common.io' (/home/runner/miniconda3/lib/python3.12/site-packages/conda/common/io.py))

CondaValueError: You have chosen a non-default solver backend (libmamba) but it was not recognized. Choose one of: classic

Error: Process completed with exit code 1.
```

jakirkham: Thanks Michael! 🙏

Think I can see what is going on. Have filed an upstream issue: https://github.com/conda/conda/issues/15287

Also to help with debugging future Conda or conda-forge issues, have suggested a few commands we can add to CI in issue: https://github.com/facebookresearch/faiss/issues/4599

This would help in providing additional detail when debugging. Additionally this would help in upstream bug reports

PR #4528, opened 2025-08-11
Remove invalid assertion checking #neighbors == graph degree

0ctopus13prime:
# Description
As fixed in https://github.com/facebookresearch/faiss/pull/4516, graph degree can be adjusted when small number of vectors are provided. Therefore, the number of neighbors (precisely, the maximum number of neighbors) can be smaller than graph degree after it was adjusted.

This PR is to remove the assertion check that fails if #neighbors not equals to graph degree, which is not always true.


0ctopus13prime:
Build w\ ROCm seems taking more than 5hr, is it normal?

0ctopus13prime:
@mnorris11 Could you disable ROCm for this PR?
Thank you.

mnorris11:
We will skip that signal, yes.

0ctopus13prime:
Thank you @mnorris11 :)

0ctopus13prime:
Hi team, could you push this if it looks good on you?
Not sure why ROCm is pending, but this change does not require it, we can safely skip it

mnorris11:
Yes, we will do it today.

PR #4532, opened 2025-08-14
Adding Idx tags in extended APIs

jinsolp:
Adds idx tags in `_ex` API. Current support for `search_ex` and `add_with_ids_ex`. Right now, it is designed to throw an error if the type is not `int64`.


contrib/torch_utils.py
```diff
@@ -79,6 +79,15 @@ def swig_ptr_from_IndicesTensor(x):
     return faiss.cast_integer_to_idx_t_ptr(
         x.untyped_storage().data_ptr() + x.storage_offset() * 8)
 
+
+def swig_ptr_from_Int64Tensor(x):
```
mnorris11, line 83: sorry, why did swig_ptr_from_IndicesTensor not work?

jinsolp: I was trying to distinguish `swig_ptr_from_IndicesTensor` from `swig_ptr_from_Int64Tensor` for future use with different idx types!

limqiying:
Hey @jinsolp @junjieqi  I have to revert this change. There is an issue with class_wrappers, where faiss.Int64 is not valid. I think it needs to be defined first. Will revert this change as it is blocking some of our users from using the library.

Screenshot_2025-08-27_at_17 41 47

limqiying:
#4563

jinsolp:
I see, yes please revert it for now! I'm not sure why the issue happened though because `Int64` is defined as an enum in `faiss/Index.h`?

limqiying:
@jinsolp yeah -- we will work with the reporter to figure out the issue, so we can reinstate the change. I'm asking for a repro of the issue now.

PR #4539, opened 2025-08-19
bugfix: add a macro guard for avx512 operation

conanhujinming:
### What to fix
There is a small bug in the faiss/utils/rabitq_simd.h that some avx512 only supported operation is not guarded by a corresponding macro. As a result, the compile will fail if the arch does not support avx512.

### What I did
I add the macro guard.

### What I checked
I have compiled locally.

This should be an easy fix.


faiss/utils/rabitq_simd.h
```diff
@@ -93,7 +94,7 @@ inline __m512i get_lookup_512() {
             /* 1 */ 1,
             /* 0 */ 0);
 }
-
+#endif
 /**
  * Returns the lookup table for AVX2 popcount operations.
  * This table is used for lookup-based popcount implementation.
```
mnorris11, line 101: get_lookup_256 will also need an include guard, for AVX2.

faiss/utils/rabitq_simd.h
```diff
@@ -152,7 +154,7 @@ inline __m512i popcount_lookup_avx512(__m512i v_and) {
     const __m512i popcnt2 = _mm512_shuffle_epi8(lookup, hi);
     return _mm512_add_epi8(popcnt1, popcnt2);
 }
-
+#endif
 /**
  * Performs lookup-based popcount on AVX2 registers.
  *
```
mnorris11, line 163: popcount_lookup_avx2 will need an AVX2 guard.

mnorris11:
Thank you for this bug fix! We actually also need AVX2 guards. Would you be able to add those?

faiss/utils/rabitq_simd.h
mnorris11: (It looks like `#if defined(__AVX2__)`)

conanhujinming:
Done. Please review, thanks.

alexanderguzhva:
looks good overall

PR #4550, opened 2025-08-25
Simplify RaBitQ slightly, improving speed and recall

ddrcoder:
## Introduction
The [RaBitQ paper](https://arxiv.org/abs/2405.12497) describes a method for implementing vector inner products with bitwise operations, quantizing the query with a uniform scalar quantizer, then performing a **bitwise and** of a single bit position of the query vector with the document vector, then using a `popcount()` to accumulate the inner product. This is done for each query bit position in `[0, B_q)`, and an additional `popcount()` on the document vector alone. This additional `popcount()` can be avoided by rearranging the arithmetic.

The quantization of the query, `q_u`, is specified in equation 18 of the paper, reproduced here:
```lang=txt
q' = q @ P.T
v_l, v_h = min(q'), max(q')
delta = v_h - v_l
u_i[i] = sample(uniform(0,1))
q_u[i] = floor((q'[i] - v_l) / (v_h - v_l) + u[i])
```

The *unsigned dot product* is then estimated as (eq. 22):
```lang=txt
x_b[i] = (o @ P.T)[i] > 0 ? 1 : 0
q_b[i, j] = (q_u[i] >> j) & 1
unsigned_inner_product = sum_j(2**j * sum_i(x_b[i] * q[i, j])
```
Note that `x_b[i] * q[i, j]` is implemented as a bitwise AND.

This is then used to estimate the real inner product (eq. 20):
```lang=txt
inner_product_estimate
  = (2 * delta * unsigned_inner_product
     + 2 * v_l * sum_i(x_b[i])
     - delta * sum(q_u[i])
     - 1) / sqrt(D)
 ```

## Adjustment
The presence of `sum_i(x_b[i])` can be eliminated by redefining the quantization scheme used to construct `q_b[i,j]`. Instead of mapping the query to an unsigned integer, the original value is scaled, then rounded to a **signed, odd integer** in a manner very similar to a paper which builds upon RaBitQ (https://arxiv.org/abs/2409.09913).
```lang=txt
           |-------|-------|-------|-------|-------|-------|-------|-------|
 q'[i]*m: -8   v  -6   v  -4   v  -2   v   0   v  +2   v  +4   v  +6   v  +8
               |       |       |       |       |       |       |       |
 odd_int:     -7      -5      -3      -1      +1      +3      +5      +7
               |       |       |       |       |       |       |       |
               v       v       v       v       v       v       v       v
  q_s[i]:      0       1       2       3       4       5       6       7
```

This has the convenient property that `odd_int = sum(2**j * (2 * q_s[i, j] - 1))`. That is, each component is encoded as `(±4) + (±2) + (±1)`.

With both queries and documents now mapped to the signed, odd integer domain, the signed integer dot product can be computed more directly. All query bits represent `±2^j`, and all document bits represent `±1`, so the maximum possible dot product is `(2^0 + 2^1 + ... + 2^j) * D = (2^(B_q-1) - 1) * D`. Every *mismatched bit* represents a *mismatched sign*, so each such bit mismatch represents a replacement of an inner product term `+2^j` with `-2^j`, a reduction of `2 * 2^j`.

So, the signed dot product can be estimated as:
```lang=txt
signed_inner_product
  = (2 ** B_q - 1) * D
  - 2 * sum_j(2**j * sum_i(xor(q_s[i,j], x_b[i])))
```
Which can then be simply rescaled to to estimate the real inner product:
```
scalar = norm(q) / (norm(q_s) * sqrt(D))
inner_product_estimate = signed_inner_product * scalar
```

## Results
But because queries are expected to be centered around 0 due to random rotation, there is minimal loss in accuracy by defining the scalar quantizer such that the bias is fixed at zero. To the contrary, the centering of the scalar quantizer appears to **correct a systematic bias** in the RaBitQ quantizer which may lead to query components being disproportionately rounded up or down depending on the value of the `bias`. Particularly for smaller values of `B_q`, this imbalance can lead to significant systematic bias, adversely affecting recall. For `B_q = 1` in particular, experiments show that this centering **halves recall loss**.

Costs and recall losses are consistently reduced across all `B_q` values tested, for both a 1024-d and 128-d test corpus. Results for a 1024-d test:
* `B_q = 1`: 60.3% -> 86.16% recall (2.9x loss reduction)
* `B_q = 2`: 87.3% -> 87.5% recall (+0.2 pp)
* `B_q = 3`: 87.9% -> 88.1% recall (+0.2 pp)

## Interfaces / Compatibility
This new mode is gated by a parameter on the distance computer, exposed through `IVFRaBitQSearchParameters::centered` among other APIs. For fairer performance comparisons, the **existing implementation** is also sped up with a specialized SIMD implementations for the doc-side `popcount()`.

### DIFF
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -329,49 +329,17 @@ float RaBitDistanceComputerQ::distance_to_code(const uint8_t* code) {
              metric_type == MetricType::METRIC_INNER_PRODUCT));
 
     // split the code into parts
+    size_t size = (d + 7) / 8;
     const uint8_t* binary_data = code;
-    const FactorsData* fac =
-            reinterpret_cast<const FactorsData*>(code + (d + 7) / 8);
+    const FactorsData* fac = reinterpret_cast<const FactorsData*>(code + size);
 
-    // // this is the baseline code
-    // //
-    // // compute <q,o> using integers
-    // size_t dot_qo = 0;
-    // for (size_t i = 0; i < d; i++) {
-    //     // extract i-th bit
-    //     const uint8_t masker = (1 << (i % 8));
-    //     const uint8_t bit = ((binary_data[i / 8] & masker) == masker) ? 1 :
-    //     0;
-    //
-    //     // accumulate dp
-    //     dot_qo += bit * rotated_qq[i];
-    // }
-
-    // this is the scheme for popcount
-    const size_t di_8b = (d + 7) / 8;
-    const size_t di_64b = (di_8b / 8) * 8;
-
-    // Use the optimized popcount function from rabitq_simd.h
-    float dot_qo =
-            rabitq_dp_popcnt(rearranged_rotated_qq.data(), binary_data, d, qb);
+    // See RaBitDistanceComputerNotQ::distance_to_code() for baseline code.
+    float dot_qo = rabitq::bitwise_and_dot_product(
+            rearranged_rotated_qq.data(), binary_data, size, qb);
 
     // It was a willful decision (after the discussion) to not to pre-cache
-    //   the sum of all bits, just in order to reduce the overhead per vector.
-    uint64_t sum_q = 0;
-    {
-        // process 64-bit popcounts
-        for (size_t i = 0; i < di_64b; i += 8) {
-            const auto yv = *(const uint64_t*)(binary_data + i);
-            sum_q += __builtin_popcountll(yv);
-        }
-
-        // process leftovers
-        for (size_t i = di_64b; i < di_8b; i++) {
-            const auto yv = *(binary_data + i);
-            sum_q += __builtin_popcount(yv);
-        }
-    }
-
+    // the sum of all bits, just in order to reduce the overhead per vector.
+    auto sum_q = rabitq::popcount(binary_data, size);
     float final_dot = 0;
     // dot-product itself
     final_dot += query_fac.c1 * dot_qo;
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -32,6 +32,8 @@ struct QueryFactorsData {
 
     float qr_to_c_L2sqr = 0;
     float qr_norm_L2sqr = 0;
+
+    float int_dot_scale = 1;
 };
 
 static size_t get_code_size(const size_t d) {
@@ -310,6 +312,7 @@ struct RaBitDistanceComputerQ : RaBitDistanceComputer {
 
     // the number of bits for SQ quantization of the query (qb > 0)
     uint8_t qb = 8;
+    bool centered = false;
     // the smallest value divisible by 8 that is not smaller than dim
     size_t popcount_aligned_dim = 0;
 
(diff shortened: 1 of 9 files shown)


alexanderguzhva:
@ddrcoder let me review this

alexanderguzhva:
@ddrcoder Do you have any numbers with improvements, just in case?

alexanderguzhva:
yep, I see the same. It is pretty effective for 1-bit and 2-bit query quantization

alexanderguzhva:
https://github.com/facebookresearch/faiss/commit/9945c11e53ce3c4273f47d7045f346f85c9d513c shows noticably worse recall rate vs the initial version in certain situations

ddrcoder:
I've update this to resolve a few bugs I'd introduced:
* `INNER_PRODUCT` is fixed, I'd swapped out the wrong value.
* The `float->int` mapping no longer branches based on `centered`.
* The bound has been adjusted to `(abs(v_min)+abs(v_max))/2`
* The quantized integer is now clamped to a valid range.

More significantly, I've switched the range selection for lower `qb` values to be based on the vector norm, not any particular component value.  Using the extreme values for the endpoints of a scalar quantizer when it has few buckets ends up allocating too few buckets for the values near the mean. Instead a static radius is selected to cover the normal distribution, with z-scores informed by offline analysis. For `qb=2`, `v_radius = 2 / sqrt(d)` (`|z|<2`) works well, and each additional bit widens this by `0.35`, up to `3.05` for `qb=5`.

As for results, my testing shows recall is generally improved for `qb` in `[1,4]`, beyond which there are minimal returns. Results vary greatly depending on test database size, `k`, overfetching factor, dimensionality, and the presence of random rotation.

Surprisingly, the improvement is *larger* for *unrotated* vectors. Perhaps these end up making more use of the outlier buckets of the quantizer?

For one internal Meta vector workload, with `d=1024`, measuring 10-recall@20, running with `AVX512-VPOPCNTDQ`, results are shown below, with `c={0,1}` indicating whether `centering` was enabled:

`metric=L2`, `AVX512-VPOPCNTDQ`:
```
qps=14942         recall= 32.911%           c=0/qb=1
qps=15728 [1.05x] recall= 84.323% [+51.41%] c=1/qb=1

qps=14403         recall= 85.383%           c=0/qb=2
qps=14568 [1.01x] recall= 92.419% [ +7.04%] c=1/qb=2

qps=13217         recall= 94.073%           c=0/qb=3
qps=14705 [1.11x] recall= 94.870% [ +0.80%] c=1/qb=3

qps=12726         recall= 95.304%           c=0/qb=4
qps=12602 [0.99x] recall= 95.466% [ +0.16%] c=1/qb=4

qps=11019         recall= 95.587%           c=0/qb=5
qps=11617 [1.05x] recall= 95.589% [ +0.00%] c=1/qb=5

```
`metric=L2`, `AVX2`:
```
qps=13286         recall= 32.869%           c=0/qb=1
qps=14756 [1.11x] recall= 84.324% [+51.45%] c=1/qb=1

qps=13384         recall= 85.307%           c=0/qb=2
qps=14377 [1.07x] recall= 92.441% [ +7.13%] c=1/qb=2

qps=10954         recall= 94.126%           c=0/qb=3
qps=12026 [1.10x] recall= 94.880% [ +0.75%] c=1/qb=3

qps=10785         recall= 95.304%           c=0/qb=4
qps=10542 [0.98x] recall= 95.512% [ +0.21%] c=1/qb=4

qps= 8735         recall= 95.591%           c=0/qb=5
qps= 9805 [1.12x] recall= 95.615% [ +0.02%] c=1/qb=5
```

`metric=IP`, `AVX2`:
```
qps=13036         recall= 22.734%           c=0/qb=1
qps=15319 [1.18x] recall= 83.144% [+60.41%] c=1/qb=1

qps=11802         recall= 81.354%           c=0/qb=2
qps=12929 [1.10x] recall= 91.638% [+10.28%] c=1/qb=2

qps=10861         recall= 93.434%           c=0/qb=3
qps=11474 [1.06x] recall= 94.404% [ +0.97%] c=1/qb=3

qps= 9241         recall= 94.744%           c=0/qb=4
qps=10815 [1.17x] recall= 94.983% [ +0.24%] c=1/qb=4

qps= 8651         recall= 95.133%           c=0/qb=5
qps= 8951 [1.03x] recall= 95.146% [ +0.01%] c=1/qb=5
```

alexanderguzhva:
@ddrcoder thanks, let me test it as well

alexanderguzhva:
please don't delete this PR, it is still interesting :)

faiss/impl/RaBitQuantizer.cpp
```diff
@@ -402,11 +383,30 @@ float RaBitDistanceComputerQ::distance_to_code(const uint8_t* code) {
     }
 }
 
+namespace {
+
+// Ideal quantizer radii for quantizers of 1..8 bits, optimized to minimize L2
+// reconstruction error.
+const float z_max_by_qb[8] = {
```
alexanderguzhva, line 390: any comments on how this got produced?

ddrcoder: In a Python notebook, I generated 100,000 1024-d unit normal vectors, then used `scipy.optimize.shgo` to find the values which minimized the L2 reconstruction error for each value of `qb`.

I'd done this earlier in the year for another project, and yesterday I learned that Elastic Search arrived at this approach, as they describe nicely in a blog post: https://www.elastic.co/search-labs/blog/scalar-quantization-optimization

PR #4553, opened 2025-08-26
dynamic dispatch distances_simd

subhadeepkaran:
Summary:
`fvec_madd` is the first function to test dispatching to AVX and AVX512

distances_simd.cpp is split into specialized files distances_avx2.cpp distances_avx512.cpp that are compiled with appropriate flags.

Reviewed By: mnorris11

Differential Revision: D72937708


faiss/utils/simd_impl/distances_aarch64.cpp
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <faiss/utils/distances.h>
+
+#define AUTOVEC_LEVEL SIMDLevel::ARM_NEON
```
alexanderguzhva, line 10: the name of the file should be changed to `distances_arm_neon.cpp`

PR #4569, opened 2025-09-03
Optimize exhaustive_L2sqr_blas_cmax with AVX-512 intrinsics.

mulugetam:
Add a template specialization of `exhaustive_L2sqr_blas_cmax` for `SIMDLevel::AVX512` in `distances_avx512.cpp`, alongside the existing `AVX2` and `ARM_SVE` specializations. The new kernel doubles the SIMD width of the post-`sgemm` fixup pass from 8 to 16 lanes, processing 32 elements per inner-loop iteration instead of 16.

**Benchmark**

`benchs/bench_index_flat.py` delivers up to a **1.16x** speedup compared to the AVX2 implementation.

```
+---------+----------+------------+---------+
|    d    | AVX2 k=1 | AVX512 k=1 | Speedup |
+---------+----------+------------+---------+
|   128   | 0.181814 |   0.156076 |  1.16x  |
|   256   | 0.309386 |   0.288921 |  1.07x  |
|   768   | 0.836036 |   0.814894 |  1.03x  |
+---------+----------+------------+---------+
```

**K-means** on `SIFT1M`, using `benchs/bench_all_ivf/bench_kmeans.py`, achieves up to a **1.06x** speedup relative to the existing `AVX2` implementation.

### DIFF
--- a/faiss/utils/simd_impl/distances_avx512.cpp
+++ b/faiss/utils/simd_impl/distances_avx512.cpp
@@ -9,6 +9,35 @@
 
 #include <immintrin.h>
 
+#include <memory>
+
+#include <faiss/impl/AuxIndexStructures.h>
+#include <faiss/impl/ResultHandler.h>
+#include <faiss/utils/distances_fused/distances_fused.h>
+#include <faiss/utils/simd_impl/exhaustive_L2sqr_blas_cmax.h>
+
+#ifndef FINTEGER
+#define FINTEGER long
+#endif
+
+extern "C" {
+
+int sgemm_(
+        const char* transa,
+        const char* transb,
+        FINTEGER* m,
+        FINTEGER* n,
+        FINTEGER* k,
+        const float* alpha,
+        const float* a,
+        FINTEGER* lda,
+        const float* b,
+        FINTEGER* ldb,
+        float* beta,
+        float* c,
+        FINTEGER* ldc);
+}
+
 #define THE_SIMD_LEVEL SIMDLevel::AVX512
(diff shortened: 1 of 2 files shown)


mdouze:
Thanks for the contribution. We are in the process of overhauling the SIMD support for Faiss, so we may prefer to defer integrating this PR. CC: @subhadeepkaran

mulugetam:
Refactored to use SIMD DD. Could you please review? @mdouze @subhadeepkaran @mnorris11

mulugetam:
@mnorris11 Could you please review this as well? Thanks!

faiss/utils/simd_impl/distances_avx512.cpp
```diff
@@ -1104,4 +1133,208 @@ int fvec_madd_and_argmin<SIMDLevel::AVX512>(
...
+                       &one,
+                       y + j0 * d,
+                       &di,
+                       x + i0 * d,
+                       &di,
+                       &zero,
+                       ip_block.get(),
+                       &nyi);
+            }
+            for (int64_t i = i0; i < (int64_t)i1; i++) {
```
mnorris11: 1. Should this inner loop be parallelized like it is in AVX2 and SVE versions? 
2. (nit) There are a few places where the AVX2 version uses static_cast and this version uses (int64_t).

mnorris11: On local benchmark, I see avx512 regress without pragma, probably good to add. Otherwise LGTM!

mulugetam:
@mnorris11 rebased with changes.

Issue #4313, opened 2025-04-25
AVX512_SPR may be non-loadable for python


alexanderguzhva:
1. Grab the master branch
2. Compile faiss
``` bash
cmake -B build_rel -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=ON -DBUILD_TESTING=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DFAISS_OPT_LEVEL=avx512_spr .
make -j 8 -C build_rel swigfaiss_avx512_spr
cd build_rel/faiss/python
python3 setup.py install
```
3. load faiss in python
``` bash
$ python3
Python 3.11.12 (main, Apr 25 2025, 15:07:13) [GCC 9.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import faiss
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/zilliz/faiss111/faiss/build_rel/faiss/python/faiss/__init__.py", line 17, in <module>
    from .loader import *
  File "/home/ubuntu/zilliz/faiss111/faiss/build_rel/faiss/python/faiss/loader.py", line 149, in <module>
    from .swigfaiss import *
ModuleNotFoundError: No module named 'faiss.swigfaiss'
>>> exit()
```
4. this works if the environment variable is added
``` bash
$ FAISS_OPT_LEVEL=AVX512_SPR python3
Python 3.11.12 (main, Apr 25 2025, 15:07:13) [GCC 9.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import faiss
>>> exit()

```

mnorris11:
cc @mulugetam

mnorris11:
Took a look. This is due to AVX512_SPR not being available in instruction sets: https://github.com/facebookresearch/faiss/blob/e5902b00b5b3710918c67c6907db39b2fdd97baa/faiss/python/loader.py#L91

(The env var is None for both AVX512 and AVX512_SPR)

We're using numpy 1.26, so we get here: https://github.com/facebookresearch/faiss/blob/e5902b00b5b3710918c67c6907db39b2fdd97baa/faiss/python/loader.py#L52 and avx512_spr is not in __cpu_features__. The results on my system are:
```
[...]

PR #4577, opened 2025-09-10
Added code to catch 'ModuleNotFoundError' exception

abhi10691:
I added a fix for the following opened bug:

https://github.com/facebookresearch/faiss/issues/4313


junjieqi:
@abhi10691 I'm wondering how I could reproduce the error and I'm not sure if the right way to capture error which are not related with it.

abhi10691:
@junjieqi I reproduced the error on my system, by following the steps mentioned by @alexanderguzhva on the bug (https://github.com/facebookresearch/faiss/issues/4313).

abhi10691:
> @abhi10691 I'm wondering how I could reproduce the error and I'm not sure if the right way to capture error which are not related with it.

@junjieqi I captured all the scenarios, in the error message, as suggested in the following comment by @alexanderguzhva (https://github.com/facebookresearch/faiss/issues/4313#issuecomment-2852974981).

abhi10691:
@alexanderguzhva @junjieqi Can you review this? Also tried to add @mnorris11, but wasn't able to.

alexanderguzhva:
@abhi10691 looks good to me, that's quite a detailed comment about the problem!

faiss/python/loader.py
```diff
@@ -150,7 +151,28 @@ def is_sve_supported():
...
+                 + "- If 'SVE' (case sensitive) is supported on your system, you can set the FAISS_OPT_LEVEL=SVE to build the SWIG wrapper with 'SVE' support."
+                 + " You will have to build the 'swigfaiss_sve' target in this case.\n"
+                 + "- If none of the above instruction sets are supported on your system, you can execute 'cmake' without setting the FAISS_OPT_LEVEL variable"
+                 + " and build the 'swigfaiss' target.\n\n"
+                 + "After doing the above, setting the FAISS_OPT_LEVEL environment variable to the chosen instruction set, so that the python interpreter picks it, would help,"
+                 + " as it would prevent the code from trying to import the wrong SWIG wrapper. This is not necessary, as there are exception catch blocks in place, incase the"
+                 + " line to import the wrong SWIG wrapper is executed. Even if the FAISS_OPT_LEVEL environment variable is not set, the correct SWIG wrapper will be imported, possibly"
+                 + " after unsuccessful imports of a few incorrect SWIG wrappers.")
+
+        sys.exit(1)
```
mnorris11: Hi @abhi10691 , thanks for this change! Can we reformat this?

I removed the last section after the bullet points. You mentioned it is not necessary to set the FAISS_OPT_LEVEL after choosing a bulleted step. If so, we can just omit it.
```
        formatted_ins_sets = ", ".join(supported_instruction_sets())
        error_msg = f"""
            No module named 'faiss.swigfaiss' found. To fix this, you must \
            both:
            A) Set the correct FAISS_OPT_LEVEL value when executing 'cmake'.
            B) Build the correct SWIG wrapper.

            These are the supported instruction sets on your system:
            {formatted_ins_sets}
            - If 'AVX512_SPR' (case insensitive) is supported on your system, \
            you can set the FAISS_OPT_LEVEL=avx512_spr \
            to build the SWIG wrapper with 'AVX512-SPR' support.
            You will have to build the 'swigfaiss_avx512_spr' \
            target in this case.
            - If 'AVX512' (case insensitive) is supported on your system, \
            you can set the FAISS_OPT_LEVEL=avx512 to build the SWIG wrapper \
            with 'AVX512' support.
            You will have to build the 'swigfaiss_avx512' target in this case.
            - If 'AVX2' (case sensitive) is supported on your system, you can \
            set the FAISS_OPT_LEVEL=AVX2 to build the SWIG wrapper \
            with 'AVX2' support.
            You will have to build the 'swigfaiss_avx2' target in this case.
            - If 'SVE' (case sensitive) is supported on your system, you can \
            set the FAISS_OPT_LEVEL=SVE to build the SWIG wrapper with \
            'SVE' support.
            You will have to build the 'swigfaiss_sve' target in this case.
            - If none of the above instruction sets are supported on your \
            system, you can execute 'cmake' without setting the \
            FAISS_OPT_LEVEL variable and build the 'swigfaiss' target.
        """
        logger.error(error_msg)
```

abhi10691: Hi @mnorris11, I did the reformatting. To preserve indentation, I had to use f" on each line, instead of using f"""!

Also I am attaching a snapshot of the error message, when 'import faiss' is executed.
snapshot

Does this seem fine?

mnorris11: Hi @abhi10691 , the issue is that the line lengths go over 80 characters and it is getting flagged by the linter. That is why I added all the \ in the prior comment.

faiss/python/loader.py
```diff
@@ -150,7 +151,44 @@ def is_sve_supported():
...
-    # we import * so that the symbol X can be accessed as faiss.X
-    logger.info("Loading faiss.")
-    from .swigfaiss import *
-    logger.info("Successfully loaded faiss.")
+    try:
+        # we import * so that the symbol X can be accessed as faiss.X
+        logger.info("Loading faiss.")
+        from .swigfaiss import *
+        logger.info("Successfully loaded faiss.")
+    except ModuleNotFoundError as e:
```
mnorris11: This line is one more getting flagged by the linter. We can leave out the `as e`.
```
except ModuleNotFoundError:
```

faiss/python/loader.py
abhi10691: Hi @mnorris11, I ensured that the line lengths do not go over 80 characters.

faiss/python/loader.py
abhi10691: Made this change also.

mnorris11:
Thanks for the fix @abhi10691 !

PR #4581, opened 2025-09-16
fix: initializing order of gpu compile options

simshi:
# issue
I met gpu_compile_options initialization order issue over gcc 8.4.0. Maybe related with LTO bug?

# solution
make it deterministic, no "crossing fingers" required. :smile_cat:


mnorris11:
Hi @simshi , thanks for the contribution; what was the error and how did you repro?

simshi:
> Hi @simshi , thanks for the contribution; what was the error and how did you repro?

The result is get_compile_options() returns without "GPU".
Should be caused by: InitGpuCompileOptions_instance is created/initialized before gpu_compile_options, cause gpu_compile_options reset to empty.

No guarantee to reproduce, my dev environment is ubuntu18 with gcc 8.4.0

PR #4587, opened 2025-09-23
Fix AIX compilation issue while building python extension

### LINKED ISSUE #4586: Fix AIX compilation issue while building  python extension
AIX currently has errors while building python extension of faiss.
The issues are 
1) Swigfaiss can be a SHARED Library , but is currently  a module.
2) Suffix in AIX should be .a instead of .so .
3) Minor syntax changes due to CMAKE>= 4.0 .
4) avx files are not generated in AIX , so don't want to check for assertion of avx files.

Will be raising a pull request  to fix these errors.

KV2773:
Closes #4586.

### DIFF
--- a/faiss/python/CMakeLists.txt
+++ b/faiss/python/CMakeLists.txt
@@ -99,19 +99,11 @@ else()
   find_package(faiss REQUIRED)
 endif()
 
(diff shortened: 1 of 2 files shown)


KV2773:
yes it works, should I squash the PR with this change.

PR #4594, opened 2025-09-30
Solve error when build faiss

nicabreon:
Solve error when build faiss
fatal error: faiss/perf_tests/utils.h: No such file or directory

Screenshot 2025-09-30 at 10 07 03

Screenshot 2025-09-30 at 10 07 23


mnorris11:
This builds in the CI. It indicates your conda environment may have a setup issue.

peter-joo:
here, the current test target uses its own header, so why not simply do
```
#include "utils.h"
```

This has been solved and merged already into main via:
https://github.com/facebookresearch/faiss/pull/4792/changes

PR #4595, opened 2025-09-30
RaBitQ Fast Scan

alibeklfc:
Summary:
**Introduction**

This diff adds a new index called the IndexRaBitQFastScan algorithm. The algorithm is based on the existing IndexRaBitQ but achieves higher speed as it processes batches of 32 data vectors concurrently. It leverages the established IndexFastScan architecture to enable efficient batch processing and parallelism.

**Implementation**

* **New Source and Header Files**: Added implementations for IndexRaBitQFastScan, following a similar interface to IndexRaBitQ.

* **Batched Processing**: The search operation processes multiple (32) data vectors in a single batch, taking advantage of low-level parallelism to improve throughput.

* **Specialized Post-processing Handler**:  A dedicated handler was added for IndexRaBitQFastScan to perform necessary post-processing during search because the LUT accumulates only partial distances. Unlike AQ Fast Scan's simple scalar post-processing, RaBitQ requires complex distance adjustments depending on both query and database vector factors.

* **LUT**: IndexRaBitQFastScan produces slightly different results than IndexRaBitQ due to an extra quantization step in the IndexFastScan architecture. Specifically:
  * The LUT computes a float value as c1 * inner_product + c2 * popcount, which is then quantized. This quantization can cause the results to differ slightly from those of IndexRaBitQ.
  * It is possible to avoid this by storing only the inner_product in the LUT, but doing so would require calculating all data vector popcounts during search, introducing a tradeoff between speed and accuracy.
  * With the idea proposed in diff D80904214, the algorithm can be modified in the future to eliminate the popcount calculation step, potentially improving both efficiency and accuracy.
* **Query Offset Parameter**: RaBitQ uses query factors in distance calculations that should be computed in `compute_float_LUT` method (the most efficient place since we are calculating `rotated_qq` anyways) and used for final distance calculations in handlers. However, the previous version of `compute_quantized_LUT` that calls `compute_float_LUT` did not know the query_offset, preventing proper storage of query factors at their global indices. To solve this, I added the extra parameter `query_offset` to both `compute_quantized_LUT` and `compute_float_LUT` methods. After this change, computed query factors can be accessed by the correct global query index during distance calculations, avoiding expensive recalculation.

**Testing**

* Conducted comprehensive tests in the test_rabitq suite covering accuracy comparisons with IndexRaBitQ for L2 and Inner Product metrics, encoding/decoding consistency, query quantization bit settings, small dataset functionality, performance against PQFastScan, serialization, memory management, error handling, and thread safety.
* All tests passed successfully, validating the correctness and robustness of IndexRaBitQFastScan.

**Results**
results_rabitq
* **Performance Dependency**: Performance measurements confirm that IndexRaBitQFastScan is notably faster than IndexRaBitQ when the qb value is high. While the original IndexRaBitQ experiences increased runtime with higher qb values, the fast scan variant maintains consistent runtime regardless of qb.
* **Parallelized Training Loop**: The training loop is parallelized, greatly reducing training time. This parallelism should also be added to the original IndexRaBitQ.
* **Consistency Across Metrics**: The performance advantages of IndexRaBitQFastScan hold true for both L2 and Inner Product metrics, demonstrating robustness across different distance measures.
* One of the next steps is to benchmark IndexRaBitQFastScan against other algorithms to evaluate its performance in a broader context.

Differential Revision: D81787307

### DIFF
--- a/faiss/IndexAdditiveQuantizerFastScan.cpp
+++ b/faiss/IndexAdditiveQuantizerFastScan.cpp
@@ -11,6 +11,7 @@
 #include <memory>
 
 #include <faiss/impl/FaissAssert.h>
+#include <faiss/impl/FastScanDistancePostProcessing.h>
 #include <faiss/impl/LocalSearchQuantizer.h>
 #include <faiss/impl/LookupTableScaler.h>
 #include <faiss/impl/ResidualQuantizer.h>
@@ -123,7 +124,8 @@ void IndexAdditiveQuantizerFastScan::estimate_norm_scale(
     }
 
     std::vector<float> dis_tables(n * M * ksub);
-    compute_float_LUT(dis_tables.data(), n, x);
+    FastScanDistancePostProcessing empty_context;
+    compute_float_LUT(dis_tables.data(), n, x, empty_context);
 
     // here we compute the mean of scales for each query
     // TODO: try max of scales
@@ -153,7 +155,8 @@ void IndexAdditiveQuantizerFastScan::compute_codes(
 void IndexAdditiveQuantizerFastScan::compute_float_LUT(
         float* lut,
         idx_t n,
-        const float* x) const {
+        const float* x,
+        const FastScanDistancePostProcessing&) const {
     if (metric_type == METRIC_INNER_PRODUCT) {
         aq->compute_LUT(n, x, lut, 1.0f);
     } else {
@@ -200,10 +203,12 @@ void IndexAdditiveQuantizerFastScan::search(
     }
 
     NormTableScaler scaler(norm_scale);
+    FastScanDistancePostProcessing context;
+    context.norm_scaler = &scaler;
     if (metric_type == METRIC_L2) {
-        search_dispatch_implem<true>(n, x, k, distances, labels, &scaler);
+        search_dispatch_implem<true>(n, x, k, distances, labels, context);
     } else {
-        search_dispatch_implem<false>(n, x, k, distances, labels, &scaler);
+        search_dispatch_implem<false>(n, x, k, distances, labels, context);
     }
 }
 
--- a/faiss/IndexAdditiveQuantizerFastScan.h
+++ b/faiss/IndexAdditiveQuantizerFastScan.h
@@ -62,7 +62,11 @@ struct IndexAdditiveQuantizerFastScan : IndexFastScan {
 
     void compute_codes(uint8_t* codes, idx_t n, const float* x) const override;
 
-    void compute_float_LUT(float* lut, idx_t n, const float* x) const override;
+    void compute_float_LUT(
+            float* lut,
+            idx_t n,
+            const float* x,
+            const FastScanDistancePostProcessing& context) const override;
 
     void search(
             idx_t n,
--- a/faiss/IndexFastScan.cpp
+++ b/faiss/IndexFastScan.cpp
@@ -7,17 +7,20 @@
 
 #include <faiss/IndexFastScan.h>
 
-#include <cassert>
-#include <climits>
+#include <omp.h>
+#include <cstring>
 #include <memory>
 
-#include <omp.h>
-
+#include <faiss/impl/CodePacker.h>
 #include <faiss/impl/FaissAssert.h>
+#include <faiss/impl/FastScanDistancePostProcessing.h>
 #include <faiss/impl/IDSelector.h>
 #include <faiss/impl/LookupTableScaler.h>
-#include <faiss/impl/ResultHandler.h>
+#include <faiss/impl/RaBitQUtils.h>
+#include <faiss/impl/pq4_fast_scan.h>
+#include <faiss/impl/simd_result_handlers.h>
 #include <faiss/utils/hamming.h>
+#include <faiss/utils/utils.h>
 
 #include <faiss/impl/pq4_fast_scan.h>
 #include <faiss/impl/simd_result_handlers.h>
@@ -163,14 +166,14 @@ void estimators_from_tables_generic(
         size_t k,
         typename C::T* heap_dis,
         int64_t* heap_ids,
-        const NormTableScaler* scaler) {
+        const FastScanDistancePostProcessing& context) {
     using accu_t = typename C::T;
 
     for (size_t j = 0; j < ncodes; ++j) {
         BitstringReader bsr(codes + j * index.code_size, index.code_size);
         accu_t dis = 0;
         const dis_t* dt = dis_table;
-        int nscale = scaler ? scaler->nscale : 0;
+        int nscale = context.norm_scaler ? context.norm_scaler->nscale : 0;
 
         for (size_t m = 0; m < index.M - nscale; m++) {
             uint64_t c = bsr.read(index.nbits);
@@ -178,10 +181,10 @@ void estimators_from_tables_generic(
             dt += index.ksub;
         }
 
-        if (nscale) {
+        if (nscale && context.norm_scaler) {
             for (size_t m = 0; m < nscale; m++) {
                 uint64_t c = bsr.read(index.nbits);
-                dis += scaler->scale_one(dt[c]);
+                dis += context.norm_scaler->scale_one(dt[c]);
                 dt += index.ksub;
             }
         }
@@ -193,29 +196,46 @@ void estimators_from_tables_generic(
     }
 }
 
-template <class C>
-ResultHandlerCompare<C, false>* make_knn_handler(
+} // anonymous namespace
+
+// Default implementation of make_knn_handler with centralized fallback logic
+void* IndexFastScan::make_knn_handler(
+        bool is_max,
         int impl,
         idx_t n,
         idx_t k,
         size_t ntotal,
         float* distances,
         idx_t* labels,
-        const IDSelector* sel = nullptr) {
-    using HeapHC = HeapHandler<C, false>;
-    using ReservoirHC = ReservoirHandler<C, false>;
-    using SingleResultHC = SingleResultHandler<C, false>;
+        const IDSelector* sel,
+        const FastScanDistancePostProcessing&) const {
+    // Create default handlers based on k and impl
+    if (is_max) {
+        using HeapHC = HeapHandler<CMax<uint16_t, int>, false>;
+        using ReservoirHC = ReservoirHandler<CMax<uint16_t, int>, false>;
+        using SingleResultHC = SingleResultHandler<CMax<uint16_t, int>, false>;
 
-    if (k == 1) {
-        return new SingleResultHC(n, ntotal, distances, labels, sel);
-    } else if (impl % 2 == 0) {
-        return new HeapHC(n, ntotal, k, distances, labels, sel);
-    } else /* if (impl % 2 == 1) */ {
-        return new ReservoirHC(n, ntotal, k, 2 * k, distances, labels, sel);
+        if (k == 1) {
+            return new SingleResultHC(n, ntotal, distances, labels, sel);
+        } else if (impl % 2 == 0) {
+            return new HeapHC(n, ntotal, k, distances, labels, sel);
+        } else {
+            return new ReservoirHC(n, ntotal, k, 2 * k, distances, labels, sel);
+        }
+    } else {
+        using HeapHC = HeapHandler<CMin<uint16_t, int>, false>;
+        using ReservoirHC = ReservoirHandler<CMin<uint16_t, int>, false>;
+        using SingleResultHC = SingleResultHandler<CMin<uint16_t, int>, false>;
+
+        if (k == 1) {
+            return new SingleResultHC(n, ntotal, distances, labels, sel);
+        } else if (impl % 2 == 0) {
+            return new HeapHC(n, ntotal, k, distances, labels, sel);
+        } else {
+            return new ReservoirHC(n, ntotal, k, 2 * k, distances, labels, sel);
+        }
     }
 }
-
-} // anonymous namespace
 
 using namespace quantize_lut;
 
@@ -223,10 +243,11 @@ void IndexFastScan::compute_quantized_LUT(
         idx_t n,
         const float* x,
         uint8_t* lut,
-        float* normalizers) const {
+        float* normalizers,
+        const FastScanDistancePostProcessing& context) const {
     size_t dim12 = ksub * M;
     std::unique_ptr<float[]> dis_tables(new float[n * dim12]);
-    compute_float_LUT(dis_tables.get(), n, x);
+    compute_float_LUT(dis_tables.get(), n, x, context);
 
     for (uint64_t i = 0; i < n; i++) {
         round_uint8_per_column(
@@ -263,10 +284,12 @@ void IndexFastScan::search(
             !params, "search params not supported for this index");
     FAISS_THROW_IF_NOT(k > 0);
 
+    FastScanDistancePostProcessing empty_context{};
     if (metric_type == METRIC_L2) {
-        search_dispatch_implem<true>(n, x, k, distances, labels, nullptr);
+        search_dispatch_implem<true>(n, x, k, distances, labels, empty_context);
     } else {
-        search_dispatch_implem<false>(n, x, k, distances, labels, nullptr);
+        search_dispatch_implem<false>(
+                n, x, k, distances, labels, empty_context);
     }
 }
 
@@ -277,7 +300,7 @@ void IndexFastScan::search_dispatch_implem(
         idx_t k,
         float* distances,
         idx_t* labels,
-        const NormTableScaler* scaler) const {
+        const FastScanDistancePostProcessing& context) const {
     using Cfloat = typename std::conditional<
             is_max,
             CMax<float, int64_t>,
@@ -308,15 +331,20 @@ void IndexFastScan::search_dispatch_implem(
         FAISS_THROW_MSG("not implemented");
     } else if (implem == 2 || implem == 3 || implem == 4) {
         FAISS_THROW_IF_NOT(orig_codes != nullptr);
-        search_implem_234<Cfloat>(n, x, k, distances, labels, scaler);
+        search_implem_234<Cfloat>(n, x, k, distances, labels, context);
     } else if (impl >= 12 && impl <= 15) {
         FAISS_THROW_IF_NOT(ntotal < INT_MAX);
         int nt = std::min(omp_get_max_threads(), int(n));
+        // Fall back to single-threaded implementations when parallelization not
+        // beneficial:
+        // - Single-core system (omp_get_max_threads() = 1)
+        // - Single query (n = 1)
+        // - OpenMP disabled (omp_get_max_threads() = 1)
         if (nt < 2) {
             if (impl == 12 || impl == 13) {
-                search_implem_12<C>(n, x, k, distances, labels, impl, scaler);
+                search_implem_12<C>(n, x, k, distances, labels, impl, context);
             } else {
-                search_implem_14<C>(n, x, k, distances, labels, impl, scaler);
+                search_implem_14<C>(n, x, k, distances, labels, impl, context);
             }
         } else {
             // explicitly slice over threads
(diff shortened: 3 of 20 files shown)


faiss/IndexAdditiveQuantizerFastScan.cpp
```diff
@@ -153,7 +153,8 @@ void IndexAdditiveQuantizerFastScan::compute_codes(
 void IndexAdditiveQuantizerFastScan::compute_float_LUT(
         float* lut,
         idx_t n,
-        const float* x) const {
+        const float* x,
+        idx_t) const {
     if (metric_type == METRIC_INNER_PRODUCT) {
```
alexanderguzhva, line 160: please check that the last parameter is zero, something like
```C++
void IndexAdditiveQuantizerFastScan::compute_float_LUT(..., idx_t query_offset) const {
    FAISS_THROW_IF_NOT_MSG(query_offset == 0, "query_offset is expected to be zero");
    ...
}
```
This should be the same for other child classes (such as `IndexPQFastScan`) that introduce this new variable and do not use it

faiss/IndexFastScan.cpp
```diff
@@ -193,40 +192,58 @@ void estimators_from_tables_generic(
     }
 }
 
-template <class C>
-ResultHandlerCompare<C, false>* make_knn_handler(
+} // anonymous namespace
+
+using namespace quantize_lut;
```
alexanderguzhva: please bring this specific line `using namespace quantize_lut;` down where it was, right before `void IndexFastScan::compute_quantized_LUT(...) {...}`, otherwise it is very confusing to understand why this line is needed at all.
Same for `IndexIVFFastScan`, please

faiss/IndexFastScan.cpp
```diff
@@ -193,40 +192,58 @@ void estimators_from_tables_generic(
...
+
+// Default implementation of make_knn_handler with centralized fallback logic
+void* IndexFastScan::make_knn_handler(
```
alexanderguzhva, line 202: why not something like 
```C++
template <class C, bool IsMax>
ResultHandlerCompare<C, IsMax>* make_knn_handler(...) {...}
```
?

faiss/IndexFastScan.cpp
```diff
@@ -314,24 +331,41 @@ void IndexFastScan::search_dispatch_implem(
         int nt = std::min(omp_get_max_threads(), int(n));
         if (nt < 2) {
             if (impl == 12 || impl == 13) {
-                search_implem_12<C>(n, x, k, distances, labels, impl, scaler);
+                search_implem_12<C>(
+                        n, x, k, distances, labels, impl, scaler, 0);
```
alexanderguzhva: are you sure that you should pass 0, not `query_offset`? if it is still `0`, then please add a comment why `0` here

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,527 @@
...
+#include <faiss/IndexRaBitQFastScan.h>
+#include <faiss/impl/RaBitQUtils.h>
+#include <faiss/impl/pq4_fast_scan.h>
+#include <faiss/utils/distances.h>
+#include <algorithm>
+#include <cmath>
+
+namespace faiss {
+
+inline size_t roundup(size_t a, size_t b) {
```
alexanderguzhva: `static inline`

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+    // Pre-allocate storage vectors for efficiency
+    factors_storage.clear();
+    query_factors_storage.clear();
+}
+
+IndexRaBitQFastScan::IndexRaBitQFastScan(const IndexRaBitQ& orig, int bbs)
+        : rabitq(orig.rabitq) {
+    // RaBitQ-specific validation
+    FAISS_THROW_IF_NOT_MSG(orig.d > 0, "Dimension must be positive");
```
alexanderguzhva, line 56: no `d == 0` case supported? :)

faiss/IndexRaBitQFastScan.h
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Extracted factors storage for batch processing
+    /// Size: ntotal, stores factors separately from packed codes
+    /// Mutable because compute_codes needs to resize this during encoding
+    mutable std::vector<FactorsData> factors_storage;
+
+    /// Query factors storage for reconstruction during search
+    /// Size: n_queries, stores query-specific factors for distance computation
+    /// Mutable because search needs to resize this during query processing
+    mutable std::vector<QueryFactorsData> query_factors_storage;
```
alexanderguzhva: these mutable variables definitely look Very TOTALLY **WRONG**, because it becomes impossible to perform a multi-threaded search, which is 99% use cases in prod. I understand that maybe Meta uses a single-threaded compilation flavor of the `Faiss` library, but not all other ppl. 

I understand that the problem comes from a limitation of `Index::compute_codes(uint8_t* codes, idx_t n, const float* x)` interface, because it does not allow multiple `codes`. Unfortunately, you have to stick with this interface. You don't have to rely on `IndexFastScan` as a last resort, not even to derive from `IndexFastScan` if the situation demands this, just use its components. Or you may introduce `IndexRaBitQFastScan::search_ex(...) {...}`. But please don't spoil the `::search()` function.

@mdouze your opinion is needed

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,527 @@
...
+        // Normalize distance from LUT lookup
+        const float normalized_distance = d32tab[i] * one_a + bias;
+
+        // Apply RaBitQ distance formula based on quantization mode
+        const auto& db_factors = rabitq_index->factors_storage[db_idx];
+        float adjusted_distance;
+
+        if (rabitq_index->centered) {
+            // For centered mode: normalized_distance contains the raw XOR
+            // contribution Apply the signed odd integer quantization formula:
```
alexanderguzhva: a dot is missing
`...XOR constribution. Apply...`

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,527 @@
...
+        used_centered = rabitq_params->centered;
+    } else if (params != nullptr) {
+        FAISS_THROW_IF_NOT_MSG(
+                false, "search params not supported for this index");
+    }
+
+    FAISS_THROW_IF_NOT(k > 0);
+
+    // Temporarily update qb and centered for this search
+    const_cast<IndexRaBitQFastScan*>(this)->qb = used_qb;
```
alexanderguzhva: please no const casts

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,527 @@
...
+template <class C, bool with_id_map>
+void RaBitQHeapHandler<C, with_id_map>::begin(const float* norms) {
+    normalizers = norms;
+    // Heap initialization is now done in constructor
+}
+
+template <class C, bool with_id_map>
+void RaBitQHeapHandler<C, with_id_map>::end() {
+// Reorder final results
+#pragma omp parallel for
```
alexanderguzhva: usually, `#pragma omp parallel for if (nq > XYZ)`, where `XYZ` is something like 1000. Please check how it is done in other `Faiss` code pieces

faiss/impl/RaBitQUtils.h
```diff
@@ -0,0 +1,186 @@
...
+        std::vector<uint8_t>& rotated_qq);
+
+/** Extract bit value from RaBitQ code in standard format.
+ * Used by IndexRaBitQ which stores bits sequentially.
+ *
+ * @param code          RaBitQ code data
+ * @param bit_index     which bit to extract (0 to d-1)
+ * @return              bit value (true/false)
+ */
+inline bool extract_bit_standard(const uint8_t* code, size_t bit_index) {
```
alexanderguzhva: `static inline`

faiss/impl/RaBitQUtils.h
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/** Extract bit value from FastScan code format.
+ * Used by IndexRaBitQFastScan which packs bits into 4-bit sub-quantizers.
+ *
+ * @param code          FastScan code data
+ * @param bit_index     which bit to extract (0 to d-1)
+ * @return              bit value (true/false)
+ */
+inline bool extract_bit_fastscan(const uint8_t* code, size_t bit_index) {
```
alexanderguzhva: `static inline`
down below as well

alexanderguzhva (changes requested):
Please talk to @mdouze as well

alexanderguzhva:
@alibeklfc technically speaking, you could go with dirty hacks using something like  `thread_local` variables (or maybe `std::mutex` + `std::unordered_map` + `std::thread::id()`), but only if you believe that you can handle it properly and without memory leaks.
Also, if you decide to go this way, then please put all these internals into a separate class, not inside `IndexRaBitQFastScan`:
```C++
class SomeBuffers {...};

class IndexRaBitQFastScan {
    ...
    // this is used for internal buffers during the search
    SomeBuffers some_buffers;

    // of course, this means that copy & move ops need to be customized as well
    IndexRaBitQFastScan(const IndexRaBitQFastScan&);
    IndexRaBitQFastScan(IndexRaBitQFastScan&&) nothrow;
    IndexRaBitQFastScan& operator=(const IndexRaBitQFastScan&);
    IndexRaBitQFastScan& operator=(IndexRaBitQFastScan&&) nothrow;
};
```

faiss/IndexRaBitQFastScan.h
mdouze: I gave the same feedback internally.

faiss/IndexRaBitQFastScan.cpp
alibeklfc: Added keyword static

faiss/impl/RaBitQUtils.h
alibeklfc: We cannot use static inline in .h file. The options are using inline in .h or moving the function to the implementation file

faiss/impl/RaBitQUtils.h
alibeklfc: We cannot use static inline in .h file. The options are using inline in .h or moving the function to the implementation file

faiss/IndexRaBitQFastScan.cpp
alibeklfc: Modified to #pragma omp parallel for if (nq > 100)

faiss/IndexRaBitQFastScan.cpp
alibeklfc: After the discussion, we have decided to not support search parameters as a result this const_cast was removed.

faiss/IndexRaBitQFastScan.cpp
alibeklfc: Added a dot

faiss/IndexRaBitQFastScan.cpp
alibeklfc: The d == 0 check throws an error early to reject non-meaningful data, since RaBitQ quantization does not make sense for zero-dimensional vectors

faiss/IndexFastScan.cpp
alibeklfc: Added a comment explaining why query offset is set to 0. There is either a single thread or a single query.

faiss/IndexFastScan.cpp
alibeklfc: Done

faiss/IndexAdditiveQuantizerFastScan.cpp
alibeklfc: query_offset is not always zero. It is passed correctly and can have other value. We are just not using it in AQFastsScan and PQFastScan

faiss/IndexRaBitQFastScan.h
alibeklfc: I have eliminated all const_cast usage by storing the query factors in the newly created ProcessingContext object. We now pass this object instead of NormTableScaler. ProcessingContext contains information about the NormScaler used by AQFastScan and the QueryFactorsProcessor used by RaBitQFastScan. The mutable keyword was removed from the database factors by moving the factor calculations to the add method. This approach is thread-safe since multithreading is only applied during search operations (indexing is single-threaded by design).

faiss/IndexFastScan.cpp
alibeklfc: We are overriding make_knn_handler in RaBitQFastScan, and template methods can't be virtual in C++. That is the reason, I have decided to use this architecture.

alibeklfc:
@alexanderguzhva thank you for your feedback, I have addressed your points.

faiss/IndexAdditiveQuantizerFastScan.cpp
alexanderguzhva: this is a safe-guard for classes that do not use this value. It MUST be zero, thus there should be a verification for this

faiss/impl/RaBitQUtils.h
alexanderguzhva: O_o
any modern C++ compiler will give you a warning. You can and you definitely should use `static inline` for functions in a header file, otherwise you'll get compilation errors in certain cases.

faiss/impl/RaBitQUtils.h
alexanderguzhva: same

alexanderguzhva:
@alibeklfc added some more
I'll review your changes as well

faiss/IndexAdditiveQuantizerFastScan.h
```diff
@@ -62,7 +62,11 @@ struct IndexAdditiveQuantizerFastScan : IndexFastScan {
 
     void compute_codes(uint8_t* codes, idx_t n, const float* x) const override;
 
-    void compute_float_LUT(float* lut, idx_t n, const float* x) const override;
+    void compute_float_LUT(
+            float* lut,
+            idx_t n,
+            const float* x,
+            const FastScanDistancePostProcessing& context = {}) const override;
```
alexanderguzhva: you should not speficy default values in overrides.
the note applies for similar cases in this PR as well

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    // Call parent implementation which will call compute_codes()
+    // compute_codes() now only needs to compute bit patterns
+    IndexFastScan::add(n, x);
+}
+
+void IndexRaBitQFastScan::compute_codes(uint8_t* codes, idx_t n, const float* x)
+        const {
+    FAISS_ASSERT(codes != nullptr);
```
alexanderguzhva, line 230: No.

Basically, the design of `compute_codes()` function is to produce a code for every given input vector that can be reconstructed back into a corresponding input vector (to the degree which it is possible, of course). I understand the situation in which `compute_codes()` refers to some internal table or dictionary, which is produced during the `train()` function and that does not depend on the input data itself. But this particular implementation does not store `rbq factors` for every code, which is an absolute must; `rbq factors` is not a table, it depends on a dataset. As a result, this implementation violates the design. 
The implementation of `sa_decode()` down below implies that its design is to work correctly only if you pass the dataset that was used for `train()`.

@mdouze do you have any comments?

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,528 @@
...
+    }
+}
+
+template <class C, bool with_id_map>
+void RaBitQHeapHandler<C, with_id_map>::handle(
+        size_t q,
+        size_t b,
+        simd16uint16 d0,
+        simd16uint16 d1) {
+    ALIGNED(32) uint16_t d32tab[32];
```
alexanderguzhva, line 473: the approach for this function looks good

faiss/impl/FastScanDistancePostProcessing.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Simple context object that holds processors for FastScan operations.
+ * */
+struct FastScanDistancePostProcessing {
+    /// Norm scaling processor for Additive Quantizers (nullptr if not needed)
+    const NormTableScaler* norm_scaler = nullptr;
+
+    /// Query factors data array for RaBitQ (nullptr if not needed)
+    /// Memory management is handled by the caller
+    rabitq_utils::QueryFactorsData* query_factors = nullptr;
```
alexanderguzhva, line 29: this leads back to the problem that the baseline class `IndexFastScan` becomes aware of rabitq. This could be solved by OOP inheritance. Just create a derived class, something like `RaBitQFastScanDistancePostProcessing` and override `IndexFastScan` functions that instantiate `FastScanDistancePostProcessing`.
Alternatively, please put `rabitq_utils::QueryFactorsData` class into a separate small independent file that does not carry complex dependencies, similar to `NormTableScaler`. 
I leave it up to you.

@mdouze any comments?

faiss/IndexAdditiveQuantizerFastScan.cpp
alibeklfc: In the new version, we have eliminated the query offset parameter passing. Now it is part of that context class. We are discussing about eliminating it completely and using pointer arithmetic

faiss/impl/RaBitQUtils.h
alibeklfc: The internal tests did not allow using static inline in the .h file. I have moved the function to the implementation file

faiss/IndexAdditiveQuantizerFastScan.h
alibeklfc: Resolved

faiss/IndexRaBitQFastScan.cpp
alibeklfc: Hi. I have resolved this issue.
The codes in compute_codes() now store both bit patterns and RaBitQ factors together. In sa_decode(), we extract both components from the embedded codes to reconstruct the original vector. I have added a stride parameter to the packing methods, so SIMD operations only process the bit patterns and skip the embedded factors. The codes variable is not stored after packing, so in add() method we populate the factors_storage from the codes so that handler have access to it.

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,584 @@
...
+    FAISS_ASSERT(
+            (metric_type == MetricType::METRIC_L2 ||
+             metric_type == MetricType::METRIC_INNER_PRODUCT));
+    if (n == 0) {
+        return;
+    }
+
+    // Hoist loop-invariant computations
+    const float* centroid_data = center.data();
+    const size_t bit_pattern_size = (d + 3) / 4;
```
alexanderguzhva: I'm not sure I follow why it is `(d + 3) / 4`. It is 1 bit per dimension, so it should be `(d + 7) / 8`, no? Otherwise, please add a comment why it is `(d + 3) / 4`, maybe I'm misinterpreting something

faiss/IndexRaBitQFastScan.cpp
```diff
@@ -0,0 +1,584 @@
...
+void IndexRaBitQFastScan::sa_decode(idx_t n, const uint8_t* bytes, float* x)
+        const {
+    const float* centroid_in =
+            (center.data() == nullptr) ? nullptr : center.data();
+    const uint8_t* codes = bytes;
+    FAISS_ASSERT(codes != nullptr);
+    FAISS_ASSERT(x != nullptr);
+
+    const float inv_d_sqrt = (d == 0) ? 1.0f : (1.0f / std::sqrt((float)d));
+    const size_t bit_pattern_size = (d + 3) / 4;
```
alexanderguzhva: same. why `(d + 3) / 4`?

faiss/impl/index_read.cpp
```diff
@@ -1224,6 +1225,27 @@ Index* read_index(IOReader* f, int io_flags) {
...
+        READVECTOR(idxqfs->center);
+        READ1(idxqfs->qb);
+        READVECTOR(idxqfs->factors_storage);
+        READ1(idxqfs->bbs);
+        READ1(idxqfs->ntotal2);
+        READ1(idxqfs->M2);
+        READ1(idxqfs->code_size);
+
+        // Need to initialize the FastScan base class fields
+        const size_t M_fastscan = (idxqfs->d + 3) / 4;
```
alexanderguzhva, line 1241: same, why `(d + 3) / 4`?

faiss/IndexRaBitQFastScan.cpp
alibeklfc: You are correct, I have changed it to (d + 7) / 8

faiss/IndexRaBitQFastScan.cpp
alibeklfc: You are correct, I have changed it to (d + 7) / 8

faiss/impl/index_read.cpp
alibeklfc: Here, (d + 3) / 4 is correct because it is the number of fastscan sub-quantizers. RaBitQ packs 4 dimensions (4 bits) into each sub-quantizer, so we need (d + 3) / 4 sub-quantizers

alexanderguzhva: not `(d + 1) / 2`, correct?

faiss/IndexRaBitQFastScan.cpp
alexanderguzhva: the next question: why this problem and similar ones were not caught by unit tests? What if other `(d + x) / y` things are not correct?

alibeklfc: Using (d + 3) / 4, I allocated more space than needed, which is why the tests passed. Recall is correct, we were allocating more space than needed. 

bit_pattern_size is the variable that calculates the size of the bit pattern portion in the codes. Since we store 1 bit per dimension, (d + 7) / 8 is the correct formula.

I have verified this experimentally: using values where x > 7 or y > 8 in (d + x) / y causes test failures, proving that (d + 7) / 8 is correct

faiss/impl/index_read.cpp
alibeklfc: (d + 3) / 4 is correct, because it is the number of fastscan sub-quantizers. One sub-quantizer is 4 dimensions (4 bits). Therefore, (d + 3) / 4 is correct

PR #4596, opened 2025-09-30
Implementation of IndexIVFRaBitQFastScan

alibeklfc:
Summary:
**Introduction**

This diff implements `IndexIVFRaBitQFastScan`, a high-performance approximate nearest neighbor search index that combines Inverted File (IVF) partitioning, RaBitQ quantization, and FastScan batch processing for significant performance improvements in vector similarity search.

**Implementation**
* **New Source and Header Files**: Added implementations for IndexIVFRaBitQFastScan, following a similar interface to IndexIVFRaBitQ.

* **Batched Processing**: The new index inherits from `IndexIVFFastScan`. It processes 32 database vectors simultaneously using SIMD operations.

* **Specialized Post-processing Handler**:  Extends the FastScan framework with RaBitQ-aware handler factories that automatically detect and optimize for RaBitQ indices, maintaining compatibility with existing IVF search parameters while enabling specialized optimizations.

* **LUT**: IndexRaBitQFastScan produces slightly different results than IndexIVFRaBitQ due to an extra quantization step in the IndexFastScan architecture. Computes lookup tables using query residuals (query minus coarse centroid) with probe-specific query factors, ensuring accurate distance calculations across different inverted list probes.
* **Centered parameter**: Implements both centered and non-centered query quantization modes

* **Query Offset Parameter**: Added a query offset parameter similarly as in D81787307

**Testing**
* Includes extensive test suite covering performance benchmarks, distance calculation correctness, recall measurements, and cross-validation against reference implementations

* All tests passed successfully, validating the correctness and robustness of IndexRaBitQFastScan.

**Results**
 ivfrabitq2
 ivfrabitq1
* **Performance Dependency**: Performance measurements confirm that IndexIVFRaBitQFastScan is notably faster than IndexIVFRaBitQ when the qb value is high. While the original IndexIVFRaBitQ experiences increased runtime with higher qb values, the fast scan variant maintains consistent runtime regardless of qb.

Differential Revision: D83264335


faiss/IndexFastScan.cpp
```diff
@@ -163,11 +164,23 @@ void estimators_from_tables_generic(
         size_t k,
         typename C::T* heap_dis,
         int64_t* heap_ids,
-        const NormTableScaler* scaler) {
+        const NormTableScaler* scaler,
+        const QueryFactorsData* query_factors = nullptr) {
     using accu_t = typename C::T;
 
+    // Check if this is RaBitQ for specialized distance computation
+    const IndexRaBitQFastScan* rabitq_index =
```
alexanderguzhva: This totally violates the fundamentals of object-oriented programming. The base class `IndexFastScan` must know nothing about a derived class `IndexRaBitQFastScan`. Please consider reorganizing this.

faiss/IndexFastScan.cpp
```diff
@@ -454,25 +539,55 @@ void IndexFastScan::search_implem_12(
...
-    handler->disable = bool(skip & 2);
-    handler->normalizers = normalizers.get();
+    // Check if this is an IndexRaBitQFastScan to use RaBitQHeapHandler
+    const IndexRaBitQFastScan* rabitq_index =
+            dynamic_cast<const IndexRaBitQFastScan*>(this);
+
+    if (rabitq_index != nullptr) {
+        // Use RaBitQHeapHandler for RaBitQ (direct processing, no
+        // post-processing needed)
+        RaBitQHeapHandler<C> rabitq_handler(
```
alexanderguzhva: same story here and below

faiss/IndexIVFFastScan.cpp
```diff
@@ -400,7 +402,29 @@ SIMDResultHandlerToFloat* make_knn_handler(
         idx_t k,
         float* distances,
         idx_t* labels,
-        const IDSelector* sel) {
+        const IDSelector* sel,
+        const IndexIVFFastScan* index = nullptr,
+        idx_t query_offset = 0) {
+    // Check if this is IndexIVFRaBitQFastScan and create specialized handler
+    if (index) {
+        auto* rabitq_index = dynamic_cast<const IndexIVFRaBitQFastScan*>(index);
```
alexanderguzhva: same story about `IndexIVFFastScan` and `IndexIVFRaBitQFastScan`

faiss/IndexIVFRaBitQFastScan.cpp
```diff
@@ -0,0 +1,694 @@
...
+    query_factors.qr_to_c_L2sqr = fvec_norm_L2sqr(residual, d);
+
+    for (size_t j = 0; j < d; j++) {
+        rotated_q[j] = residual[j];
+    }
+
+    float v_min = std::numeric_limits<float>::max();
+    float v_max = std::numeric_limits<float>::lowest();
+
+    if (centered) {
```
alexanderguzhva: do I get it correct that this is a copy-pasted code from `impl/RaBitQuantizer.cpp`? This creates a need to synchronize the changes between this code and `impl/RaBitQuantizer.cpp`. Please reorganize the code to have only 1 instance of these facilities, if possible

alexanderguzhva:
@alibeklfc I would strongly recommend refactoring the candidate code

alibeklfc:
@alexanderguzhva Thank you for your valuable feedback!

I have addressed your comments. The IndexFastScan and IndexIVFFastScan do not know anything about the child classes now. I have made the make_knn_handler function part of the main class and we override it in the child classes.

I have also created the RaBitQUtils namespace to eliminate code duplication.

alexanderguzhva:
@alibeklfc thanks, let me go through both diffs carefully. Would you like to have #4595 commited prior to #4596?

alibeklfc:
@alexanderguzhva #4595 should be committed before #4596, but I want both of them to commit as soon as possible for benchmarking and internal usage. Thank you!

alexanderguzhva:
@alibeklfc let's deal with #4595 first, because might need some work

faiss/IndexIVFFastScan.cpp
```diff
@@ -971,6 +994,11 @@ void IndexIVFFastScan::search_implem_10(
             handler.ntotal = ls;
             handler.id_map = ids.get();
 
+            // Set context information for handlers that need additional data
+            std::vector<int> probe_map(1);
```
alexanderguzhva: this allocation operation is a costly one. Would it be possible to allocate this vector once, before `for (idx_t i = 0; i < n; i++) {` line?

Also, if you need to pass vectors (like down below), then the best strategy is the following:
```C++
std::vector<int> some_vector;
some_vector.reserve(initial_guess);

for (size_t i = 0; i < n; i++) {
    ...
    // the trick is that an internal vector implementation may skip immediately shrinking
    //   the amount of allocated memory, so this operation is very cheap, especially if
    //   your `initial_guess` size covers all new sizes  
    some_vector.resize(new_size);

    // use the vector
    ...
}
```
This allows you not to waste your time on memory allocations

faiss/IndexIVFFastScan.h
```diff
@@ -91,8 +108,56 @@ struct IndexIVFFastScan : IndexIVF {
...
 
+    /** Add vectors with specific IDs to the index
+     *
+     * @param n     number of vectors to add
+     * @param x     vectors to add (n * d)
+     * @param xids  IDs for the vectors (n)
+     */
     void add_with_ids(idx_t n, const float* x, const idx_t* xids) override;
 
+   protected:
```
alexanderguzhva: I don't believe that it is a good idea to insert `protected` inside the class
```C++
class Foo {
public:
  ...
protected:
  // lolwhat? Move `protected:` to the end, after the `public:`, but before the `private:`
public:
  ...
};
```

faiss/IndexIVFRaBitQFastScan.h
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Default number of bits to quantize a query with
+    uint8_t qb = 8;
+
+    /// Use zero-centered scalar quantizer for queries
+    bool centered = false;
+
+    /// Extracted factors storage for batch processing
+    /// Size: ntotal, stores factors separately from packed codes
+    std::vector<FactorsData> factors_storage;
```
alexanderguzhva, line 63: this needs to be changed, because the current scenario makes the it thread-unsafe.
Please consider the following possible pattern:
```C++
struct SAPreprocessed {};
struct IndexIVFFastScan {
    virtual void sa_encode_preprocess(
            idx_t n,
            const uint8_t* flat_codes,
            idx_t start_global_idx,
            SAPreprocessed** sa_preprocessed
    ) = {};

    virtual void IndexIVFFastScan::consume_preprocessed(
            const SAPreprocessed* preprocessed
    ) = {};
}; 

struct RaBitQSAPreprocessed: public SAPreprocessed {
    std::vector<float> factors;
};

void IndexIVFRaBitQFastScan::sa_encode_preprocess(..., SAPreprocessed** preprocessed) {
    RaBitQSAPreprocessed* rabitq_preprocessed = new RaBitQSAPreprocessed;
    // populate rabitq_preprocessed
    ...

   *preprocessed = rabitq_preprocessed;
}

void IndexIVFRaBitQFastScan::consume_preprocessed(..., const SAPreprocessed* preprocessed) {
    RaBitQSAPreprocessed* rabitq_preprocessed = dynamic_cast<RaBitQSAPreprocessed*>(preprocessed);
    // use rabitq_preprocessed
}
```

alexanderguzhva (changes requested):

faiss/IndexIVFFastScan.cpp
alibeklfc: Resolved

faiss/IndexIVFFastScan.h
alibeklfc: resolved

faiss/IndexIVFRaBitQFastScan.h
alibeklfc: Add() is not thread safe for most (?) Index algorithms in FAISS. I believe that the adding is single threaded and searching is multithreaded in FAISS by design. Lets discuss this in the next meeting.

alexanderguzhva: @alibeklfc  I see. The source of my confusion is the method name. I kinda assumed that `sa_encode_preprocess()` will be used for `sa_encode()`. Would it be possible to rename the method name into something unrelated to `sa_encode()`, please?

PR #4600, opened 2025-10-03
Fix memory bloat in IndexBinaryHash search due to argument copying in dispatch_HammingComputer

tracywong117:
## Summary
Fix excessive memory usage (3x index size) in IndexBinaryHash search operations caused by pass-by-value semantics in variadic template dispatch.

## Problem
When searching a 114GB IndexBinaryHash index, RAM usage peaked at ~342GB. 
Root cause:
- dispatch_HammingComputer used Types... args (pass-by-value)
- Run_search_single_query::f used Types... args (pass-by-value)

This caused arguments to be copied for each instantiation, inflating memory usage.

## Solution
- faiss/IndexBinaryHash.cpp: Update Run_search_single_query call sites to pass pointers to arguments.

## Testing
- Tested with 114GB IndexBinaryHash index
- RAM usage reduced from ~342GB to ~114GB (expected size)
- Search results remain identical
- No performance regression observed

## Backward Compatibility
- Function name and external API remain stable for Faiss users
- Behavior unchanged except reduced memory footprint during search

### DIFF
--- a/faiss/IndexBinaryHash.cpp
+++ b/faiss/IndexBinaryHash.cpp
@@ -191,7 +191,8 @@ void search_single_query(
         size_t& nlist,
         size_t& ndis) {
     Run_search_single_query r;
-    dispatch_HammingComputer(index.code_size, r, &index, &q, &res, &n0, &nlist, &ndis);
+    dispatch_HammingComputer(
+            index.code_size, r, &index, &q, &res, &n0, &nlist, &ndis);
 }
 
 } // anonymous namespace


alexanderguzhva:
@tracywong117 great, thanks! If you believe that there will be no side effects from this change, then please feel free to replace the original dispatch version with a new one without creating an additional facility

mdouze:
Hi @tracywong117 
Thanks for catching this bug! 
The preferred fix would be to pass pointers rather than making a new dispatching function. This would make the pass-by-value vs. pass-by-reference more explicit than the `_perfect` fix in the version that you suggest. 
Thanks!

tracywong117:
Hi @mdouze @alexanderguzhva 

Thanks for the feedback! I'll update the PR to use pointers instead of perfect forwarding. Will push the changes shortly.

junjieqi:
@tracywong117  It is an awesome fix! Thank for making the change! Would you mind fixing the format issue so we could merge the code? Thank!

PR #4603, opened 2025-10-06
Add IndexBinaryIDMap2 support to index binary factory

### LINKED ISSUE #3875: Add support for IndexBinaryIDMap2 to index_binary_factory
# Summary

# Reproduction instructions
```
code_size = 16
bindex = faiss.IndexBinaryIDMap2(faiss.IndexBinaryFlat(code_size))
n = 10
d = code_size // 8
ids = np.arange(n)
x = np.zeros((n, d), dtype=np.uint8)
bindex.add_with_ids(x, ids)
```
should be possible with
```
bindex = faiss.index_binary_factory("IDMap2,BFlat")
```
but index_binary_factory does not currently parse for IDMap2 like index_factory does: https://github.com/facebookresearch/faiss/blob/main/faiss/index_factory.cpp#L908

ashishresu2001:
## Summary
- Add support for IDMap2/IDMap prefix and suffix parsing in index_binary_factory
- Normalize description string to handle whitespace
- Add unit tests for IDMap2 prefix, suffix, and whitespace cases

## Testing
- Ran: pytest tests/test_binary_factory.py

Fixes #3875

### DIFF
--- a/faiss/index_factory.cpp
+++ b/faiss/index_factory.cpp
@@ -937,66 +937,51 @@ IndexBinary* index_binary_factory(
         bool own_invlists) {
     IndexBinary* index = nullptr;
 
-    // Handle IDMap2 and IDMap wrappers (prefix or suffix)
+    std::smatch sm;
     std::string desc_str(description);
-    // Remove whitespace
-    desc_str.erase(
-            std::remove_if(desc_str.begin(), desc_str.end(), ::isspace),
-            desc_str.end());
 
-    // Check for IDMap2 or IDMap as prefix
-    if (desc_str.rfind("IDMap2,", 0) == 0 && desc_str.size() > 7) {
-        std::string sub_desc = desc_str.substr(7);
+    // Handle IDMap2 and IDMap wrappers (prefix or suffix)
+    if (re_match(desc_str, "(.+),IDMap2", sm) ||
+        re_match(desc_str, "IDMap2,(.+)", sm)) {
         IndexBinary* sub_index =
-                index_binary_factory(d, sub_desc.c_str(), own_invlists);
-        return new IndexBinaryIDMap2(sub_index);
+                index_binary_factory(d, sm[1].str().c_str(), own_invlists);
+        IndexBinaryIDMap2* idmap2 = new IndexBinaryIDMap2(sub_index);
+        idmap2->own_fields = true;
+        return idmap2;
     }
-    if (desc_str.rfind("IDMap,", 0) == 0 && desc_str.size() > 6) {
-        std::string sub_desc = desc_str.substr(6);
+
+    if (re_match(desc_str, "(.+),IDMap", sm) ||
+        re_match(desc_str, "IDMap,(.+)", sm)) {
         IndexBinary* sub_index =
-                index_binary_factory(d, sub_desc.c_str(), own_invlists);
-        return new IndexBinaryIDMap(sub_index);
-    }
-    // Check for IDMap2 or IDMap as suffix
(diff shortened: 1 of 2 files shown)


mnorris11:
Hi @ashishresu2001 , sorry for the delay, I will review this today

faiss/index_factory.cpp
```diff
@@ -934,32 +937,69 @@ IndexBinary* index_binary_factory(
         bool own_invlists) {
     IndexBinary* index = nullptr;
 
+    // Handle IDMap2 and IDMap wrappers (prefix or suffix)
+    std::string desc_str(description);
+    // Remove whitespace
+    desc_str.erase(
+            std::remove_if(desc_str.begin(), desc_str.end(), ::isspace),
+            desc_str.end());
```
mnorris11: We do not need this part, we can assume it is incorrect if there are incorrect spaces. (We can follow the pattern of the other parts of the index_factory which do not have this handling)

faiss/index_factory.cpp
```diff
@@ -934,32 +937,69 @@ IndexBinary* index_binary_factory(
...
+
+    // Check for IDMap2 or IDMap as prefix
+    if (desc_str.rfind("IDMap2,", 0) == 0 && desc_str.size() > 7) {
```
mnorris11: Are you able to use the same approach of `re_match` as the other IDMap2 handling for non-binary index? i.e. like here: https://github.com/facebookresearch/faiss/blob/041ac84318dac299739e568fafee318317ede6ef/faiss/index_factory.cpp#L703-L709
(ideally we can keep it consistent)
This lets us consolidate to fewer conditions, since prefix and suffix should behave the same.

faiss/index_factory.cpp
```diff
@@ -934,32 +937,69 @@ IndexBinary* index_binary_factory(
...
+        desc_str.compare(desc_str.size() - 6, 6, ",IDMap") == 0) {
+        std::string sub_desc = desc_str.substr(0, desc_str.size() - 6);
+        IndexBinary* sub_index =
+                index_binary_factory(d, sub_desc.c_str(), own_invlists);
+        return new IndexBinaryIDMap(sub_index);
+    }
+
     int ncentroids = -1;
     int M, nhash, b;
+    const char* desc_cstr = desc_str.c_str();
```
mnorris11: Why did we need to convert to a C style string here, if the input is a C style string?

mnorris11:
Thanks for this contribution @ashishresu2001 ! I just left a few small comments: we can make it closer to the existing approach. Otherwise looks great.

tests/test_binary_factory.py
```diff
@@ -42,3 +42,18 @@ def test_factory_MultiHash(self):
...
+
+    def test_factory_IDMap2_suffix(self):
+        index = faiss.index_binary_factory(16, "BFlat,IDMap2")
+        assert isinstance(index, faiss.IndexBinaryIDMap2)
+        assert index.index.code_size == 2
+
+    def test_factory_IDMap2_with_spaces(self):
+        index = faiss.index_binary_factory(16, "  IDMap2 ,  BFlat  ")
+        assert isinstance(index, faiss.IndexBinaryIDMap2)
+        assert index.index.code_size == 2
```
mnorris11: We are very grateful you added tests! We can remove this spaces one though.

ashishresu2001: Hi @mnorris11 , Thanks for the review. I’ve made the requested updates. Removed whitespace handling, switched to re_match for consistency, dropped the redundant C-string conversion, and removed the extra test.

PR #4606, opened 2025-10-12
Integrate Panorama into `IndexIVFFlatPanorama`

AlSchlo:
# Panorama: Fast-Track Nearest Neighbors for IVFFlat

This PR integrates [Panorama](https://arxiv.org/abs/2510.00566) into Faiss' `IndexIVFFlat`, as a new index called `IndexIVFFlatPanorama`.

## What is Panorama?

Panorama is a machine learning-driven approach that accelerates the verification/refinement bottleneck in Approximate Nearest Neighbor Search (ANNS) through learned orthogonal transforms and incremental distance computation. It uses data-adaptive transforms that concentrate over 90% of signal energy into the first half of dimensions, enabling early candidate pruning with partial distance computations.

## How it benefits IVFFlat

Panorama addresses a critical issue: in modern high-dimensional neural embeddings, the refinement phase accounts for 75-99% of query latency. By incrementally computing distances and pruning candidates whose lower bound exceeds the running threshold, Panorama dramatically reduces wasted computation.

The integration uses a level-major memory layout that processes candidates in batches, improving cache locality and reducing branch mispredictions. This reorganization groups vectors into batches and processes them level-by-level rather than vector-by-vector, consolidating pruning decisions and regularizing control flow.

## Expected speedups

On IVFFlat specifically, Panorama achieves dramatic speedups of **5-20×** across diverse datasets with **no recall loss**, thanks to the contiguous memory access patterns and efficient SIMD utilization.

The following graph is generated by running the demo in the PR:

image

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -30,6 +30,7 @@ set(FAISS_SRC
   IndexIVFPQFastScan.cpp
   IndexIVFPQR.cpp
   IndexIVFRaBitQ.cpp
+  IndexIVFRaBitQFastScan.cpp
   IndexIVFSpectralHash.cpp
   IndexLSH.cpp
   IndexNNDescent.cpp
@@ -131,6 +132,7 @@ set(FAISS_HEADERS
   IndexIVFPQFastScan.h
   IndexIVFPQR.h
   IndexIVFRaBitQ.h
+  IndexIVFRaBitQFastScan.h
   IndexIVFSpectralHash.h
   IndexLSH.h
   IndexNeuralNetCodec.h
--- a/faiss/IndexIVF.h
+++ b/faiss/IndexIVF.h
@@ -497,12 +497,12 @@ struct InvertedListScanner {
     /// compute a single query-to-code distance
     virtual float distance_to_code(const uint8_t* code) const = 0;
 
-    /** scan a set of codes, compute distances to current query and
+    /** scan a set of codes, compute distances to current query, and
      * update heap of results if necessary. Default implementation
      * calls distance_to_code.
      *
-     * @param n      number of codes to scan
-     * @param codes  codes to scan (n * code_size)
+     * @param n          number of codes to scan
+     * @param codes      codes to scan (n * code_size)
      * @param ids        corresponding ids (ignored if store_pairs)
      * @param distances  heap distances (size k)
      * @param labels     heap labels (size k)
--- a/faiss/IndexIVFAdditiveQuantizerFastScan.cpp
+++ b/faiss/IndexIVFAdditiveQuantizerFastScan.cpp
@@ -14,6 +14,7 @@
 
 #include <faiss/impl/AuxIndexStructures.h>
 #include <faiss/impl/FaissAssert.h>
+#include <faiss/impl/FastScanDistancePostProcessing.h>
 #include <faiss/impl/LookupTableScaler.h>
 #include <faiss/impl/pq4_fast_scan.h>
 #include <faiss/invlists/BlockInvertedLists.h>
@@ -212,7 +213,9 @@ void IndexIVFAdditiveQuantizerFastScan::estimate_norm_scale(
     size_t index_nprobe = nprobe;
     nprobe = 1;
     CoarseQuantized cq{index_nprobe, coarse_dis.data(), coarse_ids.data()};
-    compute_LUT(n, x, cq, dis_tables, biases);
+    FastScanDistancePostProcessing empty_context{};
+
+    compute_LUT(n, x, cq, dis_tables, biases, empty_context);
     nprobe = index_nprobe;
 
     float scale = 0;
@@ -314,8 +317,10 @@ void IndexIVFAdditiveQuantizerFastScan::search(
     }
 
     NormTableScaler scaler(norm_scale);
+    FastScanDistancePostProcessing context;
+    context.norm_scaler = &scaler;
     IndexIVFFastScan::CoarseQuantized cq{nprobe};
-    search_dispatch_implem(n, x, k, distances, labels, cq, &scaler);
+    search_dispatch_implem(n, x, k, distances, labels, cq, context);
 }
 
 /*********************************************************
@@ -383,7 +388,8 @@ void IndexIVFAdditiveQuantizerFastScan::compute_LUT(
         const float* x,
         const CoarseQuantized& cq,
         AlignedTable<float>& dis_tables,
-        AlignedTable<float>& biases) const {
+        AlignedTable<float>& biases,
+        const FastScanDistancePostProcessing&) const {
     const size_t dim12 = ksub * M;
     const size_t ip_dim12 = aq->M * ksub;
     const size_t nprobe = cq.nprobe;
--- a/faiss/IndexIVFAdditiveQuantizerFastScan.h
+++ b/faiss/IndexIVFAdditiveQuantizerFastScan.h
@@ -102,7 +102,8 @@ struct IndexIVFAdditiveQuantizerFastScan : IndexIVFFastScan {
             const float* x,
             const CoarseQuantized& cq,
             AlignedTable<float>& dis_tables,
-            AlignedTable<float>& biases) const override;
+            AlignedTable<float>& biases,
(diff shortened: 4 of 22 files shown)


AlSchlo:
Core implementation done.

Next steps: tests & demo/benches & design decisions in PR description.

faiss/invlists/InvertedLists.h
```diff
@@ -276,6 +276,67 @@ struct ArrayInvertedLists : InvertedLists {
...
+    /// by Faiss (i.e. `InvertedListsIterator` API). The iterator would require
+    /// to allocate and reassemble the vector at each call.
+    /// Hence, we override this method to throw an error, this effectively
+    /// disables the `iterate_codes` and `iterate_codes_range` methods.
+    InvertedListsIterator* get_iterator(
+            size_t list_no,
+            void* inverted_list_context = nullptr) const override;
+
+    /// Reconstructs a single code from level-oriented storage to flat format.
+    /// Returns pointer to internal buffer - valid until next call to this
```
mdouze: This is not thread-safe. 
It should return a pointer to a newly allocated buffer that is deallocated with `release_codes`

faiss/invlists/InvertedLists.h
```diff
@@ -276,6 +276,67 @@ struct ArrayInvertedLists : InvertedLists {
     ~ArrayInvertedLists() override;
 };
 
+/// Level-oriented storage as defined in the IVFFlat section of Panorama
+/// (https://www.arxiv.org/pdf/2510.00566).
+struct ArrayInvertedListsPanorama : ArrayInvertedLists {
+    static constexpr size_t kBatchSize = 128;
```
mdouze, line 282: Would it be feasible to express this as a `BlockInvertedLists` ? 
https://github.com/facebookresearch/faiss/blob/main/faiss/invlists/BlockInvertedLists.h#L33
This is a collective representation for a fixed number of vectors (similar to Panorama's batches).

tests/test_panorama.py
```diff
@@ -0,0 +1,454 @@
...
+            index.add(xb)
+        if nprobe is not None:
+            index.nprobe = nprobe
+        return index
+
+    def create_panorama(self, d, nlist, nlevels, xt, xb=None, nprobe=None, make_direct_map=False):
+        """Create and initialize IndexIVFFlatPanorama."""
+        quantizer = faiss.IndexFlatL2(d)
+        index = faiss.IndexIVFFlatPanorama(quantizer, d, nlist, nlevels)
+        # Keep quantizer alive by adding it to referenced_objects
```
mdouze: please add `IndexIVFFlatParorama` here https://github.com/facebookresearch/faiss/blob/main/faiss/python/__init__.py#L174

tests/test_panorama.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+class TestIndexIVFFlatPanorama(unittest.TestCase):
+    """Test Suite for IndexIVFFlatPanorama."""
+
+    # Helper methods for index creation and data generation
+    
+    def generate_data(self, d, nt, nb, nq, seed=42):
+        """Generate random training, database, and query vectors."""
+        np.random.seed(seed)
```
mdouze: What about using the SyntheticDataset? 
The reason why this dataset exists is because random uniform data does not have a typical behavior for vector search -- in particular because the intrinsic dimension is the same as the actual dimension.

faiss/invlists/InvertedLists.h
aknayar: Ah thanks! I wasn't aware of `release_codes`, should be fixed now!

tests/test_panorama.py
aknayar: Done!

tests/test_panorama.py
aknayar: Good point, updated!

alexanderguzhva:
@AlSchlo Thanks for the code and the article!
Do you plan to add the code for the Section 4 from your article?

AlSchlo:
@alexanderguzhva For all methods that don't need quantization, PCA works well empirically. This is simply a linear transform that can be defined upstream (we will push demo code about this very soon). The Cayley transforms are a nice knob for IVFPQ. We can make a PR for that too, but in a second stage.

Next PR coming after this one will be HNSWPanorama :-)

faiss/invlists/InvertedLists.h
AlSchlo: I just went through this tonight and it seems that this would require an override of `add_entries` as it gets codes that are already packed. This kind of defeats the purpose IMO.

faiss/IndexIVFFlatPanorama.cpp
```diff
@@ -0,0 +1,348 @@
...
+            const ArrayInvertedListsPanorama* storage,
+            bool store_pairs,
+            const IDSelector* sel)
+            : InvertedListScanner(store_pairs, sel), vd(vd), storage(storage) {
+        keep_max = vd.is_similarity;
+        code_size = vd.d * sizeof(float);
+        cum_sums.resize(storage->n_levels + 1);
+    }
+
+    const float* xi;
```
alexanderguzhva: please initialize members with default values, just in case

faiss/IndexIVFFlatPanorama.cpp
```diff
@@ -0,0 +1,348 @@
...
+    std::vector<float> cum_sums;
+    float q_norm;
```
alexanderguzhva: same

faiss/IndexIVFFlatPanorama.cpp
```diff
@@ -0,0 +1,348 @@
...
+                        fvec_inner_product(query_level, yj, actual_level_width);
+
+                exact_distances[idx] -= 2.0f * dot_product;
+
+                float cum_sum = level_cum_sums[idx];
+                float cauchy_schwarz_bound = 2.0f * cum_sum * query_cum_norm;
+                float lower_bound = exact_distances[idx] - cauchy_schwarz_bound;
+
+                active_indices[next_active] = idx;
+                next_active += C::cmp(threshold, lower_bound);
```
alexanderguzhva: this is an implicit bool-to-integer conversion, I'd write `C::cmp(...) ? 1 : 0`, just to avoid compiler warnings

alexanderguzhva:
@AlSchlo Is there a plan to put the core algorithm 1 from section 2 of the article into a standalone facility, or would you guys prefer to add specialized versions of it, once needed?

mnorris11:
Btw, we can usually ignore the "internal - builds & tests" failing signal. The internal build system is not CMake, so we have to edit some other files which do not get exported to OSS. I will edit those on the imported version after edits are done here.

AlSchlo:
@alexanderguzhva Yes, let's keep it demand-driven. I'm not a big fan of over-abstracting unless we run into significant duplication. That said, as we start supporting more distance metrics & indexes, it will probably become necessary.

mnorris11:
After the quick comments from Alexandr, I will re-import and this one should be good to merge.

aknayar:
@mnorris11 All comments should be addressed now, thank you all for the helpful reviews!

mnorris11:
Still reviewing internally. Targeting merge in the next couple business days.

mnorris11:
Hi @aknayar @AlSchlo , when you have time, we had a question come up about the distance_to_code in IVFFlatPanorama: Is this unimplemented due to algorithmic design, or just something we can get to later?

AlSchlo:
Hi @mnorris11 ,

In standard `IVFFlat`, the function is simply:
```
float distance_to_code(const uint8_t* code) const final {
    const float* yj = (float*)code;
    return vd(xi, yj);
}
```

However, with Panorama’s storage layout, you would need to reconstruct the full vector because the data is stored in batches. The same issue existed in FastScan, although it appears to have been resolved in [this PR](https://github.com/facebookresearch/faiss/pull/4821).

When I checked the codebase, `distance_to_code` was only used within iterators, which is suboptimal for traversing candidate points. That said, it could be implemented if direct access is required for different reasons (e.g. public API).

mnorris11:
Thanks @AlSchlo! It is not a priority on our end, but good to know.

PR #4611, opened 2025-10-15
Fix ARM64 compilation error in IndexRaBitQFastScan

alibeklfc:
Summary:
Use explicit template parameter for std::min to resolve type mismatch on ARM64 platforms.

On ARM64 (macOS), idx_t is typically int64_t/long long. When subtracting size_t from idx_t, the result gets promoted to unsigned long long on ARM64, but remains unsigned long on x64. This causes std::min to fail template argument deduction on ARM64 because it receives conflicting types: size_t (unsigned long) and unsigned long long.

The fix uses std::min<size_t> with an explicit template parameter, which eliminates the need for type deduction and allows the compiler to implicitly convert both arguments to size_t. This is the idiomatic C++ solution for this pattern and adds zero runtime overhead.

Differential Revision: D84721816


alexanderguzhva:
lgtm

PR #4619, opened 2025-10-18
Remove use of __AMDGCN_WAVEFRONT_SIZE macro with it being deprecated in ROCm 7

kevinjoseph1995:
# Summary:
Usage of `__AMDGCN_WAVEFRONT_SIZE` is a compilation error when using amdclang++ shipped with ROCm 7.x.x. For versions >= 7 we delegate to rocprim to get the right constexpr value that can be used in device code. We continue using __AMDGCN_WAVEFRONT_SIZE for older ROCm versions to maintain compatibility.

Note that the usage of kWarpSize in host code is still non-sensical however the current usage of this variable seems to always be in device functions.

See https://rocm.docs.amd.com/en/docs-7.0.2/about/release-notes.html#amdgpu-wavefront-size-compiler-macro-deprecation for deprecation notice.


mnorris11:
Thanks for this contribution @kevinjoseph1995!
Right now, our ROCm runner in the CI is not running. We are working on getting it back up and running again. After it is, we can run this in the CI and then get it merged.

kevinjoseph1995:
Got it, thanks for letting me know! I’ll wait for the ROCm runner to be back online.

mnorris11:
We are enabling the ROCM runner again, I will check this again soon, thanks for your patience!

kevinjoseph1995:
> We are enabling the ROCM runner again, I will check this again soon, thanks for your patience!

That's awesome, my hope is that this still works as expected. I can try this out with ROCm 7.2 later this week on my side, just to make sure that this still works.

mnorris11:
ROCM runner enabled again, so re-triggering CI

kevinjoseph1995:
Looks like the ROCm build succeeded with this change, however it looks like it was using ROCm 6.2 https://github.com/facebookresearch/faiss/blob/main/.github/actions/build_cmake/action.yml#L115. Would it make sense to bump that to the latest version(7.2.1) as part of this PR?

ItsPitt:
> Looks like the ROCm build succeeded with this change, however it looks like it was using ROCm 6.2 https://github.com/facebookresearch/faiss/blob/main/.github/actions/build_cmake/action.yml#L115. Would it make sense to bump that to the latest version(7.2.1) as part of this PR?

We have plans to push to ROCm 7+ on an upcoming PR.

PR #4621, opened 2025-10-20
Integrate Panorama into `IndexHNSWFlatPanorama`

AlSchlo:
This PR introduces **Panorama** into `HNSWFlat`, following our [paper](https://arxiv.org/pdf/2510.00566). Panorama achieves up to **4× lower latency** on higher-dimensional data, making it a great option for medium-sized datasets that don't benefit much from quantization.

Below are some benchmarks on **SIFT-128**, **GIST-960**, and **synthetic 2048-dimensional** data. I recommend checking out the paper for more results. As expected, Panorama is not a silver bullet when combined with HNSW—it’s only worthwhile for high-dimensional data.

It might be worth considering, in the future, adding a function that dynamically sets the number of levels. However, this would require reorganizing the cumulative sums.

## SIFT-128

**Note:** SIFT-128 performs slightly worse here than in our paper because we use 8 levels, whereas the paper explored several level configurations. Eight levels introduce quite a bit of overhead for 128-dimensional data, but I kept it consistent across all benchmarks for comparison.

bench_hnsw_flat_panorama_SIFT1M

## GIST-960
bench_hnsw_flat_panorama_GIST1M

## Synthetic-2048
bench_hnsw_flat_panorama_Synthetic2048D


faiss/impl/HNSW.cpp
```diff
@@ -966,13 +1219,28 @@ HNSWStats HNSW::search(
...
-    if (bounded_queue) { // this is the most common branch
+    if (bounded_queue) { // this is the most common branch, for now we only
+                         // support Panorama search in this branch
         MinimaxHeap candidates(ef);
 
         candidates.push(nearest, d_nearest);
 
-        search_from_candidates(
-                *this, qdis, res, candidates, vt, stats, 0, 0, params);
+        if (is_panorama) {
```
AlSchlo: An alternative would be to make `search_from_candidates` a class function, and then create an `HNSWPanorama` class that overrides `search_from_candidates`.

We could then store `cum_sums` and other relevant state within the `HNSWPanorama` object. This is my preferred option, but it would mean that the function is no longer in the `faiss` namespace. I wanted to check whether that has any consequences.

faiss/IndexHNSW.cpp
```diff
@@ -647,6 +647,155 @@ IndexHNSWFlat::IndexHNSWFlat(int d, int M, MetricType metric)
     is_trained = true;
 }
 
+IndexHNSWFlat::IndexHNSWFlat(int d, int M, IndexFlat* storage)
+        : IndexHNSW(storage, M) {
+    own_fields = true;
```
mdouze: In general when storage is passed as a parameter, the fields are not owned by default.

faiss/IndexHNSW.cpp
```diff
@@ -647,6 +647,155 @@ IndexHNSWFlat::IndexHNSWFlat(int d, int M, MetricType metric)
...
+
+    float sum = 0.0f;
+    dst_cum_sums[n_levels] = 0.0f;
+    for (int level = n_levels - 1; level >= 0; level--) {
+        int start_idx = level * level_width;
+        int end_idx = std::min(start_idx + level_width, d);
+        for (int j = start_idx; j < end_idx; j++) {
+            sum += x[j] * x[j];
+        }
+        dst_cum_sums[level] = sqrt(sum);
```
mdouze, line 674: is there a reason to do a sqrt (ie. use L2 distance instead of squared L2 distance as usual)?

faiss/IndexHNSW.cpp
```diff
@@ -647,6 +647,155 @@ IndexHNSWFlat::IndexHNSWFlat(int d, int M, MetricType metric)
...
+    IndexHNSWFlat::add(n, x);
+}
+
+void IndexHNSWFlatPanorama::reset() {
+    cum_sums.clear();
+    IndexHNSWFlat::reset();
+}
+
+size_t IndexHNSWFlatPanorama::remove_ids(const IDSelector& sel) {
+    idx_t j = 0;
```
mdouze: Does this function make sense? remove is not supported by HNSW because it is hard to "stitch" the graph structure.

faiss/IndexHNSW.h
```diff
@@ -123,6 +123,68 @@ struct IndexHNSW : Index {
...
+/** Panorama implementation of IndexHNSWFlat following
+ * https://www.arxiv.org/pdf/2510.00566.
+ *
+ * Unlike cluster-based Panorama, the vectors have to be higher dimensional
+ * (i.e. typically d > 512) and/or be able to compress a lot of their energy in
+ * the early dimensions to be effective. This is because HNSW accesses vectors
+ * in a random order, which makes cache misses dominate the distance computation
+ * time.
+ *
+ * The num_levels parameter controls the granularity of progressive distance
```
mdouze: Would it be possible to call it something else than num_levels? HNSW has its notion of levels -- the number of hierachy levels.

faiss/IndexHNSW.h
```diff
@@ -123,6 +123,68 @@ struct IndexHNSW : Index {
...
+ * vanilla HNSW, as it uses partial distance computations with progressive
+ * refinement bounds. Instead of computing full distances immediately for all
+ * candidates, Panorama maintains lower and upper bounds that are incrementally
+ * tightened across refinement levels. Candidates are inserted into the search
+ * beam using approximate distance estimates (LB+UB)/2 and are only fully
+ * evaluated when they survive pruning and enter the result heap. This allows
+ * the algorithm to prune unpromising candidates early using Cauchy-Schwarz
+ * bounds on partial inner products. Hence, recall is not guaranteed to be the
+ * same as vanilla HNSW due to the heterogeneous precision within the search
+ * beam (exact vs. partial distance estimates affecting traversal order).
```
mdouze, line 151: Does this mean that the batch size on which the distances are computed is at most the out-degree of the HNSW graph (set to 64 by default)?

faiss/impl/HNSW.cpp
```diff
@@ -588,6 +590,28 @@ void HNSW::add_with_locks(
 using MinimaxHeap = HNSW::MinimaxHeap;
 using Node = HNSW::Node;
 using C = HNSW::C;
+
+/** Helper to parse search parameters from HNSW and SearchParameters */
```
mdouze: please call this something else than parse, which implies text.
extract_search_params ?

faiss/impl/HNSW.h
```diff
@@ -10,6 +10,7 @@
 #include <queue>
 #include <unordered_set>
 #include <vector>
+#include "faiss/impl/DistanceComputer.h"
```
mdouze: please use <>

faiss/impl/HNSW.cpp
mdouze: Let's keep it as is. Please put `if(!is_paorama)` first

mdouze:
Please share any performance comparison you have with this code vs. the HNSWFlat implementation. 
Since the data is not contiguous, the performance profile could be different from an IVFFlat index.

AlSchlo:
@mdouze thanks for the review

1) Yes we can use it in `IndexRefine`, this is a good idea. I would assume that `IndexRefine` does not have its vector sequential in memory by design? If no, this is OK but sub-optimal, as the gains of Panorama are more modest in the presence of all those cache misses. We cover this in the [paper](https://arxiv.org/pdf/2510.00566).

2) Performance of HNSW is benched in the paper too, it's still worth it on higher dimensional data, but much more ad-hoc than `IndexIVFFlatPanorama`. We will include some benches with this new cleaned up code. Here is the graph from the paper.
image

3) Panorama can work on IVFPQ (including FastScan), but integration here is a bigger effort (to support all AVX targets, etc.) as we have to interleave the codes to keep the SIMD lanes busy. In fact, this is where we have the best performance speedups.

alexanderguzhva:
@AlSchlo is it worth allowing configuring a default `(UB + LB) / 2` behavior by allowing, say, other options like just `LB`?

AlSchlo:
@alexanderguzhva excellent suggestion! so we actually [used to have an epsilon knob there](https://github.com/fasttrack-nn/panorama/blob/master/faiss/faiss/impl/HNSW.cpp#L920), but we ended up not talking about it in the paper. It's a knob that just adds confusion IMO and makes the workload more unpredictable. 

We did not study it in more detail as the paper was getting too dense.

faiss/IndexHNSW.cpp
AlSchlo: I don't quite remember why I did this. It is probably a relic of an older attempted implementation. Removed it.

faiss/IndexHNSW.cpp
AlSchlo: image
From Cauchy-Schwarz, we precompute the `sqrt` so it does not happen in the hot path. A multiplication is always going to be cheaper.

faiss/IndexHNSW.cpp
AlSchlo: It does not make sense indeed. I did not notice its virtual function was not implemented.

faiss/IndexHNSW.h
AlSchlo: Good idea, perhaps `num_panorama_levels`?

faiss/IndexHNSW.h
AlSchlo: Precisely.

AlSchlo:
Will write tests sometime this week. I also realize I need to change the `write` / `read` functions.
This slipped through the cracks for `IVFFlatPanorama` somehow.

AlSchlo:
Pending benches, I believe this implementation to be pretty much complete.
Next line of work will be (1) fixing the nits in `IVFFlat`, there are currently 2 PRs open from @aknayar and (2) implement `IndexRefinePanorama` as @mdouze suggested.

AlSchlo:
@mdouze The PR is done — could you please re-review?  

Let's also try to get [#4628](https://github.com/facebookresearch/faiss/pull/4628) merged. It's a very useful metric to have, as there's a strong correlation between these stats and empirical performance.  

Here's a nice excerpt from the paper that summarizes this:

image

Thanks!

AlSchlo:
Once this is done, I will focus on getting the `IndexRefinePanorama` PR in.

mnorris11:
Hi @AlSchlo and @aknayar , much thanks for the contributions! The stats PR has been merged. 

After discussing internally, it sounds like the priority is having the `IndexRefinePanorama`. For our (my) learning, after this is in place, is there still a need for various other indexes like IndexHNSWFlatPanorama? Or can it then be applied to all indexes?

AlSchlo:
Hi @mnorris11,

`IndexRefinePanorama` would be a great fit when Panorama cannot be directly applied in the initial search space — for instance, when the dimensionality is small, or when integration into the main index has not yet been done. As a general rule, Panorama can be integrated almost anywhere with some engineering effort.

In our paper, we integrate Panorama into `IVFPQ`, and it performs well even for low-dimensional data, thanks to a SIMD optimization technique called [byte-slicing](https://www.cs.cmu.edu/~15721-f24/papers/ByteSlice.pdf), which helps keep vector lanes fully utilized. However, that implementation requires custom SIMD kernels, which makes it less portable.

This PR instead targets the common scenario where the dimensionality is large and no downstream index is used to refine results after HNSW. That setup is quite typical in practice — for instance, it would benefit us internally at Databricks.

So, my take would be to include both implementations: they address different use cases. Longer-term, the goal is to adapt Panorama into quantization-based techniques like `RaBitQ`, so we can accelerate both the initial search and refinement phases. This is our current research direction.

AlSchlo:
TL;DR @mnorris11 

`IndexRefinePanorama` integrates Panorama into the refinement phase (which might be needed if the upstream index yields poor recall).

`IndexHNSWPanorama` integrates Panorama into the search phase of HNSW.

alexanderguzhva:
@AlSchlo the problem with rabitq that I see is that the overhead of storing additional coefficient is going to be significant, unlike 32-bit or even 16-bit floats for the refinement

AlSchlo:
@alexanderguzhva Yes, this is one issue that will need clever engineering. I was thinking of perhaps quantizing those coefficients. Also rabitq theory assumes random projection. We need to adapt the theory to be able to make it work with Cayley & PCA.

From a computational point of view however, even if the vector is binary quantized, Panorama can still be applied.

mnorris11:
Sorry for the delay on these PRs, I'm still conducting some benchmarking.

aknayar:
@mnorris11 No worries, and thank you so much for the reviews! As an update, after #4645 is confirmed, I have a local build of [`IndexRefinePanorama`](https://github.com/aknayar/faiss-pano/compare/l2flat-pano...aknayar:faiss-pano:refine-pano) ready to submit with really promising results (2x E2E speedups on GIST with `IVF256,PQ60x4fs` as the base index vs. L2Flat as the refine index—seen below). I think speedups of 3x and above could be expected from more amenable datasets (OpenAI's DBpedia-Large, etc.).

image

mnorris11:
The benchmarks look good on my end!

Regarding the `get_query()` on the FlatCodesDistanceComputer: should we just add a `const float* q` at the FlatCodesDistanceComputer level and remove the `const float* query` and `const float* q` from derived structs? I know not all of them use it like RaBitDistanceComputer, but it can just remain nullptr for that struct. And then can we remove the `get_query()`?

faiss/IndexHNSW.cpp
```diff
@@ -649,6 +649,95 @@ IndexHNSWFlat::IndexHNSWFlat(int d, int M, MetricType metric)
     is_trained = true;
 }
 
+/**************************************************************
+ * IndexHNSWFlatPanorama implementation
+ **************************************************************/
+
+void IndexHNSWFlatPanorama::compute_cum_sums(
```
mnorris11, line 656: Edit: Never mind, it is called in both the Index and in HNSW.cpp. Should be fine here.

AlSchlo:
Hi @mnorris11 ,

To clarify: are you suggesting to make `get_query()` virtual?
If so, yes this is possible. I hold no strong opinions on this matter.

mnorris11:
> Hi @mnorris11 ,
> 
> To clarify: are you suggesting to make `get_query()` virtual? If so, yes this is possible. I hold no strong opinions on this matter.

@AlSchlo I mean, is it possible to just have a `q` member on the parent struct rather than having this new `get_query()`? Then when we need to access, we can do `const float* query = flat_codes_qdis->q;` in HNSW.cpp?

How about something like this patch file below? Feel free to note any cons or pitfalls that I'm missing.

cc @mdouze  if you have other opinions. IMO it is only a small style thing. After we decide one way or another, this one should be good to merge?


```
From 975e0a2868f5f2aa30f0aa62e228876ab3f37bfd Mon Sep 17 00:00:00 2001
From: Michael Norris <mnorris@meta.com>
Date: Fri, 14 Nov 2025 12:27:54 -0800
Subject: [PATCH] test to show not having get_query (#4677)

Summary: Pull Request resolved: https://github.com/facebookresearch/faiss/pull/4677

Differential Revision: D87097189


---

diff --git a/faiss/IndexAdditiveQuantizer.cpp b/faiss/IndexAdditiveQuantizer.cpp
--- a/faiss/IndexAdditiveQuantizer.cpp
+++ b/faiss/IndexAdditiveQuantizer.cpp
@@ -58,10 +58,6 @@
         q = x;
     }
 
-    const float* get_query() const override {
-        return q;
-    }
-
     float symmetric_dis(idx_t i, idx_t j) final {
         aq.decode(codes + i * d, tmp.data(), 1);
         aq.decode(codes + j * d, tmp.data() + d, 1);
@@ -81,14 +77,12 @@
     std::vector<float> LUT;
     const AdditiveQuantizer& aq;
     size_t d;
-    const float* q;
 
     explicit AQDistanceComputerLUT(const IndexAdditiveQuantizer& iaq)
             : FlatCodesDistanceComputer(iaq.codes.data(), iaq.code_size),
               LUT(iaq.aq->total_codebook_size + iaq.d * 2),
               aq(*iaq.aq),
-              d(iaq.d),
-              q(nullptr) {}
+              d(iaq.d) {}
 
     float bias;
     void set_query(const float* x) final {
@@ -102,10 +96,6 @@
         }
     }
 
-    const float* get_query() const override {
-        return q;
-    }
-
     float symmetric_dis(idx_t i, idx_t j) final {
         float* tmp = LUT.data();
         aq.decode(codes + i * d, tmp, 1);
diff --git a/faiss/IndexFlat.cpp b/faiss/IndexFlat.cpp
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -100,7 +100,6 @@
 struct FlatL2Dis : FlatCodesDistanceComputer {
     size_t d;
     idx_t nb;
-    const float* q;
     const float* b;
     size_t ndis;
     size_t npartial_dot_products;
@@ -126,10 +125,10 @@
     explicit FlatL2Dis(const IndexFlat& storage, const float* q = nullptr)
             : FlatCodesDistanceComputer(
                       storage.codes.data(),
-                      storage.code_size),
+                      storage.code_size,
+                      q),
               d(storage.d),
               nb(storage.ntotal),
-              q(q),
               b(storage.get_xb()),
               ndis(0),
               npartial_dot_products(0) {}
@@ -138,10 +137,6 @@
         q = x;
     }
 
-    const float* get_query() const override {
-        return q;
-    }
-
     // compute four distances
     void distances_batch_4(
             const idx_t idx0,
@@ -250,10 +245,6 @@
         q = x;
     }
 
-    const float* get_query() const override {
-        return q;
-    }
-
     // compute four distances
     void distances_batch_4(
             const idx_t idx0,
@@ -378,10 +369,6 @@
         query_l2norm = fvec_norm_L2sqr(q, d);
     }
 
-    const float* get_query() const override {
-        return q;
-    }
-
     // compute four distances
     void distances_batch_4(
             const idx_t idx0,
diff --git a/faiss/IndexFlatCodes.cpp b/faiss/IndexFlatCodes.cpp
--- a/faiss/IndexFlatCodes.cpp
+++ b/faiss/IndexFlatCodes.cpp
@@ -142,10 +142,6 @@
         query = x;
     }
 
-    const float* get_query() const override {
-        return query;
-    }
-
     float operator()(idx_t i) override {
         codec.sa_decode(1, codes + i * code_size, vec_buffer.data());
         return vd(query, vec_buffer.data());
diff --git a/faiss/IndexPQ.cpp b/faiss/IndexPQ.cpp
--- a/faiss/IndexPQ.cpp
+++ b/faiss/IndexPQ.cpp
@@ -132,10 +132,6 @@
             pq.compute_inner_prod_table(x, precomputed_table.data());
         }
     }
-
-    const float* get_query() const override {
-        return q;
-    }
 };
 
 } // namespace
diff --git a/faiss/impl/DistanceComputer.h b/faiss/impl/DistanceComputer.h
--- a/faiss/impl/DistanceComputer.h
+++ b/faiss/impl/DistanceComputer.h
@@ -113,13 +113,16 @@
     const uint8_t* codes;
     size_t code_size;
 
-    /// Returns a pointer to the currently active query vector.
-    /// Only present for FlatCodesDistanceComputer subclasses
-    /// out of need for Panorama to compute the squared norm of the query.
-    virtual const float* get_query() const = 0;
+    const float* q = nullptr; // not used in all distance computers
 
-    FlatCodesDistanceComputer(const uint8_t* codes, size_t code_size)
-            : codes(codes), code_size(code_size) {}
+    FlatCodesDistanceComputer(
+            const uint8_t* codes,
+            size_t code_size,
+            const float* q = nullptr)
+            : codes(codes), code_size(code_size), q(q) {}
+
+    FlatCodesDistanceComputer(const float* q)
+            : codes(nullptr), code_size(0), q(q) {}
 
     FlatCodesDistanceComputer() : codes(nullptr), code_size(0) {}
 
diff --git a/faiss/impl/HNSW.cpp b/faiss/impl/HNSW.cpp
--- a/faiss/impl/HNSW.cpp
+++ b/faiss/impl/HNSW.cpp
@@ -801,7 +801,7 @@
     std::vector<idx_t> index_array(M);
     std::vector<float> exact_distances(M);
 
-    const float* query = flat_codes_qdis->get_query();
+    const float* query = flat_codes_qdis->q;
     std::vector<float> query_cum_sums(panorama_index->num_panorama_levels + 1);
     IndexHNSWFlatPanorama::compute_cum_sums(
             query,
diff --git a/faiss/impl/RaBitQuantizer.cpp b/faiss/impl/RaBitQuantizer.cpp
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -133,14 +133,8 @@
     // the metric
     MetricType metric_type = MetricType::METRIC_L2;
 
-    const float* q = nullptr;
-
     RaBitDistanceComputer();
 
-    const float* get_query() const override {
-        return q;
-    }
-
     float symmetric_dis(idx_t i, idx_t j) override;
 };
 
diff --git a/faiss/impl/ScalarQuantizer.h b/faiss/impl/ScalarQuantizer.h
--- a/faiss/impl/ScalarQuantizer.h
+++ b/faiss/impl/ScalarQuantizer.h
@@ -98,13 +98,7 @@
     SQuantizer* select_quantizer() const;
 
     struct SQDistanceComputer : FlatCodesDistanceComputer {
-        const float* q;
-
-        SQDistanceComputer() : q(nullptr) {}
-
-        const float* get_query() const override {
-            return q;
-        }
+        SQDistanceComputer() : FlatCodesDistanceComputer(nullptr) {}
 
         virtual float query_to_code(const uint8_t* code) const = 0;
 
diff --git a/faiss/utils/extra_distances.cpp b/faiss/utils/extra_distances.cpp
--- a/faiss/utils/extra_distances.cpp
+++ b/faiss/utils/extra_distances.cpp
@@ -126,10 +126,6 @@
     void set_query(const float* x) override {
         q = x;
     }
-
-    const float* get_query() const override {
-        return q;
-    }
 };
 
 struct Run_get_distance_computer {

--

```

AlSchlo:
@mnorris11  That's a good suggestion, makes it a bit cleaner and less invasive.

faiss/impl/HNSW.cpp
```diff
@@ -735,6 +751,254 @@ int search_from_candidates(
...
+            batch_size = next_batch_size;
+            curr_panorama_level++;
+        }
+
+        // Add surviving candidates to the result handler.
+        for (size_t i = 0; i < batch_size; i++) {
+            idx_t idx = index_array[i];
+            if (!sel || sel->is_member(idx)) {
+                if (res.add_result(exact_distances[i], idx)) {
+                    threshold = res.threshold;
```
mnorris11: It seems like `threshold` is not used after being written to here?

AlSchlo: Indeed, I did this to better mirror the original code. Please check my latest commit, where I made it a bit more idiomatic.

AlSchlo: (just realized this has a little bug which I will fix, condition seems wrong)

AlSchlo: Fixed and wrote a test for it! Please check out :) @mnorris11

mnorris11:
Sorry to mention it after so long: I realized the test is cpp, not Python. Usually we prefer Python tests if possible, to exercise the SWIG layer. Can this be converted?

Otherwise, tests are all passing internally and benchmarks look good.

I will add Panorama for various indexes to the "Guidesline to choose an index" and "Index Factory" wikis after everything is merged.

AlSchlo:
Ah, I matched `test_hnsw.cpp`, let me see what I can do!

AlSchlo:
Done. I translated it with Claude and then reviewed the output manually. It seems quite accurate to me. @mnorris11

PR #4623, opened 2025-10-20
Allow unaligned fast-scan

mnorris11:
Summary:
Codes can be loaded from disk, so alignment is not guaranteed. This diff allows unaligned pointer in pq fast scan code.

Tried microbenchmarking in a loop. Seems to be the same between unaligned and aligned.

Unit Test
-
For non avx512, the code handles it correctly with the unit test. P1998522307 shows the error without any `c.loadu(codes)` handling.

Do we need to update avx512 version?
-
- The unit test passes with whatever code size I tried. I might have to change other things like nsq? May need input here.

Differential Revision: D84364872

### DIFF
--- a/faiss/impl/pq4_fast_scan_search_qbs.cpp
+++ b/faiss/impl/pq4_fast_scan_search_qbs.cpp
@@ -53,8 +53,8 @@ void kernel_accumulate_block(
 
     // _mm_prefetch(codes + 768, 0);
     for (int sq = 0; sq < nsq - scaler.nscale; sq += 2) {
-        // prefetch
-        simd32uint8 c(codes);
+        simd32uint8 c;
+        c.loadu(codes);
         codes += 32;
 
         simd32uint8 mask(0xf);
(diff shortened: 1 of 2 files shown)


alexanderguzhva:
@mnorris11 that was a strong phrase about C++ templates lol

alexanderguzhva:
@mnorris11 technically speaking, I don't believe that unaligned loads have different for aligned loads for relatively modern CPUs. So, you should also try literally replacing aligned loads in avx SIMDlib code with unaligned (say, `_mm512_load_si512()` -> `_mm512_loadu_si512()`) and see whether the benchmarking numbers drop at all. If so, then `loadu` must be made a default operation

mnorris11:
> @mnorris11 that was a strong phrase about C++ templates lol

ah, since we have removed the template argument, I can just remove this

mnorris11:
> @mnorris11 technically speaking, I don't believe that unaligned loads have different for aligned loads for relatively modern CPUs. So, you should also try literally replacing aligned loads in avx SIMDlib code with unaligned (say, `_mm512_load_si512()` -> `_mm512_loadu_si512()`) and see whether the benchmarking numbers drop at all. If so, then `loadu` must be made a default operation

Thanks for this input! 

Before creating the PR, I tried comparing them with `std::chrono::high_resolution_clock` in a loop over some thousands of trials. The average speed for both unaligned and aligned is about 8us.

PR #4628, opened 2025-10-21
Implement PanoramaStats

aknayar:
This PR builds upon #4606, adding `PanoramaStats` to help gauge the effectiveness of Panorama's pruning.

## Evaluating Panorama
The squared distance between a query $q$ and a point $p$ is given by:
```math
\begin{aligned}
\Vert\  q\Vert_2^2 + \Vert p\Vert_2^2 - 2(q\cdot p)\\
\end{aligned}
```
Panorama's speedups come from minimizing the average fraction of dimensions processed among all $q\cdot p$ dot products (known as $\phi$, as defined in Theorem 1 of the [paper](https://arxiv.org/abs/2510.00566)). This PR exposes the empirically observed value of $\phi$ as `ratio_dims_scanned`, a member of the new `PanoramaStats` struct. 

`ratio_dims_scanned` is a lightweight and effective way to accurately gauge Panorama's benefits—lower values mean that Panorama is able to process less features and is achieving higher speedups, while higher values mean that Panorama is forced to spend more time calculating distances. In an ideal scenario (and when `nprobe` = `nlist` = 1), the observed speedup due to Panorama would be around $1/\phi$.

## Implementation
The newly created `PanoramaStats` struct exposes three values:
1. `total_dims_scanned` - The total number of dimensions across all candidates points processed as part of distance calculations. This doesn't include dimensions that have been pruned.
2. `total_dims` - The total number of dimensions among all candidate points, including pruned dimensions.
3. `ratio_dims_scanned` - The empirical value of $\phi$, calculated as `total_dims_scanned`/`total_dims`.

Since the existing `IndexIVFFlatPanorama` and all future Panorama indexes added to Faiss will need just these three values to estimate $\phi$, this struct is defined once as `PanoramaStats` (as opposed to having separate `IndexIVFFlatPanoramaStats`, `IndexHNSWFlatPanoramaStats`, etc., for each existing and future Panorama index).

Panorama indexes can populate `indexPanorama_stats`, a globally-defined instance of `PanoramaStats`. Python scripts can interact with these stats as follows:
```python
faiss.cvar.indexPanorama_stats.reset()
panorama_index.search(xq, k)
print(faiss.cvar.indexPanorama_stats.ratio_dims_scanned)
```

## Thread Safety
Since `indexPanorama_stats` is defined at the global scope, it is not robust against:
1. Concurrent calls to `search()` by any two Panorama indexes.
2. Concurrency within a single `search()` call.

As such, `indexPanorama_stats` should be used in single-threaded contexts to accurately and reliably measure the effectiveness of Panorama pruning.

Making `indexPanorama_stats` robust against (2.) would likely require overriding `search()` (and using thread-specific `PanoramaStats` objects, similar to what [IVFFlat does](https://github.com/facebookresearch/faiss/blob/70df32b2e989e0c47c04cfd78be3221c79c8d408/faiss/IndexIVF.cpp#L357-L373)). However, I'm not sure overriding this function for the sole purpose of making `indexPanorama_stats` thread-safe makes sense. I'm open to any suggestions folks may have.

## Testing
Added tests to `test_panorama.py` to verify correct tracking of `total_dims_scanned` and `total_dims`. Also updated `benchs/bench_ivf_flat_panorama.py` to record `ratio_dims_scanned` (in the form of `dims scanned`) for each test run:
```
======IVF128,Flat
        nprobe   1, Recall@10: 0.145200, speed: 2.681885 ms/query, dims scanned: 100.00%
        nprobe   2, Recall@10: 0.261300, speed: 5.421119 ms/query, dims scanned: 100.00%
        nprobe   4, Recall@10: 0.440800, speed: 10.807769 ms/query, dims scanned: 100.00%
        nprobe   8, Recall@10: 0.647900, speed: 21.465333 ms/query, dims scanned: 100.00%
        nprobe  16, Recall@10: 0.879100, speed: 42.669564 ms/query, dims scanned: 100.00%
        nprobe  32, Recall@10: 0.975400, speed: 83.598829 ms/query, dims scanned: 100.00%
        nprobe  64, Recall@10: 0.986800, speed: 158.295513 ms/query, dims scanned: 100.00%
======PCA960,IVF128,FlatPanorama8
        nprobe   1, Recall@10: 0.144400, speed: 0.607333 ms/query, dims scanned: 17.03%
        nprobe   2, Recall@10: 0.260900, speed: 1.059646 ms/query, dims scanned: 15.52%
        nprobe   4, Recall@10: 0.439300, speed: 1.890787 ms/query, dims scanned: 14.52%
        nprobe   8, Recall@10: 0.647600, speed: 3.449271 ms/query, dims scanned: 13.81%
        nprobe  16, Recall@10: 0.878200, speed: 6.400299 ms/query, dims scanned: 13.30%
        nprobe  32, Recall@10: 0.976500, speed: 11.893498 ms/query, dims scanned: 12.96%
        nprobe  64, Recall@10: 0.986800, speed: 21.931417 ms/query, dims scanned: 12.75%
```

### DIFF
--- a/.github/workflows/autoclose.yml
+++ b/.github/workflows/autoclose.yml
@@ -15,6 +15,7 @@ jobs:
       - name: Comment on autoclose label
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
         run: gh issue comment ${{ github.event.issue.number }} --body "This issue has been marked for autoclose and will be closed after 14 days of inactivity."
 
   close-issues:
--- a/c_api/VectorTransform_c.cpp
+++ b/c_api/VectorTransform_c.cpp
@@ -120,6 +120,10 @@ DEFINE_GETTER(PCAMatrix, float, eigen_power)
 
 DEFINE_GETTER(PCAMatrix, int, random_rotation)
 
+DEFINE_GETTER(PCAMatrix, int, balanced_bins)
+
+DEFINE_SETTER(PCAMatrix, int, balanced_bins)
+
 /*********************************************
  * ITQMatrix
  *********************************************/
--- a/c_api/VectorTransform_c.h
+++ b/c_api/VectorTransform_c.h
@@ -116,6 +116,12 @@ FAISS_DECLARE_GETTER(PCAMatrix, float, eigen_power)
 /// Getter for random_rotation
 FAISS_DECLARE_GETTER(PCAMatrix, int, random_rotation)
 
+/// Getter for balanced_bins
+FAISS_DECLARE_GETTER(PCAMatrix, int, balanced_bins)
+
+/// Setter for balanced_bins
+FAISS_DECLARE_SETTER(PCAMatrix, int, balanced_bins)
+
 FAISS_DECLARE_CLASS_INHERITED(ITQMatrix, VectorTransform)
 FAISS_DECLARE_DESTRUCTOR(ITQMatrix)
 
--- a/faiss/IndexIVFFlatPanorama.cpp
+++ b/faiss/IndexIVFFlatPanorama.cpp
@@ -44,6 +44,8 @@ IndexIVFFlatPanorama::IndexIVFFlatPanorama(
     this->own_invlists = own_invlists;
 }
 
+IndexIVFFlatPanorama::IndexIVFFlatPanorama() : n_levels(0) {}
+
 namespace {
 
 template <typename VectorDistance, bool use_sel>
--- a/faiss/IndexIVFFlatPanorama.h
+++ b/faiss/IndexIVFFlatPanorama.h
@@ -36,6 +36,10 @@ namespace faiss {
 /// insertion logic. The code responsible for level-oriented storage is in
 /// `ArrayInvertedListsPanorama`, which is a struct member of `IndexIVF`.
 struct IndexIVFFlatPanorama : IndexIVFFlat {
+    size_t n_levels;
+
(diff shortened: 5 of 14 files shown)


aknayar:
@alexanderguzhva Apologies for the duplicate PR, I accidentally force-pushed to the original branch while the PR was closed, so now I can't reopen it.

faiss/impl/PanoramaStats.cpp
```diff
@@ -0,0 +1,35 @@
...
+// -*- c++ -*-
+
+#include <faiss/impl/PanoramaStats.h>
+
+#include <cstring>
+
+namespace faiss {
+
+void PanoramaStats::reset() {
+    total_dims_scanned = 0;
```
aknayar, line 17: @alexanderguzhva I agree with your comment on my previous PR that `memset((void*)this, 0, sizeof(*this));` isn't the cleanest way to implement this function. Out of curiosity, why do [`IndexIVFStats`](https://github.com/facebookresearch/faiss/blob/e6510bd00478d563e380604374f41a27e8697ce6/faiss/IndexIVF.cpp#L1282-L1284) and [`IndexBinaryHashStats`](https://github.com/facebookresearch/faiss/blob/e5fee1015a90dd3667c0d1f7d866bbc2c023c475/faiss/IndexBinaryHash.cpp#L278-L280) implement `reset()` like that, is there a particular benefit in those cases?

alexanderguzhva: @aknayar I think that this is because of some weird historical reasons, and it looks like a definite mistake if so

PR #4636, opened 2025-10-24
Implement serialization for `IndexIVFFlatPanorama`

aknayar:
## Changes
This PR implements `write_index` and `read_index` for `IndexIVFFlatPanorama`.

## Testing
Added a test to `test_panorama` to verify integrity of `search()` results after writing and reading back the index from disk.

### DIFF
--- a/faiss/IndexIVFFlatPanorama.cpp
+++ b/faiss/IndexIVFFlatPanorama.cpp
@@ -43,7 +43,7 @@ IndexIVFFlatPanorama::IndexIVFFlatPanorama(
     this->own_invlists = own_invlists;
 }
 
-IndexIVFFlatPanorama::IndexIVFFlatPanorama() : n_levels(0) { }
+IndexIVFFlatPanorama::IndexIVFFlatPanorama() : n_levels(0) {}
 
 namespace {
 
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -340,7 +340,8 @@ InvertedLists* read_InvertedLists(IOReader* f, int io_flags) {
             ailp->ids[i].resize(sizes[i]);
             size_t num_elems =
                     ((sizes[i] + ArrayInvertedListsPanorama::kBatchSize - 1) /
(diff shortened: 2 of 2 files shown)


faiss/impl/index_write.cpp
```diff
@@ -247,6 +248,32 @@ void write_InvertedLists(const InvertedLists* ils, IOWriter* f) {
...
         WRITE1(h);
+    } else if (
+            const auto& ailp =
+                    dynamic_cast<const ArrayInvertedListsPanorama*>(ils)) {
+        uint32_t h = fourcc("ilpn");
+        WRITE1(h);
+        WRITE1(ailp->nlist);
+        WRITE1(ailp->code_size);
+        WRITE1(ailp->n_levels);
+        uint32_t list_type = fourcc("full");
```
aknayar, line 259: I notice that the implementation for [`ArrayInvertedLists`](https://github.com/facebookresearch/faiss/blob/a93241751271aaf6028288b0a12a9819dde8a4aa/faiss/impl/index_write.cpp#L283-L310) potentially writes a sparse buffer. However, I'm pretty sure that case is incredibly unlikely, and thus didn't implement it here. That being said, if others think it's worthwhile to implement that case, I will.

tests/test_panorama.py
```diff
@@ -508,3 +510,31 @@ def test_update_vectors(self):
...
         D_pan_2, I_pan_2 = index_panorama.search(xq, k)
         self.assert_search_results_equal(D_reg_2, I_reg_2, D_pan_2, I_pan_2)
+
+    def test_serialization(self):
+        """Test that writing and reading Panorama indexes preserves search results"""
+        d, nb, nt, nq, nlist, nlevels, k = 128, 10000, 15000, 100, 128, 8, 20
+        xt, xb, xq = self.generate_data(d, nt, nb, nq, seed=2024)
+
+        # Create and populate Panorama indexes
+        indexes = [self.create_panorama(d, nlist, nlevels, xt, xb, nprobe=32)]
```
aknayar, line 520: Maintained as a list to prepare for other incoming Panorama indexes (e.g. `IndexHNSWFlatPanorama` #4621).

PR #4645, opened 2025-10-30
Implement `IndexFlatL2Panorama`

aknayar:
This PR adds `IndexFlatL2Panorama`, integrating Panorama (as specified in the [paper](https://arxiv.org/abs/2510.00566)) into `IndexFlatL2`. This is the first step in creating an [`IndexRefinePanorama`](https://github.com/aknayar/faiss-pano/compare/aknayar:faiss-pano:l2flat-pano...aknayar:faiss-pano:refine-pano), which will use `IndexFlatL2Panorama` (or an `IndexPreTransform` with an `IndexFlatL2Panorama`) as its `refine_index`.

## Refactoring
Since the bulk of Panorama's refinement logic would be duplicated between `IndexFlatL2Panorama` and `IndexIVFFlatPanorama`, it has been factored out into a new `Panorama` struct. This struct contains key parameters (`batch_size`, `d`, etc.) and the following utility functions:
 - `copy_codes_to_level_layout`: Writes new vectors to `codes` following Panorama's storage layout
 - `compute_cumulative_sums`: Computes the cumulative sums for new vectors
 - `compute_query_cum_sums`: Computes the cumulative sums for a new query
 - `progressive_filter_batch`: Performs Panorama refinement on a batch of vectors

These utilities will be shared by most Panorama indexes, which is why I have refactored them into their own utility.

## IndexRefinePanorama
While the `IndexFlatL2Panorama` implemented in this PR technically contains all the functionality needed to implement `IndexRefinePanorama` (performing `search` on a subset of indices), it is not ready to be used as a `refine_index`. The current implementation is not optimized for the case of `IndexRefine`, where we perform search on a very small subset of the datapoints. This leads to vastly scattered memory accesses during the `search`, to the point where the overhead of maintaining `active_indices` and `exact_distances` can thwart Panorama's speedups.

As such, to optimize for `IndexRefine` we will need a standalone implementation of `search_subset` which instead does the following:
1. Iterate over the subset of indices (rather than batches of codes)
2. For each index `i`, compute its distance alone by Panorama refinement (essentially having `batch_size` = 1. In fact, for this very reason I have made `batch_size` a parameter in the constructor—`IndexRefine` will require it to be 1 due to noncontiguous memory accesses, but typical workloads would benefit from 128-1024.) 

This will unfortunately mean we cannot reuse the search utilities in the `Panorama` struct in this specific case, but will allow us to squeeze 2-5x speedups during the reordering phase of `IndexRefine`.

## Testing
1. Unit tests can be found in `tests/test_flat_l2_panorama.py`
2. A benchmark can be found in `benchs/bench_flat_l2_panorama.py`, yielding the following results:
```
======Flat
        Recall@10: 0.980000, speed: 263.214254 ms/query, dims scanned: 100.00%
======PCA960,FlatL2Panorama8_512
        Recall@10: 0.980000, speed: 37.080264 ms/query, dims scanned: 12.62%
```
The recall being less than 1.0 is perhaps due to discrepancies between faiss results and the `ground_truth` values.
bench_flat_l2_panorama

### DIFF
--- a/faiss/IVFlib.cpp
+++ b/faiss/IVFlib.cpp
@@ -558,19 +558,6 @@ void handle_ivf(
         const std::string& filename_template,
         ShardingFunction* sharding_function,
         bool generate_ids) {
-    std::vector<faiss::IndexIVF*> sharded_indexes(shard_count);
-    auto clone = static_cast<faiss::IndexIVF*>(faiss::clone_index(index));
-    clone->quantizer->reset();
-    for (int64_t i = 0; i < shard_count; i++) {
-        sharded_indexes[i] =
-                static_cast<faiss::IndexIVF*>(faiss::clone_index(clone));
-        if (generate_ids) {
-            // Assume the quantizer does not natively support add_with_ids.
-            sharded_indexes[i]->quantizer =
-                    new IndexIDMap2(sharded_indexes[i]->quantizer);
-        }
-    }
-
     // assign centroids to each sharded Index based on sharding_function, and
     // add them to the quantizer of each sharded index
     std::vector<std::vector<float>> sharded_centroids(shard_count);
@@ -588,27 +575,29 @@ void handle_ivf(
                 &reconstructed[index->quantizer->d]);
         delete[] reconstructed;
     }
+
+    auto clone = static_cast<faiss::IndexIVF*>(faiss::clone_index(index));
+    clone->quantizer->reset();
     for (int64_t i = 0; i < shard_count; i++) {
+        auto sharded_index =
+                static_cast<faiss::IndexIVF*>(faiss::clone_index(clone));
         if (generate_ids) {
-            sharded_indexes[i]->quantizer->add_with_ids(
+            // Assume the quantizer does not natively support add_with_ids.
+            sharded_index->quantizer =
+                    new IndexIDMap2(sharded_index->quantizer);
+            sharded_index->quantizer->add_with_ids(
                     sharded_centroids[i].size() / index->quantizer->d,
                     sharded_centroids[i].data(),
                     xids[i].data());
         } else {
-            sharded_indexes[i]->quantizer->add(
+            sharded_index->quantizer->add(
                     sharded_centroids[i].size() / index->quantizer->d,
                     sharded_centroids[i].data());
         }
-    }
-
-    for (int64_t i = 0; i < shard_count; i++) {
         char fname[256];
         snprintf(fname, 256, filename_template.c_str(), i);
-        faiss::write_index(sharded_indexes[i], fname);
-    }
-
-    for (int64_t i = 0; i < shard_count; i++) {
-        delete sharded_indexes[i];
+        faiss::write_index(sharded_index, fname);
+        delete sharded_index;
     }
 }
 
@@ -618,22 +607,6 @@ void handle_binary_ivf(
         const std::string& filename_template,
         ShardingFunction* sharding_function,
         bool generate_ids) {
-    std::vector<faiss::IndexBinaryIVF*> sharded_indexes(shard_count);
-
-    auto clone = static_cast<faiss::IndexBinaryIVF*>(
-            faiss::clone_binary_index(index));
-    clone->quantizer->reset();
-
-    for (int64_t i = 0; i < shard_count; i++) {
-        sharded_indexes[i] = static_cast<faiss::IndexBinaryIVF*>(
-                faiss::clone_binary_index(clone));
-        if (generate_ids) {
-            // Assume the quantizer does not natively support add_with_ids.
-            sharded_indexes[i]->quantizer =
-                    new IndexBinaryIDMap2(sharded_indexes[i]->quantizer);
-        }
-    }
-
     // assign centroids to each sharded Index based on sharding_function, and
     // add them to the quantizer of each sharded index
     int64_t reconstruction_size = index->quantizer->d / 8;
@@ -652,27 +625,30 @@ void handle_binary_ivf(
                 &reconstructed[reconstruction_size]);
         delete[] reconstructed;
     }
+
+    auto clone = static_cast<faiss::IndexBinaryIVF*>(
+            faiss::clone_binary_index(index));
+    clone->quantizer->reset();
     for (int64_t i = 0; i < shard_count; i++) {
+        auto sharded_index = static_cast<faiss::IndexBinaryIVF*>(
+                faiss::clone_binary_index(clone));
         if (generate_ids) {
-            sharded_indexes[i]->quantizer->add_with_ids(
+            // Assume the quantizer does not natively support add_with_ids.
+            sharded_index->quantizer =
+                    new IndexBinaryIDMap2(sharded_index->quantizer);
+            sharded_index->quantizer->add_with_ids(
                     sharded_centroids[i].size() / reconstruction_size,
                     sharded_centroids[i].data(),
                     xids[i].data());
         } else {
-            sharded_indexes[i]->quantizer->add(
+            sharded_index->quantizer->add(
                     sharded_centroids[i].size() / reconstruction_size,
                     sharded_centroids[i].data());
         }
-    }
-
-    for (int64_t i = 0; i < shard_count; i++) {
         char fname[256];
         snprintf(fname, 256, filename_template.c_str(), i);
-        faiss::write_index_binary(sharded_indexes[i], fname);
-    }
-
-    for (int64_t i = 0; i < shard_count; i++) {
-        delete sharded_indexes[i];
+        faiss::write_index_binary(sharded_index, fname);
+        delete sharded_index;
     }
 }
 
--- a/faiss/utils/distances.cpp
+++ b/faiss/utils/distances.cpp
@@ -321,7 +321,6 @@ void exhaustive_L2sqr_blas_default_impl(
                        ip_block.get(),
                        &nyi);
             }
-#pragma omp parallel for
             for (int64_t i = i0; i < i1; i++) {
                 float* ip_line = ip_block.get() + (i - i0) * (j1 - j0);
 
@@ -423,7 +422,6 @@ void exhaustive_L2sqr_blas_cmax_avx2(
                        ip_block.get(),
                        &nyi);
             }
-#pragma omp parallel for
             for (int64_t i = i0; i < i1; i++) {
                 float* ip_line = ip_block.get() + (i - i0) * (j1 - j0);
 
@@ -633,7 +631,6 @@ void exhaustive_L2sqr_blas_cmax_sve(
                        ip_block.get(),
                        &nyi);
             }
-#pragma omp parallel for
             for (int64_t i = i0; i < i1; i++) {
                 const size_t count = j1 - j0;
                 float* ip_line = ip_block.get() + (i - i0) * count;


tests/test_ivf_flat_panorama.py
```diff
@@ -198,9 +198,11 @@ def test_uneven_dimension_division(self):
         """Test when n_levels doesn't evenly divide dimension"""
         test_cases = [(65, 4), (63, 8), (100, 7)]
 
+        # TODO(aknayar): Test functions like get_single_code().
```
aknayar, line 206: Will add these tests in a follow-up PR.

aknayar:
I rebuilt with AVX2 on Linux and was unable to reproduce the failing tests seen [here](https://github.com/facebookresearch/faiss/actions/runs/18958114742/job/54140357566?pr=4645), any ideas what may have happened?

limqiying:
Hey, let me just copy what @mnorris11  said and we can resume the thread from there.

> When checking the log, it is just 2 elements wrong in one unit test. Some ideas:

> There could be a tie in distance, and different ids happen to get returned. If this is the case, they could update the test where if there is a mismatch, then check distances, and if they are equal then still do not fail the test.

Do you have faiss installed with numpy2? It's a recent integration, and that *could* be the reason for the difference. Let me know the conda steps you took to repro!

aknayar:
@limqiying Thank you for the ideas! It seems like it was, in fact, some weird case involving a tie. Panorama also suffers from floating-point imprecision due to how we calculate squared L2 norm: Faiss does $\Vert\  q - p\  \Vert_2^2$ while Panorama does $\Vert\  q\Vert_2^2 + \Vert p\Vert_2^2 - 2(q\cdot p)$. As a result, such cases where ties produce different results are possible (albeit rare). Simply changing the random seed on the dataset generator for the testcase seemed to fix things!

aknayar:
@mnorris11 I notice that `IndexFlatCodes`, which is an ancestor of `IndexFlatL2Panorama`, implements functions that now must be overriden by `IndexFlatL2Panorama` (`sa_encode`, `sa_decode`, `reconstruct`, `merge_from`, etc.).

Do you think it's worth it to instead define `IndexPanoramaCodes` (analogous to `IndexFlatCodes`, but accounting for Panorama's storage layout) which defines these methods once and have `IndexPanorama` (analogous to `IndexFlat`) derive from that? We could then have `IndexPanoramaL2` (and `IndexPanoramaIP` when we add support inner product) derive from `IndexPanorama`. Let me know your thoughts on this, perhaps I'm overcomplicating things.

faiss/IndexFlat.h
```diff
@@ -99,6 +100,84 @@ struct IndexFlatL2 : IndexFlat {
...
+    explicit IndexFlatPanorama(
+            idx_t d,
+            MetricType metric,
+            size_t n_levels,
+            size_t batch_size)
+            : IndexFlat(d, metric),
+              n_levels(n_levels),
+              batch_size(batch_size),
+              pano(code_size, n_levels, batch_size) {
+        FAISS_THROW_IF_NOT(metric == METRIC_L2);
```
aknayar, line 124: This will be relaxed when we add inner product support.

faiss/IndexFlat.cpp
```diff
@@ -519,4 +522,150 @@ void IndexFlat1D::search(
...
+
+    search_core<true>(handler, n, x, radius, params);
+}
+
+void IndexFlatPanorama::reset() {
+    IndexFlat::reset();
+    cum_sums.clear();
+}
+
+void IndexFlatPanorama::reconstruct(idx_t key, float* recons) const {
```
aknayar, line 658: I can implement these in a follow-up PR since this one is quite big already.

mnorris11:
Hi @aknayar , when doing some benchmarks, I cannot seem to get a speedup from IndexFlatPanorama. It could be due to my configuration for `nlevels` and `batch_size`. How do you choose these in relation to database size, dimensions, topk, number of queries etc? Say nb = 1 million vectors, d=256, k = 10 or 100, nq = 1000. (I chose 8 for nlevels and 512 for batch_size, which could be suboptimal)
```
        index_regular = faiss.index_factory(d, "Flat")
        batch_size = 512
        index = faiss.IndexFlatPanorama(d, faiss.METRIC_L2, nlevels, batch_size)
        index_panorama = faiss.IndexPreTransform(faiss.PCAMatrix(d, d), index)
        # add vectors, then query index_regular and index_panorama etc
```

aknayar:
Hi @mnorris11, this is interesting—your parameters seem fine. 

The _theoretical_ ideal for `batch_size` is 1, as it allows the most pruning, but in practice anywhere from 256-1024 yields the best speedups (too low will incur more overhead while too high will cause less pruning). It also must be larger than `k`, but that is already the case here.

`nlevels` depends mostly on the dimensionality of the dataset. For 256, either 4 or 8 should be best (for 128 dims I'd recommend 4 levels and for 1024 dims and above I'd recommend 8-32 levels). When you increase `nlevels`, you'll prune more, but also incur more overhead.

I tried your setup on our server and observed a 6.98x speedup on GIST1M:
```
akash@lutex:~/faiss-pano$ python3 benchs/bench_flat_l2_panorama.py 
======Flat
        Recall@10: 0.980000, speed: 266.571403 ms/query, dims scanned: 100.00%
======PCA960,FlatL2Panorama8_512
        Recall@10: 0.980000, speed: 38.172317 ms/query, dims scanned: 12.62%
```

Do you mind sharing your benchmarking script so I could give it a try locally? I'm also curious what `faiss.cvar.indexPanorama_stats.ratio_dims_scanned` outputs after the Panorama query. Thanks!

mnorris11:
> Hi @mnorris11, this is interesting—your parameters seem fine.
> 
> The _theoretical_ ideal for `batch_size` is 1, as it allows the most pruning, but in practice anywhere from 256-1024 yields the best speedups (too low will incur more overhead while too high will cause less pruning). It also must be larger than `k`, but that is already the case here.
> 
> `nlevels` depends mostly on the dimensionality of the dataset. For 256, either 4 or 8 should be best (for 128 dims I'd recommend 4 levels and for 1024 dims and above I'd recommend 8-32 levels). When you increase `nlevels`, you'll prune more, but also incur more overhead.
> 
> I tried your setup on our server and observed a 6.98x speedup on GIST1M:
> 
> ```
> akash@lutex:~/faiss-pano$ python3 benchs/bench_flat_l2_panorama.py 
> ======Flat
>         Recall@10: 0.980000, speed: 266.571403 ms/query, dims scanned: 100.00%
> ======PCA960,FlatL2Panorama8_512
>         Recall@10: 0.980000, speed: 38.172317 ms/query, dims scanned: 12.62%
> ```
> 
> Do you mind sharing your benchmarking script so I could give it a try locally? I'm also curious what `faiss.cvar.indexPanorama_stats.ratio_dims_scanned` outputs after the Panorama query. Thanks!

Thanks for the info. I was checking some internal datasets, along with the SyntheticDataset present within Faiss. It looks something like this, piecemeal from the actual script.
```
def generate_data(d, nt, nb, nq, seed=42):
    ds = SyntheticDataset(d, nt, nb, nq)
    return np.concatenate([ds.get_database(), ds.get_queries(), ds.get_train()])

def get_avg_query_speed(index, xq, k):
    trials = 3
    result = timeit.timeit(
        stmt="index.search(xq, k)",
        number=trials,
        globals={"index": index, "xq": xq, "k": k},
    )
    return result / trials * 1000.0  # ms

d, nb, nt, nq, nlevels, k = 2048, 1_000_000, 150_000, 1000, 8, 100
x = generate_data(d, nt, nb, nq, seed=42)
xb = x[:nb]
xq = x[nb : nb + nq]
xt = x[nb + nq : nb + nq + nt]

nlevel_list = [8, 16, 32, 64]
batch_sizes = [128, 256, 512]
for nl in nlevel_list:
    for bs in batch_sizes:
        index_regular = faiss.index_factory(d, "Flat")
        index = faiss.IndexFlatPanorama(d, faiss.METRIC_L2, nl, bs)
        index_panorama = faiss.IndexPreTransform(faiss.PCAMatrix(d, d), index)
        # index_panorama = faiss.index_factory(d, f"PCA{d},FlatL2Panorama{nl}_{bs}") # Saw some errors related to SWIG with this but didn't check deeper. Could be an issue running it on our internal build system due to SWIG version etc.
        index_regular.train(xt)
        index_regular.add(xb)
        index_panorama.train(xt)
        index_panorama.add(xb)
        first_str = f"nlevel={nl}, batch_size={bs}"

        _, I_regular = index_regular.search(xq, k)
        regular_q_speed = get_avg_query_speed(index_regular, xq, k)
        _, I_panorama = index_panorama.search(xq, k)
        panorama_q_speed = get_avg_query_speed(index_panorama, xq, k)
        _, gold_I = faiss.knn(xq, xb, k)
        result_line = f"""
{first_str}, \
regular_recall={compute_recall(gold_I, I_regular)}, \
regular_q_speed={regular_q_speed}, \
regular_train_time={regular_train_time}, \
panorama_recall={compute_recall(gold_I, I_panorama)}, \
panorama_q_speed={panorama_q_speed}, \
panorama_train_time={panorama_train_time}
"""
        print(result_line, flush=True)
```
The Flat results are all around 3000ms for this configuration for SyntheticDataset. Panorama results vary from 5000 to 8000.

Definitely let me know if you see issues in the methodology here, this was whipped together somewhat quickly. I can check the cvar a bit later.

aknayar:
@mnorris11 Thank you so much for sharing your code—I now notice something super important that I completely forgot to mention: When `nq` > 20 and there is no `sel` applied, `IndexFlatL2` will [trigger BLAS code](https://github.com/facebookresearch/faiss/blob/5d9f8d484cffd0961c374c1c7253c8c406b889df/faiss/utils/distances.cpp#L830-L834) to compute kNN. Since Panorama does a batch-wise refinement, we cannot accelerate it with BLAS. As such, Panorama will likely perform slower than L2Flat if `nq` > 20.

That being said, Panorama shines when any of the following is true:
 - `nq` <= 20
 - There is a `sel` applied
 - You are operating on a subset of the dataset

The third point is most important since this is exactly what `IndexRefine` does. As such, using `IndexRefinePanorama` (which will rely on this `IndexFlatPanorama`) will open the door to speedups regardless of `nq` or `sel`.

In the meantime, to properly gauge `IndexFlatPanorama`'s speedups, it is best to set `nq` to be less than 20. My apologies for forgetting to mention this earlier—I hope this helps.

mnorris11:
> @mnorris11 Thank you so much for sharing your code—I now notice something super important that I completely forgot to mention: When `nq` > 20 and there is no `sel` applied, `IndexFlatL2` will [trigger BLAS code](https://github.com/facebookresearch/faiss/blob/5d9f8d484cffd0961c374c1c7253c8c406b889df/faiss/utils/distances.cpp#L830-L834) to compute kNN. Since Panorama does a batch-wise refinement, we cannot accelerate it with BLAS. As such, Panorama will likely perform slower than L2Flat if `nq` > 20.
> 
> That being said, Panorama shines when any of the following is true:
> 
> * `nq` <= 20
> * There is a `sel` applied
> * You are operating on a subset of the dataset
> 
> The third point is most important since this is exactly what `IndexRefine` does. As such, using `IndexRefinePanorama` (which will rely on this `IndexFlatPanorama`) will open the door to speedups regardless of `nq` or `sel`.
> 
> In the meantime, to properly gauge `IndexFlatPanorama`'s speedups, it is best to set `nq` to be less than 20. My apologies for forgetting to mention this earlier—I hope this helps.

I actually do not see this line of code being hit in Run_search_L2sqr at all when debugging the search() section. progressive_filter_batch seems to just call fvec_inner_product. Am I missing something?

I tried to just pass
```
    params = faiss.SearchParameters()
    params.sel = faiss.IDSelectorAll()
    _, I_panorama = index_panorama.search(xq, k, params=params)
```
but this did not result in any change. (nq still 1000).

aknayar:
@mnorris11 The issue is that `IndexFlatL2` has this BLAS codepath which gets triggered for `nq` > 20 and when there's no `sel`, so I think the solution would be to pass these `params` for `index_regular` instead of the Panorama index.

mnorris11:
My mistake. Thanks for the swift replies. I see great speedup on all datasets now after passing the params correctly to the Panorama index!

It would be great if we can make this the default for Panorama. What do you think about the Panorama passing IDSelectorAll inside SearchParameters if they are not specified?

aknayar:
@mnorris11 No worries! I'm very glad you're now seeing some speedups! So it turns out that Panorama actually doesn't need the `params=params` at all—that doesn't have any effect on `index_panorama` since Panorama only has one codepath. The only reason for adding `params=params` at all is to make sure we don't trigger `index_base`'s BLAS code (to ensure a fair comparison, since Panorama fundamentally can't utilize BLAS in the same manner).

Just as a test, if you change your code to the following, you should still observe speedups:
```
...
# pass params into index_regular
_, I_regular = index_regular.search(xq, k, params=params)

# don't pass params into index_panorama
_, I_panorama = index_panorama.search(xq, k)
...
```

mdouze:
you can also change the variable `distance_compute_blas_threshold`

https://github.com/facebookresearch/faiss/blob/29f8e7263bff99e9d2bd334c2b958ec36f78d38e/faiss/utils/distances.h#L278

faiss/IndexFlat.cpp
```diff
@@ -519,4 +522,150 @@ void IndexFlat1D::search(
...
+
+    codes.resize(num_batches * batch_size * code_size);
+    cum_sums.resize(num_batches * batch_size * (n_levels + 1));
+
+    const uint8_t* code = reinterpret_cast<const uint8_t*>(x);
+    pano.copy_codes_to_level_layout(codes.data(), offset, n, code);
+    pano.compute_cumulative_sums(cum_sums.data(), offset, n, x);
+}
+
+template <bool use_radius, typename BlockHandler>
```
mdouze, line 532: what about puting this in an anonymous namespace?

faiss/IndexFlat.cpp
mdouze: I think the priority among the methods here would be `reconstruct` since it's what IVF indexes use. `remove_ids`, `merge_from`, etc. are less important.

faiss/impl/Panorama.h
```diff
@@ -0,0 +1,174 @@
...
+#include <faiss/impl/PanoramaStats.h>
+#include <faiss/utils/distances.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+namespace faiss {
+
```
mdouze, line 23: Please add a documentation block here (if it's redundant with IndexFlatPanorama redirect to it)

faiss/index_factory.cpp
```diff
@@ -569,6 +569,18 @@ Index* parse_other_indexes(
         return new IndexFlat(d, metric);
     }
 
+    // IndexFlatL2Panorama
+    if (match("FlatL2Panorama([0-9]+)(_[0-9]+)?")) {
```
mdouze, line 573: should be documented in the Faiss wiki https://github.com/facebookresearch/faiss/wiki/The-index-factory

aknayar: I can add `IndexFlatPanorama` and `IndexFlatL2Panorama` to the wiki. Should this be done post-merge? I'm also unsure on which section in the wiki this would fall into (besides the tree image that displays the hierarchy).

aknayar:
@mdouze Thanks for the comments! I have addressed them and implemented `reconstruct` (and `reconstruct_n`) for `IndexFlatPanorama`.

faiss/IndexFlat.cpp
```diff
@@ -656,7 +656,11 @@ void IndexFlatPanorama::reset() {
 }
 
 void IndexFlatPanorama::reconstruct(idx_t key, float* recons) const {
-    FAISS_THROW_MSG("reconstruct not implemented for IndexFlatPanorama");
+    pano.reconstruct(key, recons, codes.data());
+}
+
+void IndexFlatPanorama::reconstruct_n(idx_t i0, idx_t ni, float* recons) const {
+    Index::reconstruct_n(i0, ni, recons);
```
aknayar: This overrides `IndexFlatCodes`' implementation of `reconstruct_n`.

faiss/index_factory.cpp
mnorris11: @mdouze I can add to the wiki (both the Guidelines to Choosing an Index and the index factory) after merge.

faiss/IndexFlat.cpp
```diff
@@ -519,4 +522,164 @@ void IndexFlat1D::search(
...
+    IndexFlat::reset();
+    cum_sums.clear();
+}
+
+void IndexFlatPanorama::reconstruct(idx_t key, float* recons) const {
+    pano.reconstruct(key, recons, codes.data());
+}
+
+void IndexFlatPanorama::reconstruct_n(idx_t i, idx_t n, float* recons) const {
+    Index::reconstruct_n(i, n, recons);
```
aknayar, line 663: This overrides `IndexFlatCodes`' implementation of `reconstruct_n`.

mnorris11:
Thanks @aknayar @AlSchlo, the Flat PR has been merged!

aknayar:
@mnorris11 Thank you! I have a local build for `IndexRefinePanorama` ready and will aim to get that PR out for review tonight!

PR #4652, opened 2025-11-02
Optimize ScalarQuantizer

Ray0907:
RS_quantiles with nth_element

   Replace O(n log n) std::sort with O(n) std::nth_element for quantile
   computation. This addresses the TODO comment requesting quickselect.

   Changes:
   - Use std::nth_element to find lower and upper quantile positions
   - Eliminates unnecessary full sorting of x_copy vector
   - Maintains identical behavior for quantile calculation
   - Theoretical performance O(n log n) → O(n)

### DIFF
--- a/faiss/impl/ScalarQuantizer.cpp
+++ b/faiss/impl/ScalarQuantizer.cpp
@@ -1013,8 +1013,8 @@ void train_Uniform(
         if (o < 0) {
             o = 0;
         }
-        if (o > n - o) {
-            o = n / 2;
+        if (o > n - 1 - o) {
+            o = (n - 1) / 2;
         }
         // Use nth_element (O(n)) instead of sort (O(n log n))
         std::nth_element(x_copy.begin(), x_copy.begin() + o, x_copy.end());


subhadeepkaran:
feels like a corner case that needs to be handled here when `o > n/2`
then it opens the scenario for being `vmin > vmax`

i think we should ensure: `o = std::clamp(int(rs_arg * n), 0, n/2);`

i can put out a follow-up diff on this:

Ray0907:
You're right there's an issue. However, the correct bound should be (n-1)/2 instead of n/2 to ensure vmin_index <= vmax_index (i.e.,
   o <= n-1-o). I've updated the fix accordingly.

mnorris11:
Hi @Ray0907 , thanks for your contribution. Ditto the other changes: did you try benchmarking it? Do you observe speedup? If yes, can you paste the script you used here?

Ray0907:
@mnorris11  Not yet - this was based on the theoretical complexity improvement from the TODO. Happy to benchmark it! Any suggestions on test cases or data sizes I should focus on?

mnorris11:
> @mnorris11 Not yet - this was based on the theoretical complexity improvement from the TODO. Happy to benchmark it! Any suggestions on test cases or data sizes I should focus on?

To start with, how about creating an index using Faiss SyntheticDataset with nb=1M to 10M, nq=10k, k=100, d=[256,512,1024] and measuring the `train()` times before/after the change?

PR #4653, opened 2025-11-02
Use nth_element for median computation in IndexLSH

Ray0907:
Replace std::sort with std::nth_element for median calculation in
  IndexLSH training.

### DIFF
--- a/faiss/IndexLSH.cpp
+++ b/faiss/IndexLSH.cpp
@@ -86,16 +86,14 @@ void IndexLSH::train(idx_t n, const float* x) {
 
         for (idx_t i = 0; i < nbits; i++) {
             float* xi = transposed_x.get() + i * n;
-            // Use nth_element (O(n)) instead of sort (O(n log n)) for median
-            if (n % 2 == 1) {
-                std::nth_element(xi, xi + n / 2, xi + n);
-                thresholds[i] = xi[n / 2];
-            } else {
-                std::nth_element(xi, xi + n / 2, xi + n);
-                float median_high = xi[n / 2];
+            // Use nth_element (O(n)) instead of sort (O(n log n))
+            std::nth_element(xi, xi + n / 2, xi + n);
+            float median = xi[n / 2];
+            if (n % 2 == 0) {
                 std::nth_element(xi, xi + n / 2 - 1, xi + n);
-                thresholds[i] = (xi[n / 2 - 1] + median_high) / 2;
+                median = (median + xi[n / 2 - 1]) / 2;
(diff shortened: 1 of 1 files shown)


faiss/IndexLSH.cpp
```diff
@@ -86,12 +86,16 @@ void IndexLSH::train(idx_t n, const float* x) {
...
+                std::nth_element(xi, xi + n / 2, xi + n);
                 thresholds[i] = xi[n / 2];
-            else
-                thresholds[i] = (xi[n / 2 - 1] + xi[n / 2]) / 2;
+            } else {
+                std::nth_element(xi, xi + n / 2, xi + n);
+                float median_high = xi[n / 2];
+                std::nth_element(xi, xi + n / 2 - 1, xi + n);
+                thresholds[i] = (xi[n / 2 - 1] + median_high) / 2;
+            }
```
subhadeepkaran: maybe we can do like this?


```
// Use nth_element (O(n)) instead of sort (O(n log n)) for median
std::nth_element(xi, xi + n / 2, xi + n);
float median = xi[n / 2];
if (n % 2 == 0) {
    std::nth_element(xi, xi + n / 2 - 1, xi + n);
    median = (median + xi[n / 2 - 1]) / 2;
}
thresholds[i] = median;
```

Ray0907: Good point! Your version is cleaner!

mnorris11:
Hi @Ray0907 , thanks for your contribution. Ditto the other perf changes: did you try benchmarking it? Do you observe speedup? If yes, can you paste the script you used here?

mdouze:
I agree that it's the proper way of computing the median but the performance improvement could be minimal.

faiss/IndexLSH.cpp
```diff
@@ -86,12 +86,14 @@ void IndexLSH::train(idx_t n, const float* x) {
...
-            std::sort(xi, xi + n);
-            if (n % 2 == 1)
-                thresholds[i] = xi[n / 2];
-            else
-                thresholds[i] = (xi[n / 2 - 1] + xi[n / 2]) / 2;
+            // Use nth_element (O(n)) instead of sort (O(n log n))
+            std::nth_element(xi, xi + n / 2, xi + n);
+            float median = xi[n / 2];
+            if (n % 2 == 0) {
+                std::nth_element(xi, xi + n / 2 - 1, xi + n);
```
alexanderguzhva, line 93: looks good, unless running 2x `std::nth_element()` is too costly.

PR #4654, opened 2025-11-02
Parallelize compute_residuals in IndexIVFPQ

Ray0907:
Add OpenMP parallelization to compute_residuals function.
  This addresses the TODO comment about parallelization.

  Changes:
  - Add #pragma omp parallel for to residual computation loop
  - Use if(n > 1000) to avoid overhead for small batches
  - Each iteration is independent and thread-safe


mnorris11:
Hi @Ray0907  did you try benchmarking it? Do you observe speedup? If yes, can you paste the script you used here?

Ray0907:
@mnorris11 
yes I benchmarked it. but the results show only 3-4% improvement.

PR #4679, opened 2025-11-17
Implement multi-bit RaBitQ quantization (nb_bits 2-9)

alibeklfc:
Summary:
Extends RaBitQ quantization from 1-bit to support 2-9 bits per dimension, enabling flexible accuracy/memory trade-offs. Multi-bit RaBitQ achieves significantly higher recall than 1-bit (e.g., 4-bit: ~88% vs 1-bit: ~52% on benchmark datasets) with minimal performance overhead via two-stage search.

## Motivation

While 1-bit RaBitQ offers excellent compression, recall performance varies depending on dataset characteristics and query requirements. On some datasets, 1-bit recall may be sufficient, but on others, higher accuracy is needed. This diff adds multi-bit support (2-9 bits) to provide flexible accuracy/memory trade-offs, bridging the gap between 1-bit RaBitQ and higher-memory quantizers like PQ/SQ.

### Quantization Algorithm (`RaBitQuantizerMultiBit`)

Multi-bit encoding uses split representation:
- **1 sign bit**: Encodes residual sign (positive/negative)
- **ex_bits magnitude bits**: Encodes residual magnitude (0 to 2^ex_bits - 1)

For each vector:
1. Compute residual: `r = x - centroid`
2. Normalize and take absolute value: `o_abs = |r| / ||r||`
3. Find optimal scaling factor `t` via priority queue search (maximizes quantized/original inner product)
4. Quantize magnitudes: `ex_code[i] = min(⌊t * o_abs[i]⌋, 2^ex_bits - 1)`
5. Handle negative dimensions via bit flipping: `ex_code[i] = ~ex_code[i] & max_code` if `residual[i] < 0`
6. Pack into bytes and compute distance factors (`ExFactorsData`)

**Code Layout** (per vector):
```
[1-bit codes: (d+7)/8 bytes]
[Base factors: 12 bytes (or_minus_c_l2sqr, dp_multiplier, f_error)]
[Ex-bit codes: (d*ex_bits+7)/8 bytes]  // only if nb_bits > 1
[Ex factors: 8 bytes (f_add_ex, f_rescale_ex)]  // only if nb_bits > 1
```

### Two-Stage Search (`RaBitQDistanceComputer`)

Introduced new base class with three distance methods:
- `distance_to_code_1bit()`: Fast 1-bit estimate using sign bits only
- `lower_bound_distance()`: 1-bit estimate minus error bound (f_error * g_error)
- `distance_to_code_full()`: Accurate distance using all bits

**Search Strategy:**
- **1-bit** (nb_bits=1): Single-stage search using sign bits
- **Multi-bit** (nb_bits>1): Two-stage search with adaptive filtering
  1. Compute 1-bit lower bound for all candidates
  2. Filter candidates where lower_bound < current_threshold
  3. Compute full multi-bit distance only for promising candidates
  4. Update heap with refined distances

Applied in both `IndexRaBitQ::search()` and `IndexIVFRaBitQ::scan_codes()`.

### Index API Changes

**IndexRaBitQ / IndexIVFRaBitQ:**
```cpp
// Constructor now accepts nb_bits parameter (default=1 for backward compatibility)
IndexRaBitQ(idx_t d, MetricType metric = METRIC_L2, uint8_t nb_bits = 1);
IndexIVFRaBitQ(Index* quantizer, size_t d, size_t nlist,
               MetricType metric = METRIC_L2, bool own_invlists = true,
               uint8_t nb_bits = 1);

// New field
uint8_t nb_bits;  // Range: 1-9, validated in constructor
```

**Index Factory Syntax:**
```
"RaBitQ"          → 1-bit (backward compatible)
"RaBitQ2"         → 2-bit
"RaBitQ4"         → 4-bit
"RaBitQ8"         → 8-bit
"IVF16,RaBitQ"    → 1-bit IVF
"IVF16,RaBitQ4"   → 4-bit IVF
```

### Serialization

Updated `index_read.cpp` and `index_write.cpp` to persist `nb_bits` field. Maintains backward compatibility with existing 1-bit indexes.

## Performance Characteristics

**Memory Usage** (per vector, dimension d):
- **1-bit**: `(d+7)/8 + 12` bytes
  - 1-bit codes + base factors (FactorsData: 12 bytes)
- **2-bit**: `(d+7)/8 + 12 + (d+7)/8 + 8` bytes
  - 1-bit codes + base factors + 1-bit ex-codes + ex factors (ExFactorsData: 8 bytes)
- **4-bit**: `(d+7)/8 + 12 + (3*d+7)/8 + 8` bytes
  - 1-bit codes + base factors + 3-bit ex-codes + ex factors
- **8-bit**: `(d+7)/8 + 12 + (7*d+7)/8 + 8` bytes
  - 1-bit codes + base factors + 7-bit ex-codes + ex factors

**Example (d=128):**
- 1-bit: 28 bytes
- 2-bit: 52 bytes (1.86x)
- 4-bit: 84 bytes (3.0x)
- 8-bit: 148 bytes (5.3x)

**Recall @ 10** (measured on synthetic data, d=128, averaged across L2/IP metrics and qb values):
- 1-bit: ~52% recall
- 2-bit: ~74% recall (+42% relative improvement)
- 4-bit: ~88% recall (+68% relative improvement)
- 8-bit: ~95% recall (+81% relative improvement)

**Search Performance:**
- 1-bit: Single-stage, fastest
- Multi-bit: Two-stage with adaptive filtering, ~1.2-1.5x overhead vs 1-bit, significantly faster than full scan

## Files Changed

**New Files:**
- `impl/RaBitQuantizerMultiBit.{cpp,h}`: Multi-bit quantization algorithms

**Modified Core:**
- `impl/RaBitQuantizer.{cpp,h}`: Extended with nb_bits, two-stage distance computers
- `impl/RaBitQUtils.{cpp,h}`: Added ExFactorsData, f_error computation
- `IndexRaBitQ.{cpp,h}`: Added nb_bits field, two-stage search
- `IndexIVFRaBitQ.{cpp,h}`: Added nb_bits field, scan_codes override

**Factory & Serialization:**
- `index_factory.cpp`: Added "RaBitQ{nb_bits}" syntax parsing
- `impl/index_read.cpp`: Read nb_bits field
- `impl/index_write.cpp`: Write nb_bits field

**Tests:**
- `tests/test_rabitq.py`: Added 6 test classes, 78+ test methods
- `xplat.bzl`: Added RaBitQuantizerMultiBit to build

Differential Revision: D87190125

### DIFF
--- a/faiss/IndexIVFRaBitQ.cpp
+++ b/faiss/IndexIVFRaBitQ.cpp
@@ -24,9 +24,10 @@ IndexIVFRaBitQ::IndexIVFRaBitQ(
         const size_t d,
         const size_t nlist,
         MetricType metric,
-        bool own_invlists)
+        bool own_invlists,
+        uint8_t nb_bits_in)
         : IndexIVF(quantizer, d, nlist, 0, metric, own_invlists),
-          rabitq(d, metric) {
+          rabitq(d, metric, nb_bits_in) {
     code_size = rabitq.code_size;
     if (own_invlists) {
         invlists->code_size = code_size;
@@ -153,6 +154,8 @@ struct RaBitInvertedListScanner : InvertedListScanner {
     std::vector<float> query_vector;
 
     std::unique_ptr<FlatCodesDistanceComputer> dc;
+    RaBitQDistanceComputer* rabitq_dc =
+            nullptr; // For multi-bit adaptive filtering
 
     uint8_t qb = 0;
     bool centered = false;
@@ -194,6 +197,68 @@ struct RaBitInvertedListScanner : InvertedListScanner {
         return dc->distance_to_code(code);
     }
 
+    /// Override scan_codes to implement adaptive filtering for multi-bit codes
+    size_t scan_codes(
+            size_t list_size,
+            const uint8_t* codes,
+            const idx_t* ids,
+            float* simi,
+            idx_t* idxi,
+            size_t k) const override {
+        size_t ex_bits = ivf_rabitq.rabitq.nb_bits - 1;
+
+        // For 1-bit codes, use default implementation
+        if (ex_bits == 0 || rabitq_dc == nullptr) {
+            return InvertedListScanner::scan_codes(
+                    list_size, codes, ids, simi, idxi, k);
+        }
+
+        // Multi-bit: Two-stage search with adaptive filtering
+        size_t nup = 0;
+
+        for (size_t j = 0; j < list_size; j++) {
+            if (sel != nullptr) {
+                int64_t id = store_pairs ? lo_build(list_no, j) : ids[j];
+                if (!sel->is_member(id)) {
+                    codes += code_size;
+                    continue;
+                }
+            }
+
+            // Stage 1: Compute lower bound using 1-bit codes
+            float lower_bound = rabitq_dc->lower_bound_distance(codes);
+
+            // Stage 2: Adaptive filtering
+            // L2 (min-heap): filter if lower_bound < simi[0]
+            // IP (max-heap): filter if lower_bound > simi[0]
+            // Note: Using simi[0] directly (not cached) enables more aggressive
+            // filtering as the heap is updated with better candidates
+            bool is_promising = keep_max ? (lower_bound > simi[0])
+                                         : (lower_bound < simi[0]);
+
+            if (is_promising) {
+                // Lower bound is promising, compute full distance
+                float dis = distance_to_code(codes);
+
+                // Check if distance improves heap
+                bool improves_heap =
+                        keep_max ? (dis > simi[0]) : (dis < simi[0]);
+
+                if (improves_heap) {
+                    int64_t id = store_pairs ? lo_build(list_no, j) : ids[j];
+                    if (keep_max) {
+                        minheap_replace_top(k, simi, idxi, dis, id);
+                    } else {
+                        maxheap_replace_top(k, simi, idxi, dis, id);
+                    }
+                    nup++;
+                }
+            }
+            codes += code_size;
+        }
+        return nup;
+    }
+
     void internal_try_setup_dc() {
         if (!query_vector.empty() && !reconstructed_centroid.empty()) {
             // both query_vector and centroid are available!
@@ -202,6 +267,9 @@ struct RaBitInvertedListScanner : InvertedListScanner {
                     qb, reconstructed_centroid.data(), centered));
 
             dc->set_query(query_vector.data());
+
+            // Try to cast to RaBitQDistanceComputer for multi-bit support
+            rabitq_dc = dynamic_cast<RaBitQDistanceComputer*>(dc.get());
         }
     }
 };
--- a/faiss/impl/RaBitQUtils.cpp
+++ b/faiss/impl/RaBitQUtils.cpp
@@ -61,6 +61,8 @@ FactorsData compute_factors_from_intermediates(
         size_t d,
         MetricType metric_type) {
     constexpr float epsilon = std::numeric_limits<float>::epsilon();
+    constexpr float kConstEpsilon =
+            1.9f; // Error bound constant from RaBitQ paper
     const float inv_d_sqrt =
             (d == 0) ? 1.0f : (1.0f / std::sqrt(static_cast<float>(d)));
 
@@ -77,6 +79,35 @@ FactorsData compute_factors_from_intermediates(
             ? (norm_L2sqr - or_L2sqr)
             : norm_L2sqr;
     factors.dp_multiplier = inv_dp_oO * sqrt_norm_L2;
+
+    // Compute error bound for 1-bit quantization
+    const float xu_cb_norm_sqr = static_cast<float>(d) * 0.25f;
+    const float ip_resi_xucb = 0.5f * dp_oO;
+
+    float tmp_error = 0.0f;
+    if (std::abs(ip_resi_xucb) > epsilon) {
+        const float ratio_sq =
+                (norm_L2sqr * xu_cb_norm_sqr) / (ip_resi_xucb * ip_resi_xucb);
+        if (ratio_sq > 1.0f) {
+            if (d == 1) {
+                tmp_error = sqrt_norm_L2 * kConstEpsilon *
+                        std::sqrt(ratio_sq - 1.0f);
+            } else {
+                tmp_error = sqrt_norm_L2 * kConstEpsilon *
+                        std::sqrt((ratio_sq - 1.0f) /
+                                  static_cast<float>(d - 1));
+            }
+        }
+    }
+
+    // Apply metric-specific multiplier
+    if (metric_type == MetricType::METRIC_L2) {
+        factors.f_error = 2.0f * tmp_error;
+    } else if (metric_type == MetricType::METRIC_INNER_PRODUCT) {
+        factors.f_error = 1.0f * tmp_error;
+    } else {
+        factors.f_error = 0.0f;
+    }
 
     return factors;
 }
--- a/faiss/impl/RaBitQuantizer.cpp
+++ b/faiss/impl/RaBitQuantizer.cpp
@@ -9,6 +9,7 @@
 
 #include <faiss/impl/FaissAssert.h>
 #include <faiss/impl/RaBitQUtils.h>
+#include <faiss/impl/RaBitQuantizerMultiBit.h>
 #include <faiss/utils/distances.h>
 #include <faiss/utils/rabitq_simd.h>
 #include <algorithm>
@@ -20,15 +21,45 @@
 namespace faiss {
 
 // Import shared utilities from RaBitQUtils
+using rabitq_utils::BaseFactorsData;
+using rabitq_utils::ExFactorsData;
 using rabitq_utils::FactorsData;
 using rabitq_utils::QueryFactorsData;
 
-static size_t get_code_size(const size_t d) {
-    return (d + 7) / 8 + sizeof(FactorsData);
+RaBitQuantizer::RaBitQuantizer(size_t d, MetricType metric, size_t nb_bits)
+        : Quantizer(d, 0), // code_size will be set below
+          metric_type{metric},
+          nb_bits{nb_bits} {
+    // Validate nb_bits range
+    FAISS_THROW_IF_NOT(nb_bits >= 1 && nb_bits <= 9);
+
+    // Set code_size using compute_code_size
+    code_size = compute_code_size(d, nb_bits);
 }
 
-RaBitQuantizer::RaBitQuantizer(size_t d, MetricType metric)
-        : Quantizer(d, get_code_size(d)), metric_type{metric} {}
+size_t RaBitQuantizer::compute_code_size(size_t d, size_t num_bits) const {
+    // Validate inputs
+    FAISS_THROW_IF_NOT(num_bits >= 1 && num_bits <= 9);
+
+    size_t ex_bits = num_bits - 1;
+
+    // Base: 1-bit codes + base factors
+    // Layout for 1-bit: [binary_code: (d+7)/8 bytes][BaseFactorsData: 8 bytes]
+    //   base_factors = or_minus_c_l2sqr (4) + dp_multiplier (4)
+    // Layout for multi-bit: [binary_code: (d+7)/8 bytes][FactorsData: 12 bytes]
+    //   factors = or_minus_c_l2sqr (4) + dp_multiplier (4) + f_error (4)
+    size_t base_size = (d + 7) / 8 +
+            (ex_bits == 0 ? sizeof(BaseFactorsData) : sizeof(FactorsData));
+
+    // Extra: ex-bit codes + ex factors (only if ex_bits > 0)
+    // Layout: [ex_code: (d*ex_bits+7)/8 bytes][ex_factors: 8 bytes]
+    size_t ex_size = 0;
+    if (ex_bits > 0) {
+        ex_size = (d * ex_bits + 7) / 8 + sizeof(ExFactorsData);
+    }
+
+    return base_size + ex_size;
+}
 
 void RaBitQuantizer::train(size_t n, const float* x) {
     // does nothing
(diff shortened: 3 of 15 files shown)


faiss/impl/RaBitQuantizerMultiBit.h
```diff
@@ -0,0 +1,102 @@
...
+
+#include <faiss/MetricType.h>
+#include <faiss/impl/RaBitQUtils.h>
+#include <cstddef>
+#include <cstdint>
+
+namespace faiss {
+namespace rabitq_multibit {
+
+/**
```
alexanderguzhva, line 28: Please don't forget to add a reference to the article, authors, full article name & arxiv.org

faiss/IndexIVFRaBitQ.cpp
```diff
@@ -24,9 +24,15 @@ IndexIVFRaBitQ::IndexIVFRaBitQ(
...
-        bool own_invlists)
+        bool own_invlists,
+        uint8_t nb_bits_in)
         : IndexIVF(quantizer, d, nlist, 0, metric, own_invlists),
-          rabitq(d, metric) {
+          rabitq(d, metric, nb_bits_in),
+          nb_bits(nb_bits_in) {
+    // Validate nb_bits range
+    FAISS_THROW_IF_NOT_MSG(
+            nb_bits >= 1 && nb_bits <= 9, "nb_bits must be in range [1, 9]");
```
alexanderguzhva: any justification why 1..9 and not 16, for example?

faiss/impl/index_read.cpp
```diff
@@ -612,6 +612,7 @@ static void read_RaBitQuantizer(RaBitQuantizer* rabitq, IOReader* f) {
     READ1(rabitq->d);
     READ1(rabitq->code_size);
     READ1(rabitq->metric_type);
+    READ1(rabitq->nb_bits);
```
alexanderguzhva: This breaks the compatibility with already existing indices around the world, including ones used by Milvus

faiss/impl/index_read.cpp
```diff
@@ -1305,6 +1306,7 @@ Index* read_index(IOReader* f, int io_flags) {
         READVECTOR(idxq->codes);
         READVECTOR(idxq->center);
         READ1(idxq->qb);
+        READ1(idxq->nb_bits);
```
alexanderguzhva: same

faiss/impl/index_read.cpp
```diff
@@ -1314,6 +1316,12 @@ Index* read_index(IOReader* f, int io_flags) {
         READ1(ivrq->code_size);
         READ1(ivrq->by_residual);
         READ1(ivrq->qb);
+        READ1(ivrq->nb_bits);
```
alexanderguzhva: same

faiss/impl/index_write.cpp
```diff
@@ -401,6 +401,7 @@ static void write_RaBitQuantizer(const RaBitQuantizer* rabitq, IOWriter* f) {
     WRITE1(rabitq->d);
     WRITE1(rabitq->code_size);
     WRITE1(rabitq->metric_type);
+    WRITE1(rabitq->nb_bits);
```
alexanderguzhva: same

faiss/impl/index_write.cpp
```diff
@@ -935,6 +936,7 @@ void write_index(const Index* idx, IOWriter* f, int io_flags) {
         WRITEVECTOR(idxq->codes);
         WRITEVECTOR(idxq->center);
         WRITE1(idxq->qb);
+        WRITE1(idxq->nb_bits);
```
alexanderguzhva: same

faiss/impl/index_write.cpp
```diff
@@ -945,6 +947,7 @@ void write_index(const Index* idx, IOWriter* f, int io_flags) {
         WRITE1(ivrq->code_size);
         WRITE1(ivrq->by_residual);
         WRITE1(ivrq->qb);
+        WRITE1(ivrq->nb_bits);
```
alexanderguzhva: same :(

faiss/IndexIVFRaBitQ.cpp
```diff
@@ -194,6 +202,68 @@ struct RaBitInvertedListScanner : InvertedListScanner {
...
+                    codes += code_size;
+                    continue;
+                }
+            }
+
+            // Stage 1: Compute lower bound using 1-bit codes
+            float lower_bound = rabitq_dc->lower_bound_distance(codes);
+
+            // Stage 2: Adaptive filtering
+            // L2 (min-heap): filter if lower_bound < simi[0]
```
alexanderguzhva, line 232: `// L2 (min-heap): filter if lower_bound <= simi[0]`, maybe, not `<` ?

faiss/impl/RaBitQUtils.cpp
```diff
@@ -78,6 +80,29 @@ FactorsData compute_factors_from_intermediates(
             : norm_L2sqr;
     factors.dp_multiplier = inv_dp_oO * sqrt_norm_L2;
 
+    // Compute error bound for 1-bit quantization
+    const float xu_cb_norm_sqr = static_cast<float>(d) * 0.25f;
+    const float ip_resi_xucb = 0.5f * dp_oO;
+
+    float tmp_error = 0.0f;
+    if (std::abs(ip_resi_xucb) > epsilon && d > 1) {
```
alexanderguzhva: and if `d == 1` ?

faiss/impl/RaBitQUtils.cpp
```diff
@@ -78,6 +80,29 @@ FactorsData compute_factors_from_intermediates(
...
+        const float ratio_sq =
+                (norm_L2sqr * xu_cb_norm_sqr) / (ip_resi_xucb * ip_resi_xucb);
+        if (ratio_sq > 1.0f) {
```
alexanderguzhva, line 91: what if `ratio_sq <= 1` ?

faiss/impl/RaBitQuantizer.cpp
```diff
@@ -124,101 +182,149 @@ void RaBitQuantizer::decode_core(
...
-    const FactorsData* fac =
+    const FactorsData* base_fac =
             reinterpret_cast<const FactorsData*>(code + (d + 7) / 8);
 
-    // this is the baseline code
-    //
-    // compute <q,o> using floats
+    // Compute dot product using 1-bit codes
     float dot_qo = 0;
-    // It was a willful decision (after the discussion) to not to pre-cache
```
alexanderguzhva, line 177: I'd keep some / many comments

faiss/impl/RaBitQuantizer.cpp
```diff
@@ -301,48 +421,114 @@ float RaBitDistanceComputerQ::distance_to_code(const uint8_t* code) {
...
     if (metric_type == MetricType::METRIC_L2) {
-        // ||or - q||^ 2
         return pre_dist;
     } else {
-        // metric == MetricType::METRIC_INNER_PRODUCT
+        return -0.5f * (pre_dist - query_fac.qr_norm_L2sqr);
+    }
+}
+
+float RaBitQDistanceComputerQ::lower_bound_distance(const uint8_t* code) {
```
alexanderguzhva: maybe, move this function into a common `RaBitQDistanceComputer` class, because it looks identical for both derived classes, at least as a base implementation?
```C++
struct RaBitQDistanceComputer {
    virtual float lower_bound_distance(const uint8_t* code);
}

float RaBitQDistanceComputer::lower_bound_distance(const uint8_t* code) {
...
}
```

faiss/impl/RaBitQuantizer.h
```diff
@@ -71,9 +87,58 @@ struct RaBitQuantizer : Quantizer {
...
+//
+// This intermediate class exists to provide a unified interface for
+// two-stage multi-bit search. While most Faiss quantizers extend
+// FlatCodesDistanceComputer directly, RaBitQ requires this additional
+// abstraction layer due to its unique split encoding strategy
+// (1 sign bit + magnitude bits) which enables:
+//
+// 1. distance_to_code_1bit() - Fast 1-bit filtering using only sign bits
+// 2. distance_to_code_full() - Accurate multi-bit refinement using all bits
+// 3. lower_bound_distance() - Error-bounded adaptive filtering
```
alexanderguzhva, line 105: please add a note that `lower_bound_distance()` is based on 1-bit estimator, if I understand the code correctly

faiss/impl/RaBitQuantizerMultiBit.cpp
```diff
@@ -0,0 +1,358 @@
...
+ * explore different quantization levels.
+ *
+ *
+ * @param o_abs Normalized absolute residual vector (must be positive, length
+ * d)
+ * @param d Dimensionality of the vector
+ * @param nb_bits Number of bits per dimension (2-9)
+ * @return Optimal scaling factor 't'
+ */
+double compute_optimal_scaling_factor(
```
alexanderguzhva: change to `float`, because all the computations are performed in floats anyways
```C++
float compute_optimal_scaling_factor(...) {...}
```

faiss/impl/RaBitQuantizerMultiBit.cpp
```diff
@@ -0,0 +1,358 @@
...
+        xu_cb[i] = static_cast<float>(tmp_code[i]) + cb;
+    }
+
+    // Compute inner products needed for factors
+    float l2_sqr = norm * norm;
+    float ip_resi_xucb = fvec_inner_product(residual, xu_cb.data(), d);
+
+    // Handle corner case
+    if (ip_resi_xucb == 0) {
+        ip_resi_xucb = std::numeric_limits<float>::infinity();
```
alexanderguzhva: I'm not sure I follow this particular line

faiss/impl/RaBitQuantizerMultiBit.cpp
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    // Step 3: Take absolute value
+    std::vector<float> o_abs(d);
+    for (size_t i = 0; i < d; i++) {
+        o_abs[i] = std::abs(normalized_residual[i]);
+    }
+
+    // Step 4: Find optimal scaling factor
+    double t = compute_optimal_scaling_factor(o_abs.data(), d, nb_bits);
```
alexanderguzhva: `float t = ...` ?

alexanderguzhva:
@alibeklfc looks good as a reference implementation, if it works, ofc. The most severe problem is going to be the backward compability with serializing/deserializing 1-bit RaBitQ indices, which are already widely used in Milvus and maybe everywhere else...

faiss/IndexIVFRaBitQ.cpp
alibeklfc: lower_bound is the best possible distance for the vector, and the simi[0] is the current worst in the heap. So, if lower_bound == simi[0], it cannot improve current worst in the heap

faiss/impl/RaBitQuantizerMultiBit.h
alibeklfc: Added comment: // Reference:
// "Practical and asymptotically optimal quantization of high-dimensional
// vectors in euclidean space for approximate nearest neighbor search"
// Jianyang Gao, Yutong Gou, Yuexuan Xu, Yongyi Yang, Cheng Long, Raymond
// Chi-Wing Wong https://dl.acm.org/doi/pdf/10.1145/3725413

faiss/impl/index_read.cpp
alibeklfc: I believe, I have resolved this issue in the updated version. Can you please verify?

faiss/impl/RaBitQUtils.cpp
alibeklfc: Modified this part, now we compute an error for d = 1

faiss/impl/RaBitQuantizer.cpp
alibeklfc: Reverted most of the comment deletion

faiss/impl/RaBitQuantizer.cpp
alibeklfc: Done

faiss/impl/RaBitQuantizer.h
alibeklfc: Added the note

faiss/impl/RaBitQuantizerMultiBit.cpp
alibeklfc: Changed to float

faiss/impl/RaBitQuantizerMultiBit.cpp
alibeklfc: Resolved

faiss/impl/RaBitQUtils.cpp
alibeklfc: ratio_sq should never be smaller than 1. when ratio_sq == 1, it should tmp_error should be 0. So, the code is correct here

faiss/impl/RaBitQuantizerMultiBit.cpp
alibeklfc: If we do not change the value, there might be a division by zero error. With this change, we basically set  l2_sqr * ip_cent_xucb / infinity = 0 later. This part of the code was taken from the original authors

alexanderguzhva: Imagine that your residual consists of zeroes, which is a totally valid case. I would just handle it separately, I think.
Also, are there any guarantees that `l2_sqr * ip_cent_xucb / infinity` behaves as expected if you're using `-ffast-math` computations?

alexanderguzhva: Also, there are no references to original authors github code, if you're using some other's code...

faiss/impl/index_write.cpp
```diff
@@ -401,6 +401,11 @@ static void write_RaBitQuantizer(const RaBitQuantizer* rabitq, IOWriter* f) {
     WRITE1(rabitq->d);
     WRITE1(rabitq->code_size);
     WRITE1(rabitq->metric_type);
+    // Only write nb_bits for multi-bit indices (backward compatibility)
+    size_t expected_1bit_code_size = (rabitq->d + 7) / 8 + sizeof(FactorsData);
```
alexanderguzhva: So, this means that if an old library reads a binary index generated by a new library, then it crashes with a run-time error, and not with a correct expected error handling.

Please use a different approach, as suggested in our meeting:
```C++
if (const IndexRaBitQ* idxq = dynamic_cast<const IndexRaBitQ*>(idx)) {
    if (one_bit_rabitq(idxq)) {
        // old code here
        uint32_t h = fourcc("Ixrq");
        ...
    } else { 
        // new multi-bit code here for a different four cc
        // btw, Milvus uses "IxrQ" as a bit different serialization format for rabitq,
        //   so I would appreciate if you could use something not like "IxrQ", if possible
        uint32_t h = fourcc("Ixrr");
        ...
    }
}
```

Same story for IVFRaBitQ (Milvus uses `IwrQ` instead of baseline `Iwrq`) and others

faiss/impl/RaBitQuantizerMultiBit.cpp
alibeklfc: Added a reference to the rabitq-library code  in the rabitquantizermultibit file.

alibeklfc: Resolved in the new version of the code. Removed ip_resi_xucb == 0 check. We calculate correction term (l2_sqr * ip_cent_xucb / ip_resi_xucb ) separately

faiss/impl/index_write.cpp
alibeklfc: We now use Ixrr for the new version and Ixrq for the old version. I have also implemented this change to the IVFRaBitQ version.

faiss/impl/index_write.cpp
```diff
@@ -397,11 +397,33 @@ static void write_NNDescent(const NNDescent* nnd, IOWriter* f) {
     WRITEVECTOR(nnd->final_graph);
 }
 
-static void write_RaBitQuantizer(const RaBitQuantizer* rabitq, IOWriter* f) {
+// Helper function to determine if RaBitQ index is 1-bit
+static bool is_rabitq_1bit(const RaBitQuantizer* rabitq) {
+    size_t expected_1bit_code_size =
+            (rabitq->d + 7) / 8 + sizeof(rabitq_utils::BaseFactorsData);
+    return rabitq->code_size == expected_1bit_code_size && rabitq->nb_bits == 1;
```
alexanderguzhva: Just confirming that a simple `if (rabitq->nb_bits == 1)` insufficient.
If so, then a comment is needed why so

alexanderguzhva:
@alibeklfc I would also do a simple one time test. Please serialize an index using a faiss version which has no this PR, and deserialize it into this PR-based faiss. Also, serialize 1-bit and n-bit with this PR-based faiss and try to deserialize in a version without this PR. 
The code looks good, but a simple one time (!) test won't hurt. There's no need to add a unit test for that.

faiss/impl/index_write.cpp
alibeklfc: if (rabitq->nb_bits == 1) is sufficient. I have updated this part. Thank you!

alibeklfc:
@alexanderguzhva I manually tested serialization and deserialization. 
OLD indices → NEW PR branch: All passed

NEW indices → OLD master branch:

1-bit indices  - passed!
Multi-bit indices  - failed  with error "Index type 0x72727749 ("Iwrr") not recognized" which is expected

PR #4683, opened 2025-11-18
Implement `IndexRefinePanorama`

aknayar:
This PR implements `IndexRefinePanorama`, integrating point-wise Panorama (as specified in the [paper](https://arxiv.org/abs/2510.00566)) into `IndexRefine`.

## IndexRefinePanorama
In order to enable point-wise Panorama refinement on a subset of candidates, I have added a `search_subset` to Index.h. There are two main reasons behind this:
1. `IndexFlat`'s `compute_distance_subset` is not an API Panorama can conform to. By definition, Panorama's gains come from _avoiding_ full distance computations. Thus, a new `search_subset` is introduced.
2. Since Panorama indexes are typically used within an `IndexPreTransform`, both `IndexPreTransform` and `IndexFlatPanorama` need to have this new `search_subset`. I could just define it within those indexes, but then checking that `IndexRefinePanorama`'s `refine_index` has `search_subset` becomes a hairy mess of `dynamic_casts`.  

For the above reasons, I have defined `search_subset` in Index.h, with a default implementation that throws an error. I would understand if this approach is not preferred and am open to any suggestions folks may have.

## Speedups
For two key reasons, `IndexRefinePanorama` is expected to have less speedups than other Panorama indexes:
1. Loss of contiguous memory accesses: Since the `refine_index` operates on a (typically sparse) subset of the total dataset, we lose all gains due to memory layout changes and batching.
2. The unavoidable cost of the `base_index`

Despite these limitations, `IndexRefinePanorama` is able to provide 2x end-to-end speedups (seen below).

## Testing
1. Correctness tests can be found in `tests/test_refine_panorama.py`
4. A benchmark can be found in `benchs/bench_refine_panorama.py,` yielding the following results:
```
Benchmark on GIST1M with base 'IVF256,PQ60x4fs', k=10, nq=100
nprobe  k_factor   recall_flat   qps_flat   recall_pano   qps_pano   dims_scanned(%)  speedup(x)
     4      1.0      0.037000   11705.14      0.037000   11107.50            100.00       0.95x
     4      8.0      0.149000    7341.04      0.149000    7331.16             53.14       1.00x
     4     64.0      0.336000    2010.27      0.336000    2783.57             26.82       1.38x
     4    256.0      0.404000     609.62      0.404000    1086.37             19.66       1.78x
     4   1024.0      0.425000     181.52      0.425000     422.38             15.52       2.33x
    16      1.0      0.036000    5894.10      0.036000    5702.58            100.00       0.97x
    16      8.0      0.156000    3381.69      0.156000    3418.20             53.61       1.01x
    16     64.0      0.473000    1473.81      0.473000    1847.22             27.05       1.25x
    16    256.0      0.700000     559.01      0.700000     937.59             20.09       1.68x
    16   1024.0      0.818000     171.81      0.818000     365.66             16.28       2.13x
    64      1.0      0.036000    1876.12      0.036000    1778.08            100.00       0.95x
    64      8.0      0.157000    1181.66      0.157000    1292.00             53.58       1.09x
    64     64.0      0.488000     783.60      0.488000    1146.26             26.94       1.46x
    64    256.0      0.741000     422.59      0.741000     635.95             20.04       1.50x
    64   1024.0      0.942000     158.79      0.942000     299.14             16.32       1.88x
   256      1.0      0.036000     650.09      0.036000     640.78            100.00       0.99x
   256      8.0      0.157000     553.64      0.157000     617.19             53.52       1.11x
   256     64.0      0.491000     450.86      0.491000     549.18             26.93       1.22x
   256    256.0      0.746000     309.26      0.746000     413.02             20.02       1.34x
   256   1024.0      0.948000     138.77      0.948000     232.52             16.31       1.68x
   ```
bench_refine_panorama

### DIFF
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -103,13 +103,22 @@ namespace {
 struct FlatL2Dis : FlatCodesDistanceComputer {
     size_t d;
     idx_t nb;
-    const float* q;
     const float* b;
     size_t ndis;
+    size_t npartial_dot_products;
 
     float distance_to_code(const uint8_t* code) final {
         ndis++;
         return fvec_L2sqr(q, (float*)code, d);
+    }
+
+    float partial_dot_product(
+            const idx_t i,
+            const uint32_t offset,
+            const uint32_t num_components) final override {
+        npartial_dot_products++;
+        return fvec_inner_product(
+                q + offset, b + i * d + offset, num_components);
     }
 
     float symmetric_dis(idx_t i, idx_t j) override {
@@ -119,12 +128,13 @@ struct FlatL2Dis : FlatCodesDistanceComputer {
     explicit FlatL2Dis(const IndexFlat& storage, const float* q = nullptr)
             : FlatCodesDistanceComputer(
                       storage.codes.data(),
-                      storage.code_size),
+                      storage.code_size,
+                      q),
               d(storage.d),
               nb(storage.ntotal),
-              q(q),
               b(storage.get_xb()),
-              ndis(0) {}
+              ndis(0),
+              npartial_dot_products(0) {}
 
     void set_query(const float* x) override {
         q = x;
@@ -161,6 +171,50 @@ struct FlatL2Dis : FlatCodesDistanceComputer {
         dis1 = dp1;
         dis2 = dp2;
         dis3 = dp3;
+    }
+
+    void partial_dot_product_batch_4(
+            const idx_t idx0,
+            const idx_t idx1,
+            const idx_t idx2,
+            const idx_t idx3,
+            float& dp0,
+            float& dp1,
+            float& dp2,
+            float& dp3,
+            const uint32_t offset,
+            const uint32_t num_components) final override {
+        npartial_dot_products += 4;
+
+        // compute first, assign next
+        const float* __restrict y0 =
+                reinterpret_cast<const float*>(codes + idx0 * code_size);
+        const float* __restrict y1 =
+                reinterpret_cast<const float*>(codes + idx1 * code_size);
+        const float* __restrict y2 =
+                reinterpret_cast<const float*>(codes + idx2 * code_size);
+        const float* __restrict y3 =
+                reinterpret_cast<const float*>(codes + idx3 * code_size);
+
+        float dp0_ = 0;
+        float dp1_ = 0;
+        float dp2_ = 0;
+        float dp3_ = 0;
+        fvec_inner_product_batch_4(
+                q + offset,
+                y0 + offset,
+                y1 + offset,
+                y2 + offset,
+                y3 + offset,
+                num_components,
+                dp0_,
+                dp1_,
+                dp2_,
+                dp3_);
+        dp0 = dp0_;
+        dp1 = dp1_;
+        dp2 = dp2_;
+        dp3 = dp3_;
     }
 };
 
--- a/faiss/index_factory.cpp
+++ b/faiss/index_factory.cpp
@@ -493,6 +493,11 @@ IndexHNSW* parse_IndexHNSW(
         return new IndexHNSWFlat(d, hnsw_M, mt);
     }
 
+    if (match("FlatPanorama([0-9]+)?")) {
+        int nlevels = mres_to_int(sm[1], 8); // default to 8 levels
+        return new IndexHNSWFlatPanorama(d, hnsw_M, nlevels, mt);
+    }
+
     if (match("PQ([0-9]+)(x[0-9]+)?(np)?")) {
         int M = std::stoi(sm[1].str());
         int nbit = mres_to_int(sm[2], 8, 1);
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

File: faiss/index_factory.cpp [hunk 1]
```diff
@@ -748,6 +748,18 @@ std::unique_ptr<Index> index_factory_sub(
         return std::unique_ptr<Index>(idmap);
     }
 
+    // handle refine Panorama
+    // TODO(aknayar): Add tests to test_factory.py
```
aknayar, line 870: Will add tests for all Panorama indexes to test_factory in a follow-up PR.

tests/test_flat_l2_panorama.py
```diff
@@ -232,14 +232,11 @@ def test_id_selector_range(self):
         D_regular, I_regular = index_regular.search(xq, k, params=params)
         D_panorama, I_panorama = index_panorama.search(xq, k, params=params)
 
-        valid_ids = I_panorama >= 0
```
aknayar, line 238: Just simplifying these tests (same exact functionality).

tests/test_ivf_flat_panorama.py
```diff
@@ -264,13 +264,11 @@ def test_id_selector_range(self):
         D_regular, I_regular = index_regular.search(xq, k, params=params)
         D_panorama, I_panorama = index_panorama.search(xq, k, params=params)
 
-        valid_ids = I_panorama >= 0
```
aknayar, line 265: Ditto.

aknayar:
@mnorris11 The failing test seems to be relevant to RaBitQ, unrelated to this PR.

faiss/Index.cpp
```diff
@@ -104,6 +104,17 @@ void Index::search_and_reconstruct(
     }
 }
 
+void Index::search_subset(
+        idx_t n,
+        const float* x,
+        idx_t k_base,
+        idx_t* base_labels,
```
mdouze: is this const?

faiss/IndexFlat.cpp
```diff
@@ -738,4 +738,112 @@ void IndexFlatPanorama::add_sa_codes(
 void IndexFlatPanorama::permute_entries(const idx_t* /* perm */) {
```
mdouze: how comes you need to redefine this?

faiss/IndexRefine.cpp
```diff
@@ -341,4 +341,48 @@ void IndexRefineFlat::search(
...
+            (params != nullptr) ? params->base_index_params : nullptr;
+
+    FAISS_THROW_IF_NOT(k_base >= k);
+
+    FAISS_THROW_IF_NOT(base_index);
+    FAISS_THROW_IF_NOT(refine_index);
+
+    FAISS_THROW_IF_NOT(k > 0);
+    FAISS_THROW_IF_NOT(is_trained);
+    idx_t* base_labels = new idx_t[n * k_base];
```
mdouze, line 377: preferably use unique_ptrs because the search can throw

mdouze (changes requested):
LGTM in general, please fix the const.

faiss/IndexFlat.cpp
aknayar: `IndexFlatCodes` defines this function, so I should override it (due to Panorama's storage layout). I implement it in #4694.

faiss/Index.cpp
aknayar: Fixed!

faiss/IndexRefine.cpp
aknayar: Updated!

aknayar:
@mdouze Thanks for the review, comments should be addressed now.

faiss/IndexRefine.cpp
```diff
@@ -371,8 +371,13 @@ void IndexRefinePanorama::search(
 
     FAISS_THROW_IF_NOT(k > 0);
     FAISS_THROW_IF_NOT(is_trained);
+
+    std::unique_ptr<idx_t[]> del1;
```
aknayar, line 375: Similar to how is done [here](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexRefine.cpp#L119-L127).

faiss/IndexFlat.cpp
```diff
@@ -684,4 +684,112 @@ void IndexFlatPanorama::add_sa_codes(
...
+    using SingleResultHandler =
+            HeapBlockResultHandler<CMax<float, int64_t>, false>::
+                    SingleResultHandler;
+    HeapBlockResultHandler<CMax<float, int64_t>, false> handler(
+            size_t(n), distances, labels, size_t(k), nullptr);
+
+    FAISS_THROW_IF_NOT(k > 0);
+    FAISS_THROW_IF_NOT(batch_size == 1);
+
+    int nt = std::min(int(n), omp_get_max_threads());
```
mnorris11: Small request: can we change this line to:
```
    [[maybe_unused]] int nt = std::min(int(n), omp_get_max_threads());
```
(internal signals are flagging it as unused, even though it is used in the macro)

aknayar: Done!

mnorris11:
Thanks for your patience @aknayar @AlSchlo , the HNSW and Refine are merged.

PR #4690, opened 2025-11-20
Revert D85902427

mnorris11:
Summary:
This diff reverts D85902427
(The context such as a Sandcastle job, Task, SEV, etc. was not provided.)

Depends on D85902427

Reviewed By: mnorris11

Differential Revision: D87552951


AlSchlo:
Hi @mnorris11 , could you share why the PR was reverted? I want to make sure I fix whatever caused it.

mnorris11:
Hi @AlSchlo it is due to backwards incompatibility in index_read and index_write. Some customers have it on the old format. So we will have to update it to check similar to how Alibek is doing it in https://github.com/facebookresearch/faiss/pull/4679. Do you have bandwidth to take a look?

(We definitely do want to re-merge it.)

I am working on tests to ensure it can get caught in the future before merge.

AlSchlo:
Yes, I can do this today.

PR #4695, opened 2025-11-24
Add initial WebAssembly bindings for FAISS

ananyasaxena:
I was working on a project that required Similarity search in the browsers, I started using [HNSW-wasm](https://www.npmjs.com/package/hnswlib-wasm) but then came across FAISS and was curious to try to out but I couldn't find wasm binding for this, AI was able to create this so sharing it back. 

Introduces a new 'wasm' directory with CMake configuration, build scripts, TypeScript bindings, demo files, and tests to enable FAISS usage via WebAssembly in browsers and Node.js. Also refactors CMakeLists to improve path handling and adds options to enable/disable OpenMP and BLAS/LAPACK for better WASM compatibility.

I've also added a readme with details and ported some tests.


faiss/CMakeLists.txt
```diff
@@ -392,12 +392,15 @@ if(FAISS_USE_LTO)
   endif()
 endif()
 
-find_package(OpenMP REQUIRED)
-target_link_libraries(faiss PRIVATE OpenMP::OpenMP_CXX)
-target_link_libraries(faiss_avx2 PRIVATE OpenMP::OpenMP_CXX)
-target_link_libraries(faiss_avx512 PRIVATE OpenMP::OpenMP_CXX)
-target_link_libraries(faiss_avx512_spr PRIVATE OpenMP::OpenMP_CXX)
-target_link_libraries(faiss_sve PRIVATE OpenMP::OpenMP_CXX)
+option(FAISS_ENABLE_OPENMP "Enable OpenMP support" ON)
```
alexanderguzhva, line 395: This and the following changes imply that FAISS is still supposed to work without `OpenMP`, and I'm not sure whether it can.
This means that changes in the `CMakeLists.txt` must include a comment and some printout to the log, for example
```cmake
# some comments on what FAISS_ENABLE_OPENMP is, which mentions that this option was created specifically for wasm, etc.
option(FAISS_ENABLE_OPENMP "Enable OpenMP support" ON)
if(FAISS_ENABLE_OPENMP)
  ...
else()
  # something like
  message(STATUS "OpenMP has been disabled, because of the FAISS_ENABLE_OPENMP flag ")
endif()
```

similar to further changes down below for `BLAS` and `LAPACK`

wasm/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname "$0")"
+mkdir -p build
+cd build
+
+emcmake cmake .. -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=OFF -DFAISS_ENABLE_C_API=OFF -DFAISS_OPT_LEVEL=generic -DFAISS_ENABLE_OPENMP=OFF -DFAISS_ENABLE_BLAS=OFF -DBUILD_TESTING=OFF -DFAISS_ENABLE_EXTRAS=OFF
```
alexanderguzhva, line 8: will `FAISS_OPT_LEVEL=generic` affect the performance? I'm just not familiar how well a wasm-compiler can SIMD-ize the code, because a lot of the FAISS low level functions are written using CPU-specific intrinsics

wasm/src/blas_stubs.cpp
```diff
@@ -0,0 +1,94 @@
+#include <algorithm>
+#include <cctype>
+#include <iostream>
+#include <vector>
+
+extern "C" {
+
+// Naive implementation of sgemm
+// C = alpha * op(A) * op(B) + beta * C
+void sgemm_(
```
alexanderguzhva, line 10: wasm does not have any standard `BLAS` facilities, correct?

mnorris11:
We had a prior contribution for web bindings at https://github.com/facebookresearch/faiss/pull/4486. Now that 2 people have created this binding, it suggests there is broader desire than initially anticipated. Do you see this providing a speedup over just a brute force search in JS? Do you have any benchmarks?

alibeklfc:
If there is an interest, feel free to provide benchmark results and we will continue reviewing

PR #4709, opened 2025-12-02
Fix typos in demos, benchs, and other directories

subhadeepkaran:
Reviewed By: limqiying

Differential Revision: D86571400


benchs/bench_ivf_flat_panorama.py
```diff
@@ -93,6 +93,7 @@ def eval_and_plot(name, plot=True):
 eval_and_plot(f"PCA{d},IVF{nlist},FlatPanorama{nlevels}")
 
 plt.title("IVF Flat Indexes on GIST1M")
+plt.title("Indices on GIST1M")
```
aknayar, line 96: I think this should be "Indexes".

PR #4721, opened 2025-12-15
Add multi-bit support for IndexRaBitQFastScan

alibeklfc:
Summary:
Extends IndexRaBitQFastScan to support multi-bit quantization (1-9 bits) while maintaining backward compatibility with existing 1-bit indexes. This unifies the API with IndexRaBitQ and enables higher precision quantization for improved recall.

Key improvements:
- Unified constructor pattern matching IndexRaBitQ with nb_bits parameter
- Backward compatible serialization using separate fourcc codes ("Irfs" for 1-bit, "Irfm" for multi-bit)
- Factory string support for multi-bit construction (e.g., "RaBitQfs4")

Differential Revision: D88495746

PR #4722, opened 2025-12-15
Add multi-bit support for IndexIVFRaBitQFastScan

alibeklfc:
Summary:
Extends IndexIVFRaBitQFastScan to support multi-bit quantization (1-9 bits), complementing the flat IndexRaBitQFastScan multi-bit implementation. This enables higher precision quantization with IVF partitioning for improved recall on large-scale datasets while maintaining FastScan performance characteristics.

The implementation uses a two-stage search strategy: fast 1-bit SIMD filtering followed by accurate ex-bit refinement with error-bound-based candidate selection. Backward compatible serialization uses separate fourcc codes ("Iwrf" for 1-bit, "Iwrm" for multi-bit), and factory string support follows the same pattern as flat multi-bit: "IVF{nlist},RaBitQfs{nb_bits}".

Differential Revision: D88869280

### DIFF
--- a/faiss/IndexIVFRaBitQFastScan.cpp
+++ b/faiss/IndexIVFRaBitQFastScan.cpp
@@ -13,6 +13,7 @@
 #include <faiss/impl/FaissAssert.h>
 #include <faiss/impl/FastScanDistancePostProcessing.h>
 #include <faiss/impl/RaBitQUtils.h>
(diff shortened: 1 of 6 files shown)


alexanderguzhva:
looks reasonable, but I'll review again it once #4721 is landed

PR #4723, opened 2025-12-15
Add RaBitQStats for tracking two-stage search filtering effectiveness

alibeklfc:
Summary:
Adds instrumentation to track the effectiveness of the two-stage filtering strategy in multi-bit RaBitQ indexes. This enables developers and users to measure how much computation is saved by the lower-bound filtering before full distance computation.

Test results demonstrate the filtering is highly effective, with all index types (IndexRaBitQ, IndexIVFRaBitQ, IndexRaBitQFastScan, IndexIVFRaBitQFastScan) achieving 98-99.5% skip rates for both 2-bit and 4-bit configurations. This means only 0.5-2% of candidates require expensive full multi-bit distance computation.

Differential Revision: D89073817

### DIFF
--- a/faiss/IndexIVFRaBitQ.cpp
+++ b/faiss/IndexIVFRaBitQ.cpp
@@ -216,6 +216,12 @@ struct RaBitInvertedListScanner : InvertedListScanner {
         // Multi-bit: Two-stage search with adaptive filtering
         size_t nup = 0;
 
+        // Stats tracking for multi-bit two-stage search
+        // n_1bit_evaluations: candidates evaluated using 1-bit lower bound
+        // n_multibit_evaluations: candidates requiring full multi-bit distance
+        size_t local_1bit_evaluations = 0;
+        size_t local_multibit_evaluations = 0;
+
         for (size_t j = 0; j < list_size; j++) {
             if (sel != nullptr) {
                 int64_t id = store_pairs ? lo_build(list_no, j) : ids[j];
@@ -225,6 +231,8 @@ struct RaBitInvertedListScanner : InvertedListScanner {
                 }
             }
 
+            local_1bit_evaluations++;
+
             // Stage 1: Compute lower bound using 1-bit codes
             float lower_bound = rabitq_dc->lower_bound_distance(codes);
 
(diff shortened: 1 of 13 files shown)


faiss/IndexIVFRaBitQ.cpp
```diff
@@ -215,6 +215,8 @@ struct RaBitInvertedListScanner : InvertedListScanner {
 
         // Multi-bit: Two-stage search with adaptive filtering
         size_t nup = 0;
+        size_t local_total = 0;
```
alexanderguzhva: comments are needed on what both variables track
I'd call them with more clear names across the code, such as `n_1bit_evaluations` and `n_multibit_evaluations` (?). The reason is that the word `refined` leads to wrong connotations, because refining is reserved to `IndexRefine`. But I leave it up to you

faiss/IndexIVFRaBitQ.cpp
```diff
@@ -256,6 +261,13 @@ struct RaBitInvertedListScanner : InvertedListScanner {
             }
             codes += code_size;
         }
+
+        // Update global stats atomically
+#pragma omp atomic
```
alexanderguzhva, line 270: `#pragma omp atomic` protects only within OMP context, I suppose?
But this is fine, because statistics is not precisely important

faiss/IndexIVFRaBitQFastScan.cpp
```diff
@@ -705,6 +715,14 @@ void IndexIVFRaBitQFastScan::IVFRaBitQHeapHandler<C>::handle(
             }
         }
     }
+
+    // Update global stats atomically (multi-bit only)
+    if (is_multibit) {
```
alexanderguzhva: why multibit only?

faiss/impl/RaBitQStats.cpp
```diff
@@ -0,0 +1,24 @@
...
+// NOLINTNEXTLINE(facebook-avoid-non-const-global-variables)
+RaBitQStats rabitq_stats;
+
+void RaBitQStats::reset() {
+    n_total = 0;
+    n_refined = 0;
+}
+
+double RaBitQStats::skip_percentage() const {
+    return n_total > 0 ? 100.0 * (n_total - n_refined) / n_total : 0.0;
```
alexanderguzhva: I'd use the following in order to avoid concurrent reads
```C++
const size_t copy_n_total = n_total;
const size_t copy_n_refined = n_refined;
return copy_n_total > 0 ? 100.0 * (copy_n_total - copy_n_refined) / copy_n_total : 0.0;
```

faiss/IndexIVFRaBitQ.cpp
alibeklfc: Added more detailed comments both in the Index files, and in the RaBitQStats. 
Renamed variables to n_1bit_evaluations and n_multibit_evaluations

faiss/IndexIVFRaBitQFastScan.cpp
alibeklfc: Initially, I added it because we count skip percentage for multi-bit only.
Removed these conditions in IndexIVFRaBitQFastScan and IndexRaBitQFastScan

faiss/impl/RaBitQStats.cpp
alibeklfc: Resolved

PR #4728, opened 2025-12-18
Panorama Refactor and Code Cleanup 

AlSchlo:
## Summary

Uses @aknayar Panorama utility functions in `IndexHNSWFlatPanorama` to standardize the usage as much as possible.
Note that we cannot use `progressive_filter_batch` as this requires contiguous storage.

Also makes the computation of cumulative sums without any extra allocation.

## Testing Plan

Existing tests & build.

### DIFF
--- a/.github/workflows/index-io-backward-compatibility.yml
+++ b/.github/workflows/index-io-backward-compatibility.yml
@@ -60,7 +60,7 @@ jobs:
           eval "$(conda shell.bash hook)"
           conda activate
           conda list
-          conda install -y -c pytorch faiss-cpu=1.13.1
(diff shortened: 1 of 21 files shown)


AlSchlo:
Will implement dot product and cosine similarity next for all indexes.

AlSchlo:
It seems that the failing CI is unrelated to my changes.

mnorris11:
> It seems that the failing CI is unrelated to my changes.

Yes it is unrelated. We intentionally made a backwards incompatible serialization change to simplify it. We will fix it by EOD tomorrow.

PR #4731, opened 2025-12-18
Remove deprecated RAFT headers

aamijar:
We can safely remove the unused raft headers which have now been deprecated as of https://github.com/rapidsai/raft/pull/2885.

Resolves https://github.com/rapidsai/raft/issues/2901


mnorris11:
Which version of RAFT/CUVS does this deprecation happen in? We are importing and enabling 25.08 internally. If this happens after that, it would break the internal version.

aamijar:
I don't believe removing these headers would break anything. The code in faiss is not using those headers at all currently, so they can be safely removed.

On the raft side, we are deprecating the headers in release 26.02.

PR #4741, opened 2025-12-26
Add early stopping to k-means clustering

alibeklfc:
Summary: K-means clustering now stops early when the objective function converges, avoiding unnecessary iterations when niter is set to a high value as a safeguard. This prevents wasted computation when convergence happens well before the maximum iteration limit.

Differential Revision: D89748242

### DIFF
--- a/faiss/Clustering.cpp
+++ b/faiss/Clustering.cpp
@@ -562,6 +562,22 @@ void Clustering::train_encoded(
 
             index.add(k, centroids.data());
             InterruptCallback::check();
+
+            // Early stopping: if objective didn't change, we've converged.
+            // Safe to access iteration_stats[size - 2] because we push_back
+            // above, so size >= i + 1, and when i > 0 we have size >= 2.
+            if (i > 0) {
+                float prev_obj =
+                        iteration_stats[iteration_stats.size() - 2].obj;
+                if (obj == prev_obj) {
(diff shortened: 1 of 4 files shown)


faiss/Clustering.cpp
```diff
@@ -562,6 +562,20 @@ void Clustering::train_encoded(
 
             index.add(k, centroids.data());
             InterruptCallback::check();
+
+            // Early stopping: if objective didn't change, we've converged
+            if (i > 0) {
+                float prev_obj =
+                        iteration_stats[iteration_stats.size() - 2].obj;
```
alexanderguzhva, line 571: this could trigger a `SIGSEGV`?

alibeklfc: In iteration 0, we add a value to iteration_stats. Size is 1. If condition is skipped, since i == 0.
In iteration 1, we add another value, size is 2. If condition is passed (i > 0), and iteration_stats accesses valid value at position 0 (2-2). 

Am I missing something? Can you please expand on the case when it could trigger SIGSEGV?

alexanderguzhva: got it. probably, it's worth adding a comment about this, that SIGSEGV is not going to happen, bcz `iteration_stats.size() >= 2` at this line

PR #4746, opened 2026-01-06
Remove packaging dependency from loader.py

mdouze:
Replace packaging.version.Version import with a simple custom Version function that returns a tuple of ints from a version string. This removes the external packaging dependency while maintaining the same version comparison functionality.


mdouze:
The packaging module is not installed automatically in many cases, and is used only for this simple function so this PR gets rid of the dependnecy.

PR #4747, opened 2026-01-06
Enable use of svs runtime conda package instead of tarball

ethanglaser:
This PR migrates from using SVS via tarball to using a conda package

Currently the package is available at the intel channel. In future releases, it will be available from conda-forge and the channel configuration can be removed


ethanglaser:
Hi, any updates on this?

PR #4753, opened 2026-01-11
add IDSelector for knn_extra_metrics()

alexanderguzhva:
Another PR that is needed to synchronize Zilliz's fork and the baseline

### DIFF
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -44,7 +44,6 @@ void IndexFlat::search(
         float_maxheap_array_t res = {size_t(n), size_t(k), labels, distances};
         knn_L2sqr(x, get_xb(), d, n, ntotal, &res, nullptr, sel);
     } else {
-        FAISS_THROW_IF_NOT(!sel); // TODO implement with selector
         knn_extra_metrics(
                 x,
                 get_xb(),
@@ -55,7 +54,8 @@ void IndexFlat::search(
                 metric_arg,
                 k,
                 distances,
-                labels);
+                labels,
+                sel);
     }
 }
 
--- a/faiss/utils/extra_distances.cpp
+++ b/faiss/utils/extra_distances.cpp
@@ -15,6 +15,7 @@
 
 #include <faiss/impl/AuxIndexStructures.h>
 #include <faiss/impl/DistanceComputer.h>
+#include <faiss/impl/IDSelector.h>
 #include <faiss/utils/utils.h>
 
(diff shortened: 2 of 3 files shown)


faiss/utils/extra_distances.cpp
```diff
@@ -82,10 +84,12 @@ struct Run_knn_extra_metrics {
                 // maxheap_heapify(k, simi, idxi);
                 heap_heapify<C>(k, simi, idxi);
                 for (j = 0; j < ny; j++) {
-                    float disij = vd(x_i, y_j);
+                    if (!sel || sel->is_member(j)) {
```
subhadeepkaran, line 87: noobq: how expensive is this check?

faiss/IndexFlat.cpp
```diff
@@ -44,7 +44,6 @@ void IndexFlat::search(
         float_maxheap_array_t res = {size_t(n), size_t(k), labels, distances};
         knn_L2sqr(x, get_xb(), d, n, ntotal, &res, nullptr, sel);
     } else {
-        FAISS_THROW_IF_NOT(!sel); // TODO implement with selector
```
subhadeepkaran, line 47: can you also leave some brief benefits for the change? it helps the reviewer who might not be completely aware of the downsides with the current setup

subhadeepkaran: overall the PR lgtm, just curious about the benefits and query about the impact of `is_member` check

faiss/utils/extra_distances.cpp
alexanderguzhva: from a technical point of view, it should cost nothing if `sel` is `false`. The reason is that the value of `sel` remains unchanged, so it won't hurt a CPU branch predictor. If `sel` is `true`, then the cost of doing `sel->is_member(j)` call is unavoidable.

faiss/IndexFlat.cpp
alexanderguzhva: I'm not sure I follow your question. This particular PR is an implementation of an evident `TODO` and it seems that using `sel` is a de-facto standard practice in the whole `Faiss`. Other pieces of the `Faiss` code for other distance types & indices also perform `is_member` check in similar situations

subhadeepkaran: fair enough, if its standard practice throughout the `faiss`. I did see the `TODO` but the motivation for that `TODO` wasnt clear to me. approving the diff. will rebase and initiate the landing

PR #4755, opened 2026-01-13
Fix compilation on macOS ARM64: Use faiss::idx_t instead of long test_hamming

dascoder:
There's a build failure triggered by vector types in test_hamming.cpp. On  macOS ARM64 long and int64_t are distinct types. 
Here's a fix.


alibeklfc:
Hi, @dascoder. Thank you for your contribution! I was able to reproduce the error and will merge this PR after all tests pass.

dascoder:
> Hi, @dascoder. Thank you for your contribution! I was able to reproduce the error and will merge this PR after all tests pass.

Awesome Thanks!

PR #4757, opened 2026-01-14
Change default qb from 0 to 4 in RaBitQ indexes

alibeklfc:
Summary: Enable query quantization by default for RaBitQ indexes instead of using raw fp32 values. A qb value of 4 provides a good balance between search performance and accuracy.

Differential Revision: D90704860

### DIFF
--- a/faiss/IndexIVFRaBitQ.h
+++ b/faiss/IndexIVFRaBitQ.h
@@ -19,7 +19,7 @@
 namespace faiss {
 
 struct IVFRaBitQSearchParameters : IVFSearchParameters {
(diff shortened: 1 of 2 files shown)


alexanderguzhva:
@alibeklfc any benchmarks on this topic? I wonder whether this behavior is dataset-specific.

alibeklfc:
@alexanderguzhva We have decided to change default qb value from 0 to 4, because 0 is much slower. We have benchmarked on SyntheticDatase.  Flat was faster than rabitq with qb = 0. We are not changing anything in the implementation, only the default value of qb

PR #4364, opened 2025-05-26
BIVF selector

metonymic-smokey:

PR #4761, opened 2026-01-16
support SearchParameters for IndexBinary

### LINKED ISSUE #3503: Opening up SearchParameters for IndexBinaryHNSW and IndexBinaryIVF
# Summary

IndexBinaryHNSW and IndexBinaryIVF is blocking SearchParameters. Can we open it up? Will there be any issue that I am not aware of?
https://github.com/facebookresearch/faiss/blob/main/faiss/IndexBinaryHNSW.cpp#L192

gilbsgilbs:
- Add support ParameterSpace for IndexBinaryHNSW and IndexBinaryIVF
- Add support SearchParameters in search() method for IndexBinaryHNSW and IndexBinaryIVF
- Add support for passing parameters to IndexBinaryHNSW and IndexBinaryIVF from CAPI

fixes #3503

### DIFF
--- a/faiss/IndexBinaryIVF.cpp
+++ b/faiss/IndexBinaryIVF.cpp
@@ -14,6 +14,7 @@
 #include <cstdio>
 
 #include <algorithm>
+#include <limits>
 #include <memory>
 
 #include <faiss/IndexFlat.h>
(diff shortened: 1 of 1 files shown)


gilbsgilbs:
Need to safeguard against selectors (not implemented in this PR, see https://github.com/facebookresearch/faiss/pull/4364) and figure out if we need to do something about flat binary indexes.

mdouze:
Thanks for the PR! 

Would you mind adding a few tests to validate that: 
* setting the searchparameters works 
* the parameterspace  

The tests would preferably be in python

The tests would be similar to 
https://github.com/facebookresearch/faiss/blob/933ebb778125f18d81d7353ad87ad002932fcc38/tests/test_autotune.py

gilbsgilbs:
Thanks for the review @mdouze . Sorry for the delay, but after adding the tests I noticed that `max_codes` test was failing and I needed to find some time to investigate in my spare time. Turns out it's just a minor behavioral difference between how IndexBinaryIVF handles `max_codes` parameter compared to IndexIVF. Technically, I think this is a bug so I reflected the implementation from IndexIVF but I guess it's up for debate. If that's considered a breaking change, I can also adapt the test instead and revert cbd62776d128ade8d669ef595265d5bdaca8bed6 .

PR #4762, opened 2026-01-19
generic result handlers for most indexes

mdouze:
Summary:
Some search functions require special handling on the search resutls. For example collecting resutls from diverse groups.

In order to support this, we re-purpose the ResultHandler object to offer an arbitrary callback whenever a new vector should be taken in consideration.

Many indexes did already have such a callback, but in non-uniform formats. This diff consolidates these diverse implementations. This is especially the case for the InvertedListScanner that is now implemented on top of that.

Most indexes now get a `search1` that searches one query vector and takes a ResultCollector as input. This is the object that can be used a a callback to do specific handling on the results.

The current impleentatoin of search1 may not be perfectly optimized, especially for IndexFlatCodes indexes. Another todo is to support BlockResultHandlers to allow multiple queries at a time.

Differential Revision: D91003349

### DIFF
--- a/faiss/Index.h
+++ b/faiss/Index.h
@@ -14,7 +14,6 @@
 #include <faiss/impl/FaissAssert.h>
 
 #include <cstdio>
-#include <sstream>
 
 #define FAISS_VERSION_MAJOR 1
 #define FAISS_VERSION_MINOR 13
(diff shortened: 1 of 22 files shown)


faiss/Index.h
```diff
@@ -216,6 +218,12 @@ struct Index {
         }
     }
 
+    /** search one vector with a custom result handler */
+    virtual void search1(
```
alexanderguzhva, line 222: is it worth unifying this with a regular `void search()` by passing `handler` into `SearchParameters`?

PR #4764, opened 2026-01-20
Add retrieval benchmark demo for vector search evaluation

ashiksharonm:
## Summary
Added a new demo script `demos/retrieval_benchmark.py` that benchmarks three common FAISS index types:
- `IndexFlatL2` (Exact Search)
- `IndexIVFFlat` (Inverted File)
- `IndexHNSWFlat` (Graph-based)

## Why needed
This script provides a quick way for users to compare trade-offs between accuracy (recall) and speed across different index types using synthetic data.

## Sample Output
The script generates a `metrics.json` file.

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -26,3 +26,15 @@ done in parallel on several machines
 (needs not to fit in RAM)
 
 6: load and test the index
+
+
+### Retrieval Benchmark Example
+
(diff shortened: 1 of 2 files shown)


mdouze:
Thanks for your contribution. 
This script does not add value to the current benchmarks in place. Especially, uniform random data is not interesting for benchmarks.

PR #4770, opened 2026-01-26
Add search_and_reconstruct support to IndexBinaryHash

tracywong117:
## Summary:
This PR implements `search_and_reconstruct()` for `IndexBinaryHash`, enabling single-pass search and retrieval of binary vectors. This avoids the overhead of a second lookup pass (reconstruction) after search, improving efficiency.
### 1. New Function: `IndexBinaryHash::search_and_reconstruct`
-   Implemented `search_and_reconstruct` in `IndexBinaryHash.cpp`.
-   This method returns both the search results (distances, labels) and the **reconstructed binary vectors** in a single call.
-   **Implementation Detail:** It uses a specialized `HeapTriple` (Value, ID, Pointer) to store the pointer to the vector code *during* the search phase. This allows immediate reconstruction without traversing the inverted lists a second time.
### 2. Python Support
-   Exposed `search_and_reconstruct` in the Python `IndexBinary` wrapper (`class_wrappers.py`).
## Test
1. Verified that `search_and_reconstruct()` produces identical results to running `search()` followed by `reconstruct()`, but is performed in a single pass.
2. Tested with `n=5,000,000,000` vectors to ensure stability and correctness under load.


pankajsingh88:
can you provide some metric on how much perf improvement is observed? Usually k value is low, and for that reconstruct overhead should be insignificant compared to actual search. using reconstruct_n (batched version) can be another option to speed up the reconstruction.

tracywong117:
I performed a Python benchmark comparing `search_and_reconstruct` vs `search` + `reconstruct` vs `reconstruct_n` for `IndexBinaryHash`.
- **Database**: 10,000,000 binary vectors (d=128), 16-bit hash.
- **Speed**:

Method | k=1 (Total Time) | k=10 (Total Time)
-- | -- | -- 
`search_and_reconstruct` | 0.0001s | 0.0001s 
`search` + `reconstruct` | 0.88s | 8.44s 
`reconstruct_n` | 1.00s | 10.26s 

- `search_and_reconstruct` outperforms the other two. 
- Note that `search + reconstruct_n` is not a valid alternative since `reconstruct_n` can only retrieve a contiguous range of vectors rather than individual indices."

```py
import time
import numpy as np
import faiss

def run_benchmark_case(k):
    d = 128
    nb = 10000000  # database size
    nq = 100     # number of queries
    n_bits = 16  # hash bits

    print(f"\n========================================================")
    print(f"Running benchmark with k={k}")
    print(f"Parameters: d={d}, nb={nb}, nq={nq}, n_bits={n_bits}")

    # Initialize index
    index = faiss.IndexBinaryHash(d, n_bits)
    
    # Generate data
    np.random.seed(42)
    xb = np.random.randint(0, 256, size=(nb, d // 8)).astype('uint8')
    xq = np.random.randint(0, 256, size=(nq, d // 8)).astype('uint8')
    
    print("Adding vectors to index...")
    add_start = time.time()
    index.add(xb)
    print(f"Add time: {time.time() - add_start:.4f}s")
    
    # Warmup
    index.search(xq[:10], k)
  
    print("\n--- Benchmarking search_and_reconstruct ---")
    start_time = time.time()
    D_sr, I_sr, R_sr = index.search_and_reconstruct(xq, k)
    sr_time = time.time() - start_time
    print(f"Time: {sr_time:.4f} seconds")
    
    # Benchmark search + reconstruct loop
    print("\n--- Benchmarking search() + reconstruct() loop ---")
    start_time = time.time()
    D_s, I_s = index.search(xq, k)
    search_only_time = time.time() - start_time
    print(f"Search only time: {search_only_time:.4f} seconds")
    
    reconstruct_start = time.time()
    # Initialize with 0xff to match not-found behavior
    reconstructed_vectors = np.full((nq, k, d // 8), 255, dtype='uint8')
    
    # Standard Python loop reconstruction
    for i in range(nq):
        for j in range(k):
             idx = I_s[i, j]
             if idx >= 0:
                 reconstructed_vectors[i, j] = index.reconstruct(int(idx))
    
    reconstruct_loop_time = time.time() - reconstruct_start
    total_baseline_time = search_only_time + reconstruct_loop_time
    print(f"Reconstruct loop time: {reconstruct_loop_time:.4f} seconds")
    print(f"Total baseline time: {total_baseline_time:.4f} seconds")

    # Benchmark reconstruct_n (contiguous)
    print("\n--- Benchmarking reconstruct_n (contiguous range of size nq*k) ---")
    rn_start = time.time()
    # Reconstruct the same number of vectors as total search results
    # This is a throughput test, as reconstruct_n cannot target specific search indices
    index.reconstruct_n(0, nq * k)
    rn_time = time.time() - rn_start
    print(f"Time for {nq * k} contiguous vectors: {rn_time:.4f} seconds")
    
    if sr_time > 0:
        print(f"\nSpeedup vs Loop: {total_baseline_time / sr_time:.2f}x")
    else:
        print(f"\nSpeedup vs Loop: infinite (search_and_reconstruct took 0.0s)")
        
    if sr_time > 0:
        print(f"Speedup vs reconstruct_n: {rn_time / sr_time:.2f}x")

    # Verification
    print("\nVerifying results match...")
    assert np.all(D_sr == D_s), "Distances mismatch"
    assert np.all(I_sr == I_s), "Indices mismatch"
    
    if np.all(R_sr == reconstructed_vectors):
        print("Reconstruction matches!")
    else:
        print("Reconstruction MISMATCH!")

def run_benchmark():
    run_benchmark_case(k=1)
    run_benchmark_case(k=10)

if __name__ == "__main__":
    run_benchmark()

```

mnorris11:
Sorry for the big delay on this. We are working through the backlog.

I am hesitant about this one because it duplicates the existing Heap structure. 

Did you see a use case for this?

PR #4773, opened 2026-01-27
feat(svs): LeanVec OOD support

ahuber21:
This PR will add LeanVec OOD (out-of-distribution) support.

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -84,6 +84,7 @@ set(FAISS_SRC
   clone_index.cpp
   index_factory.cpp
   impl/AuxIndexStructures.cpp
+  impl/VisitedTable.cpp
   impl/ClusteringInitialization.cpp
   impl/CodePacker.cpp
   impl/IDSelector.cpp
@@ -207,6 +208,7 @@ set(FAISS_HEADERS
   index_io.h
   impl/AdditiveQuantizer.h
   impl/AuxIndexStructures.h
+  impl/VisitedTable.h
   impl/ClusteringInitialization.h
   impl/CodePacker.h
   impl/IDSelector.h
--- a/benchs/bench_visited_table.cpp
+++ b/benchs/bench_visited_table.cpp
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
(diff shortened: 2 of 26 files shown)


faiss/CMakeLists.txt
```diff
@@ -366,7 +366,7 @@ if(NOT WIN32)
 endif()
 
 if(FAISS_ENABLE_SVS)
-  find_package(svs_runtime REQUIRED)
+  find_package(svs_runtime 0.2.0 REQUIRED)
```
ethanglaser, line 455:
```suggestion
  find_package(svs_runtime) # 0.2.0 REQUIRED)
```

mdouze:
There seems to be something wrong in the Python wrapper. Could you please fix it?

faiss/python/class_wrappers.py
```diff
@@ -332,21 +332,44 @@ def replacement_assign(self, x, k, labels=None):
         self.assign_c(n, swig_ptr(x), swig_ptr(labels), k)
         return labels
 
-    def replacement_train(self, x, numeric_type = faiss.Float32):
+    def replacement_train(self, x, *, numeric_type=faiss.Float32, xq_train=None):
         """Trains the index on a representative set of vectors.
```
mdouze, line 336: This is a good interface for python: any non-trivial parameter must be passed by keyword.

faiss/python/class_wrappers.py
```diff
@@ -332,21 +332,44 @@ def replacement_assign(self, x, k, labels=None):
...
+            n_train_q, d_train = xq_train.shape
+            assert d_train == self.d
+            train_q = swig_ptr(
+                np.ascontiguousarray(xq_train, dtype=_numeric_to_str(numeric_type))
+            )
+
+        # Dispatch to train_c / train_ex
         if numeric_type == faiss.Float32:
-            self.train_c(n, swig_ptr(x))
+            self.train_c(n, swig_ptr(x), n_train_q, train_q)
```
mnorris11, line 372: train_q can still be None here which causes some tests to break.

I will just update this line to:
```
            if train_q is not None:
                self.train_c(
                    n, swig_ptr(x), n_train_q, train_q
                )
            else:
                self.train_c(n, swig_ptr(x))
```
if this change is OK with you, I will just merge it with this change since it passed the signals already. Then we don't need to wait a few hours for CI to finish again :)

mihaic: Yes, please apply the change and merge.

PR #4552, opened 2025-08-26
SIMDConfig object + SIMD detections

subhadeepkaran:
Summary:
Implementation of:
- `SIMDLevel` that can be used as a template parameter to differentiate implementations 
- `SIMDConfig` object that defines what is the enabled SIMDLevel

in the TARGETS all SIMD code is disabled by default.

Reviewed By: mnorris11

Differential Revision: D72937710

PR #4780, opened 2026-01-29
SIMDConfig object + SIMD detections (#4552)

algoriddle:
Summary:

Implementation of:
- `SIMDLevel` that can be used as a template parameter to differentiate implementations 
- `SIMDConfig` object that defines what is the enabled SIMDLevel

in the TARGETS all SIMD code is disabled by default.

Reviewed By: mnorris11

Differential Revision: D72937710

### DIFF
--- a/faiss/utils/simd_levels.cpp
+++ b/faiss/utils/simd_levels.cpp
@@ -1,0 +1,276 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <faiss/utils/simd_levels.h>
+
+#include <cstdlib>
+
+#include <faiss/impl/FaissAssert.h>
+
+namespace faiss {
+
+// Static member definitions - used in both DD and static modes
+SIMDLevel SIMDConfig::level = SIMDLevel::NONE;
+
+// Bitmask of supported SIMD levels (1 << SIMDLevel)
+uint64_t SIMDConfig::supported_simd_levels = 0;
+
+#ifdef FAISS_ENABLE_DD
+
+// =============================================================================
+// Dynamic Dispatch (DD) mode implementation
+// =============================================================================
+
+// Static initializer to run constructor at load time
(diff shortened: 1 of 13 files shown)


faiss/utils/simd_levels.h
```diff
@@ -0,0 +1,137 @@
...
+#define DISPATCH_SIMDLevel(f, ...)                         \
+    switch (SIMDConfig::level) {                           \
+        case SIMDLevel::NONE:                              \
+            return f<SIMDLevel::NONE>(__VA_ARGS__);        \
+            DISPATCH_SIMDLevel_AVX2(f, __VA_ARGS__);       \
+            DISPATCH_SIMDLevel_AVX512(f, __VA_ARGS__);     \
+            DISPATCH_SIMDLevel_AVX512_SPR(f, __VA_ARGS__); \
+            DISPATCH_SIMDLevel_ARM_NEON(f, __VA_ARGS__);   \
+        default:                                           \
+            FAISS_ASSERT(!"Invalid SIMD level");           \
```
alexanderguzhva: what does `!` mean here?

faiss/utils/simd_levels.cpp
```diff
@@ -0,0 +1,217 @@
...
+    const char* env_var = faiss_simd_level_env ? *faiss_simd_level_env
+                                               : getenv("FAISS_SIMD_LEVEL");
+
+    // check environment variable for SIMD level is explicitly set
+    if (!env_var) {
+        level = auto_detect_simd_level();
+    } else {
+        auto matched_level = to_simd_level(env_var);
+        if (matched_level.has_value()) {
+            set_level(matched_level.value());
```
alexanderguzhva: is it worth adding a debug message here about a selected SIMD level as well, just to confirm that a user-defined simd level got accepted?

faiss/utils/simd_levels.h
algoriddle: good catch, thanks, AI slop

faiss/utils/simd_levels.cpp
algoriddle: to_simd_level() now throws an exception

PR #4781, opened 2026-01-29
dynamic dispatch distances_simd (#4553)

algoriddle:
Summary:

`fvec_madd` is the first function to test dispatching to AVX and AVX512

distances_simd.cpp is split into specialized files distances_avx2.cpp distances_avx512.cpp that are compiled with appropriate flags.

Reviewed By: mnorris11

Differential Revision: D72937708


alexanderguzhva:
same comments as for #4780

PR #4784, opened 2026-01-29
[Faiss] inline scanning code for fast distance computers 

mdouze:
The new scanning code that accepts arbitrary ResultHandlers incurs virtual function calls to distance_to_code. To avoid this, this PR introduces a new header that gives sufficient visibility to the scanning code that the distance_to_code can be inlined.

### DIFF
--- a/benchs/bench_result_handler_overhead.cpp
+++ b/benchs/bench_result_handler_overhead.cpp
@@ -16,8 +16,8 @@
 #include <cmath>
 #include <iomanip>
 #include <iostream>
-#include <vector>
 #include <map>
+#include <vector>
 
 namespace faiss {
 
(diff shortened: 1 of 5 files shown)


benchs/bench_result_handler_overhead.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+std::pair<double, double> run_search(
+        IndexData& data,
+        int d,
+        int k,
+        int nprobe,
+        const char* factory_string) {
+    ParameterSpace().set_index_parameter(data.index.get(), "nprobe", nprobe);
+
+    omp_set_num_threads(1);
```
trang-nm-nguyen, line 53: I am curious what is the motivation for omp_set_num_threads = 1 here?

faiss/IndexIVFFlat.cpp
```diff
@@ -185,11 +196,7 @@ struct Run_get_InvertedListScanner {
             const IndexIVFFlat* ivf,
             bool store_pairs,
             const IDSelector* sel) {
```
trang-nm-nguyen, line 198: I am curious if we still need const IDSelector* sel given we don't need to use it anymore in this function?

alibeklfc:
This is a copy of committed #4785 and #4778. Closing this PR

PR #4787, opened 2026-01-31
Add `IndexFlatIPPanorama`

aknayar:
This PR builds on #4732, adding `IndexFlatIPPanorama` and updating unit tests.

Hide whitespace to see cleaner diffs.

Note: Renamed `test_flat_l2_panorama.py` -> `test_flat_panorama.py`.

### DIFF
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -594,7 +594,7 @@ inline auto dispatch_metric_compare(MetricType metric, Fn&& fn) {
     }
 }
 
-template <bool use_radius, typename BlockHandler, typename C>
+template <bool use_radius, typename C, typename BlockHandler>
 inline void flat_pano_search_core(
         const IndexFlatPanorama& index,
         BlockHandler& handler,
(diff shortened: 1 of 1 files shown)


tests/test_flat_panorama.py
```diff
@@ -0,0 +1,695 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
aknayar, line 1: Just renamed `test_flat_l2_panorama` and added iteration over metric types.

faiss/index_factory.cpp
```diff
@@ -708,6 +708,18 @@ Index* parse_other_indexes(
         }
     }
 
+    // IndexFlatIPPanorama
+    if (match("FlatIPPanorama([0-9]+)(_[0-9]+)?")) {
```
aknayar, line 712: Ideally I would replace this and `FlatL2Panorama` with just `FlatPanorama` (similar to how is done for [IndexFlat](https://github.com/aknayar/faiss-pano/blob/60ea56c5d64c5068dc05a110d173b2445adf0cf7/faiss/index_factory.cpp#L694-L697)), but I think that would break backward compatibility for those using `index_factory("FlatL2Panorama...")`.

PR #4788, opened 2026-02-01
Upgrade cuVS Version to 26.02

tarang-jain:


trang-nm-nguyen:
@tarang-jain looks like there is a failed test on GPU with cuVS if you can help take a look into

mnorris11:
Hoping to merge today, waiting on 1 more review

PR #4801, opened 2026-02-11
Enable SVS IVF Index in FAISS

ibhati:
This PR introduces SVS IVF (Inverted File) index support in FAISS, complementing the existing SVS Vamana graph‑based index. The implementation integrates Intel's Scalable Vector Search (SVS) runtime to provide a high‑performance, partition‑based ANN index with strong scalability on modern hardware.

## Key Advantages of SVS IVF

- **AMX‑accelerated builds** for significantly faster index training on supported Intel platforms
- **Hierarchical clustering** (default ON) enabling much faster k‑means on large datasets
- **LVQ and LeanVec compression** applied at search time to boost throughput while preserving recall
- **Intra-query parallelism** to reduce latency for small‑batch or single‑query workloads
- **Static index mode** for read‑only workloads with optimized memory layout and search performance

## New Factory Strings

| Factory String | IndexClass | Storage |
|---|---|---|
| `SVSIVF{nlist}` | IndexSVSIVF | FP32 (default) |
| `SVSIVF{nlist},FP16` | IndexSVSIVF | Float16 |
| `SVSIVF{nlist},SQI8` | IndexSVSIVF | Scalar quantized int8 |
| `SVSIVF{nlist},LVQ4x0` | IndexSVSIVFLVQ | LVQ primary=4, no residual |
| `SVSIVF{nlist},LVQ4x4` | IndexSVSIVFLVQ | LVQ primary=4, residual=4 |
| `SVSIVF{nlist},LVQ4x8` | IndexSVSIVFLVQ | LVQ primary=4, residual=8 |
| `SVSIVF{nlist},LVQ8x0` | IndexSVSIVFLVQ | LVQ primary=8, no residual |
| `SVSIVF{nlist},LeanVec4x4_{dims}` | IndexSVSIVFLeanVec | LeanVec 4x4 |
| `SVSIVF{nlist},LeanVec4x8_{dims}` | IndexSVSIVFLeanVec | LeanVec 4x8 |
| `SVSIVF{nlist},LeanVec8x8_{dims}` | IndexSVSIVFLeanVec | LeanVec 8x8 |

## Static vs Dynamic IVF Index

SVS IVF supports two modes controlled by the `is_static` constructor parameter:

### Dynamic IVF (default, `is_static = false`)
- Supports `train()`, `add()`, `remove_ids()`, and `search()`
- Vectors can be added or removed after index creation
- Suitable for workloads where the dataset changes over time

### Static IVF (`is_static = true`)
- **All data must be provided during `train()`** — the index is built and populated in a single step
- `add()` and `remove_ids()` will throw an exception after creation
- Optimized for **read‑only workloads** where the dataset is fixed at build time
- Uses `svs_runtime::IVFIndex` (static) instead of `svs_runtime::DynamicIVFIndex`
- Serialization and deserialization are fully supported

**Usage Examples(C++):**

```cpp
// Dynamic IVF with intra-query parallelism
faiss::IndexSVSIVF index(d, nlist);

// intra_query_threads must be set BEFORE train() — changes after
// train() are not propagated to the SVS backend (current SVS runtime
// API limitation, will be fixed in a future SVS version).
index.intra_query_threads = 4;
index.n_probes = 20;

index.train(n, train_data);
index.add(n, base_data);
index.search(nq, queries, k, distances, labels);

// num_threads CAN be changed between searches (synced at search time).
index.num_threads = 8;
index.search(nq, queries, k, distances, labels);

// Static IVF (read-only, all data at build time)
faiss::IndexSVSIVF index(d, nlist, faiss::METRIC_L2,
                         faiss::SVSStorageKind::SVS_FP32, /*is_static=*/true);
index.intra_query_threads = 4;  // set before train()
index.train(n, data);           // builds and populates the index
index.search(nq, queries, k, distances, labels);
// index.add(...)        // throws — not supported on static index
// index.remove_ids(...) // throws — not supported on static index

### CODE REVIEW

File: tests/test_svs.cpp [hunk 1]
```diff
@@ -489,3 +492,324 @@ TEST_F(SVSNoLL, LVQAndLeanVecThrowWhenNotEnabled) {
...
+
+    faiss::SearchParameters params;
+    params.sel = &selector;
+
+    std::vector<float> distances(nq * k);
+    std::vector<faiss::idx_t> labels(nq * k);
+
+    // IVF search does not support IDSelector via the SVS runtime,
+    // so we just test that search works without crashing
+    ASSERT_NO_THROW(index.search(nq, xq, k, distances.data(), labels.data()));
```
mnorris11, line 692: I saw there is `auto sp = make_ivf_search_parameters(*this, params);` in the svsIVF search, and Vamana has a `make_faiss_id_filter`.  Is this like a pending item in SVS library for SVSIVF to support id filter? It is OK either way, just trying to understand the current state

faiss/svs/IndexSVSIVFLVQ.h
```diff
@@ -0,0 +1,43 @@
...
+namespace faiss {
+
+struct IndexSVSIVFLVQ : IndexSVSIVF {
+    IndexSVSIVFLVQ();
+
+    IndexSVSIVFLVQ(
+            idx_t d,
+            size_t nlist,
+            MetricType metric = METRIC_L2,
+            SVSStorageKind storage = SVSStorageKind::SVS_LVQ4x4,
```
mnorris11: `IndexSVSIVFLVQ::IndexSVSIVFLVQ()` defaults to SVS_LVQ4x0, but this defaults to SVS_LVQ4x4. Should we make them the same?

faiss/svs/IndexSVSIVF.cpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+    FAISS_THROW_IF_NOT(k > 0);
+    FAISS_THROW_IF_NOT(is_trained);
+
+    // Sync thread settings to the SVS backend before searching
+    {
+        size_t current_threads = 0;
+        impl->get_num_threads(&current_threads);
+        if (num_threads > 0 && current_threads != num_threads) {
+            auto st = impl->set_num_threads(num_threads);
```
mnorris11, line 184: so if someone does 
1. train()  (create the impl)
2. sets `intra_query_threads` on their own
3. search()

this intra_query_threads is not respected. It's probably OK, but if we can add a usage example in the PR and/or unit test, that would be great. Edit: or in wiki.

faiss/svs/IndexSVSIVF.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+void IndexSVSIVF::train(idx_t n, const float* x) {
+    FAISS_THROW_IF_MSG(impl, "Index already trained or contains data.");
+    create_impl(n, x);
+    is_trained = true;
+}
+
+void IndexSVSIVF::add(idx_t n, const float* x) {
+    FAISS_THROW_IF_NOT_MSG(
+            !is_static,
```
mnorris11: Let's not double negative: we can do `FAISS_THROW_IF_MSG(is_static`

mnorris11:
Maybe in the PR description we can add a description of the "static" index?

Alternatively we can do wiki updates

faiss/svs/IndexSVSIVF.cpp
```diff
@@ -0,0 +1,320 @@
...
+            ? svs_runtime::IVFIndex::check_storage_kind(svs_storage)
+            : svs_runtime::DynamicIVFIndex::check_storage_kind(svs_storage);
+    if (!status.ok()) {
+        FAISS_THROW_MSG(status.message());
+    }
+}
+
+bool IndexSVSIVF::is_lvq_leanvec_enabled() {
+    auto lvq = to_svs_storage_kind(SVS_LVQ4x0);
+    auto status = svs_runtime::DynamicIVFIndex::check_storage_kind(lvq);
```
mnorris11: Will there ever be a difference in getting status for `svs_runtime::IVFIndex::check_storage_kind` and this `svs_runtime::DynamicIVFIndex::check_storage_kind`? like if this is static, then is_lvq_leanvec_enabled should return false?

tests/test_svs.cpp
ibhati: Yes, this is a pending item in the SVS runtime. The current `svs_runtime::IVFIndex::search()` API doesn't accept an `IDFilter` parameter (unlike `VamanaIndex::search()` which does). We do use `FaissIDFilter` for `remove_ids()` on the dynamic IVF index since the runtime exposes `remove_selected(IDFilter&)` there. Search-time ID filtering for IVF will be added in a future SVS runtime release, after which we can wire it up here using the same `make_faiss_id_filter()` pattern as Vamana. I've also clarified the test comment in `IVFSearchWithIDSelector` to make this explicit (commit c8e745cc5).

faiss/svs/IndexSVSIVFLVQ.h
ibhati: Good catch — fixed in commit b8ec39bfb. Both the default and parameterized constructors now use `SVS_LVQ4x0` consistently for both `IndexSVSIVFLVQ` and `IndexSVSVamanaLVQ`.

faiss/svs/IndexSVSIVF.cpp
ibhati: You're right — once `train()` creates the impl, changing `intra_query_threads` is silently ignored because the SVS runtime API doesn't currently expose a setter for it (it only takes it as a build/load parameter). `num_threads` works because the runtime exposes `set_num_threads()` which we sync at search time.

I've documented this in three places:
 - Doc comment on the field in IndexSVSIVF.h (commit b8ec39bfb)
 - New test `SVS.IVFIntraQueryThreadsSetBeforeTrain` in test_svs.cpp (commit e9d59b037) showing the supported pattern (set before `train()`)
 - Updated PR description with usage examples and a note

 A `set_intra_query_threads()` API will be added in a future SVS runtime version, after which we can mirror the `num_threads` sync logic.

faiss/svs/IndexSVSIVF.cpp
ibhati: Done in commit b8ec39bfb. Cleaned up in `add()`, `remove_ids()`, and `dynamic_impl()` to use `FAISS_THROW_IF_MSG(is_static, ...)`.

ibhati:
> Maybe in the PR description we can add a description of the "static" index?
> 
> Alternatively we can do wiki updates

Added a "Static vs Dynamic IVF Index" section to the PR description, plus a usage example contrasting the two modes. Let me know if you'd prefer this in the wiki instead and I'll move/duplicate it there.

faiss/svs/IndexSVSIVF.cpp
ibhati: Yes, in principle they could differ — they're separate runtime entry points and could be feature-gated independently in a future SVS build. Updated in commit b8ec39bfb to check both `svs_runtime::IVFIndex::check_storage_kind` (static) and `svs_runtime::DynamicIVFIndex::check_storage_kind` (dynamic) for both LVQ and LeanVec storage kinds, so the helper now correctly reports availability for both index variants.

tests/test_svs.cpp
```diff
@@ -489,3 +492,324 @@ TEST_F(SVSNoLL, LVQAndLeanVecThrowWhenNotEnabled) {
...
+    if constexpr (std::is_same_v<std::decay_t<T>, faiss::IndexSVSIVFLeanVec>) {
+        auto* leanvec_loaded = dynamic_cast<faiss::IndexSVSIVFLeanVec*>(loaded);
+        ASSERT_NE(leanvec_loaded, nullptr);
+        EXPECT_EQ(leanvec_loaded->leanvec_d, index.leanvec_d);
+        EXPECT_NE(leanvec_loaded->training_data, nullptr);
+    }
+
+    // Verify search works on loaded index
+    ASSERT_NO_THROW(
+            index.search(nq, xb.data(), k, distances.data(), labels.data()));
```
mnorris11: I believe this should be `loaded.search`. After the change, this crashes for me internally with BUCK build system. Does it for you with cmake?

I tried to fix it by patching the internal version of SVS (eventually I will just submit a PR to SVS so its simpler), but I was not successful.  It was only fixable by patching the wrapper here in Faiss:
IndexSVSIVF::deserialize_impl needed this at the end:
```
    size_t current_threads = 0;
    impl->get_num_threads(&current_threads);
    auto st = impl->set_num_threads(current_threads);
    if (!st.ok()) {
        FAISS_THROW_MSG(st.message());
    }
```
I will ping you to discuss this part

ibhati: Thanks for the catch! I applied the index.search → loaded->search fix and reran the static IVF tests against the public conda-forge libsvs-runtime=0.3.0 (the same version CI uses). All SVS tests cases pass cleanly without the workaround, so the crash doesn't reproduce on the OSS cmake build — it appears specific to BUCK.
To unblock this PR, I'm happy to land the workaround you suggested in IndexSVSIVF::deserialize_impl (re-applying set_num_threads(get_num_threads()) after load). On the SVS side, I'll investigate the underlying issue and roll a proper fix into the next runtime release (0.4.0), along with the other follow-ups from your review comments. Once 0.4.0 ships, we can drop the wrapper-side workaround.

PR #4908, opened 2026-03-11
Fix: Add class name and dtype validation in copy_array_to_AlignedTable

LincolnBurrows2017:
This PR fixes a TODO in the `copy_array_to_AlignedTable` function to add proper class name checking.

Previously, the function would fail with a confusing error if the wrong type of AlignedTable was passed. Now it provides clear type checking and error messages.

### Changes:
- Added class name validation to ensure the input is an AlignedTable
- Added dtype validation to ensure array matches the expected type
- Provides clear TypeError messages instead of cryptic assertion errors


mnorris11:
Hi @LincolnBurrows2017 , thanks for fixing this TODO, if you sign the CLA we can import and merge it.

PR #4921, opened 2026-03-12
Add batched SIMD distance computation with early abort for brute-force search

amrith:
## Summary

This PR adds early-abort distance computation to FAISS's brute-force sequential search paths. Vector dimensions are processed in fixed-size batches using existing SIMD functions, with a threshold check at batch boundaries. If the partial distance already exceeds the k-th best distance in the result heap, computation aborts early and skips the remaining dimensions.

This is a port of the optimization from the [nmslib knn-opt branch](https://github.com/nmslib/nmslib), adapted to FAISS's architecture.

## Changes

- `fvec_L2sqr_batched()`, `fvec_inner_product_batched()`, `fvec_Linf_batched()` — new batched distance functions with threshold parameter
- `DistanceComputer` base class extended with `operator()(idx_t i, float threshold)`, `supports_threshold()`, and `degenerate_threshold()` virtual methods (default implementations fall back to standard computation, so existing code is unaffected)
- `FlatL2Dis` and `FlatIPDis` override the threshold methods to use batched functions
- `exhaustive_L2sqr_seq()` and `exhaustive_inner_product_seq()` now use batched distances with the result handler's threshold
- 20 new unit tests: 9 batched distance correctness tests, 9 early-abort/edge-case tests, 2 BLAS-vs-sequential knn equivalence tests
- Benchmark tool and documentation

## Where it applies

The optimization runs in the sequential brute-force scan path, which is used when `nx < distance_compute_blas_threshold` (default 20) or when an `IDSelector` is active. The BLAS path is unaffected.

## Where it does NOT apply

HNSW and graph-based search — early abort returns partial distances that corrupt candidate ordering in graph traversal, reducing recall. This optimization is only safe for brute-force scans where distance is used solely for result filtering.

## Benchmark results (x86_64, AVX2, d=2048, k=10, 500K vectors)

| Data Distribution | Batch Size | Speedup |
|-------------------|------------|---------|
| L2, clustered     | 32         | **4.83x** |
| L2, clustered     | 64         | 3.21x   |
| L2, random        | 256        | 1.19x   |
| IP, random        | 256        | 1.02x   |

The optimization is data-dependent — clustered data with many distant candidates sees the largest gains. Random data shows modest improvement at larger batch sizes and slight overhead at smaller ones.

## Test results

- Full test suite: 197/197 passed on Linux x86_64, 208/208 passed on macOS ARM (2 mmap failures on macOS are pre-existing)
- BLAS-vs-sequential knn tests confirm both paths produce identical top-k results

## AI Use Disclaimer

The essence of the change is based on a similar change published to [nmslib](https://github.com/nmslib/nmslib/pull/572) and a prototype in [Lucene](https://github.com/amrith/lucene/tree/knn-opt).  Kiro, Amazon's AI-powered development assistant, was used to prepare this pull request. After an initial prototype fix was hand coded and verified, Kiro cleaned up the implementation, eliminated all the unnecessary test experiment code that was no longer required, refactored the implementation, augmented the tests and benchmark utility, and formatted the commit messages.


alibeklfc:
Hi! We are currently refactoring the SIMD code, and we will review your diff once that’s done. Thanks for your contribution.

amrith:
> Hi! We are currently refactoring the SIMD code, and we will review your diff once that’s done. Thanks for your contribution.

Thanks for the response. Looking forward to re-engaging once you have completed the refactor.

PR #4926, opened 2026-03-14
introduce `SingleQueryResultCollectHandler`

alexanderguzhva:
This PR introduces the following:
* a new type of a `ResultHandler` called `SingleQueryResultCollectHandler`
* adds a filtering capability to the core of the FastScan code
* `SIMDResultHandler::size_t in_range_num` and `SIMDResultHandlerToFloat::scan_cnt`, which can be effectively used for early stop checks. These are not currently used in the baseline `FAISS` code, but are used in the `Knowhere` fork 

@mdouze @mnorris11 @junjieqi please comment and suggest unit tests + locations that need to be written, given some possible unusual use cases on your end

### DIFF
--- a/faiss/impl/fast_scan/accumulate_loops.h
+++ b/faiss/impl/fast_scan/accumulate_loops.h
@@ -57,12 +57,17 @@ void accumulate_fixed_blocks(
         const Scaler& scaler,
         size_t block_stride) {
     constexpr int bbs = 32 * BB;
-    for (size_t j0 = 0; j0 < nb; j0 += bbs) {
+    for (size_t j0 = 0; j0 < nb; j0 += bbs, codes += block_stride) {
+        res.set_block_origin(0, j0);
+        if constexpr (has_sel_member_v<ResultHandler>) {
+            if (whether_all_vectors_filtered_out(
+                        res, std::min<size_t>(bbs, res.ntotal - j0)))
+                continue;
+        }
+
         FixedStorageHandler<NQ, 2 * BB> res2;
         kernel_accumulate_block<NQ, BB>(nsq, codes, LUT, res2, scaler);
-        res.set_block_origin(0, j0);
         res2.to_other_handler(res);
-        codes += block_stride;
     }
 }
 
@@ -123,7 +128,14 @@ void accumulate_q_4step_256(
     constexpr int Q4 = (QBS >> 12) & 15;
     constexpr int SQ = Q1 + Q2 + Q3 + Q4;
 
-    for (size_t j0 = 0; j0 < ntotal2; j0 += 32) {
+    for (size_t j0 = 0; j0 < ntotal2; j0 += 32, codes += block_stride) {
+        res.set_block_origin(0, j0);
+        if constexpr (has_sel_member_v<ResultHandler>) {
+            if (whether_all_vectors_filtered_out(
+                        res, std::min<size_t>(32, res.ntotal - j0)))
+                continue;
+        }
+
         FixedStorageHandler<SQ, 2> res2;
         const uint8_t* LUT = LUT0;
         pq4_kernel_qbs_256<Q1>(nsq, codes, LUT, res2, scaler);
@@ -142,9 +154,7 @@ void accumulate_q_4step_256(
             res2.set_block_origin(Q1 + Q2 + Q3, 0);
             pq4_kernel_qbs_256<Q4>(nsq, codes, LUT, res2, scaler);
         }
-        res.set_block_origin(0, j0);
         res2.to_other_handler(res);
-        codes += block_stride;
     }
 }
 
@@ -195,7 +205,14 @@ void pq4_accumulate_loop_qbs_fixed_scaler_256(
     }
 
     // Default: qbs not known at compile time
-    for (size_t j0 = 0; j0 < ntotal2; j0 += 32) {
+    for (size_t j0 = 0; j0 < ntotal2; j0 += 32, codes += block_stride) {
+        res.set_block_origin(0, j0);
+        if constexpr (has_sel_member_v<ResultHandler>) {
+            if (whether_all_vectors_filtered_out(
+                        res, std::min<size_t>(32, res.ntotal - j0)))
+                continue;
+        }
+
         const uint8_t* LUT = LUT0;
         int qi = qbs;
         int i0 = 0;
@@ -219,7 +236,6 @@ void pq4_accumulate_loop_qbs_fixed_scaler_256(
             i0 += nq;
             LUT += nq * nsq * 16;
         }
-        codes += block_stride;
     }
 }
 
--- a/faiss/impl/fast_scan/simd_result_handlers.h
+++ b/faiss/impl/fast_scan/simd_result_handlers.h
@@ -9,6 +9,7 @@
 
 #include <algorithm>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 #include <faiss/impl/simdlib/simdlib_dispatch.h>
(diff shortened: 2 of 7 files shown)


faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -785,6 +838,92 @@ struct PartialRangeHandler : RangeHandler<C, with_id_map, SL> {
...
+    }
+
+    void handle(size_t q, size_t b, simd16uint16 d0, simd16uint16 d1) final {
+        if (this->disable) {
+            return;
+        }
+
+        this->adjust_with_origin(q, d0, d1);
+
+        uint32_t lt_mask = this->get_lt_mask(C::neutral(), b, d0, d1);
```
mdouze, line 904: so this mask is all-true?

faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -312,10 +340,13 @@ struct SingleResultHandler : ResultHandlerCompare<C, with_id_map, SL> {
                 auto real_idx = this->adjust_id(b, j);
                 lt_mask -= 1 << j;
                 if (this->sel->is_member(real_idx)) {
+                    this->scan_cnt++;
                     T d = d32tab[j];
                     if (C::cmp(idis[q], d)) {
                         idis[q] = d;
                         ids[q] = real_idx;
+
+                        this->in_range_num += 1;
```
mdouze: += 1 is more pythonic than C++

faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -785,6 +838,92 @@ struct PartialRangeHandler : RangeHandler<C, with_id_map, SL> {
...
+ * Unlike HeapHandler/ReservoirHandler which maintain a top-k structure,
+ * this handler appends every result that passes the threshold to a
+ * vector of (distance, id) pairs. Useful for exhaustive collection
+ * (e.g., "return all" searches) in the fast_scan SIMD path.
+ */
+template <
+        class C,
+        bool with_id_map = false,
+        SIMDLevel SL = SINGLE_SIMD_LEVEL_256>
+struct SingleQueryResultCollectHandler
```
mdouze, line 873: Usage looks legitimate, even if we could have the same result with a range search with infinite threshold

faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -312,10 +340,13 @@ struct SingleResultHandler : ResultHandlerCompare<C, with_id_map, SL> {
                 auto real_idx = this->adjust_id(b, j);
                 lt_mask -= 1 << j;
                 if (this->sel->is_member(real_idx)) {
+                    this->scan_cnt++;
```
mnorris11, line 364: On line 343, scan_cnt is incremented above the `cmp`, but on 360 it is incremented inside the cmp. I think they should be the same, either both inside or both outside? (even though this branch checks selector)

faiss/impl/fast_scan/accumulate_loops.h
```diff
@@ -57,12 +57,29 @@ void accumulate_fixed_blocks(
...
+                    auto real_idx = res.adjust_id(0, jj);
+                    if (res.sel->is_member(real_idx)) {
+                        skip_flag = false;
+                        break;
+                    }
+                }
+                if (skip_flag) {
+                    continue;
+                }
+            }
```
mnorris11: I know the blocks are slightly different in places, but is there any way of factoring the loops out to 1 function?

faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -785,6 +838,92 @@ struct PartialRangeHandler : RangeHandler<C, with_id_map, SL> {
...
+                T dis = d32tab[j];
+                int64_t idx = this->adjust_id(b, j);
+                collect.emplace_back(idx, dis);
+                this->in_range_num += 1;
+            }
+        }
+    }
+
+    void end() override {
+        if (normalizers) {
```
mnorris11, line 938: Reset `scan_cnt` here?

mnorris11:
Possible test locations and cases (suggested by Claude; I made a pass to remove the complete nonsense ones but not all of these might be relevant)

test_fast_scan.cpp or a new test_single_query_collect_handler.cpp:
- Test SingleQueryResultCollectHandler returns an expected number of results with a certain threshold
- Test with an IDSelector to verify filtering works

Block-skip filtering tests:
- Build an index, search with an IDSelectorNot that excludes all vectors in certain blocks, verify results match a non-fast-scan index
- Verify that filtering doesn't break when blocks are partially filtered

Edge cases:
- Empty selector (all filtered out)
- ntotal not a multiple of 32 (partial last block)

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: I tend to write `+= 1;` rather than `++;`, because it is easy to read and to replace with `+= n;`, if needed :)
Let me tweak it.

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: definitely, something is wrong here. let me see

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: is there anything I need to do?

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: yes. Given this, it was still decided to leave the following code to be consistent with the other similar blocks and let the compiler / CPU branch predictor handle it.

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: I'll add one for the consistency

faiss/impl/fast_scan/accumulate_loops.h
alexanderguzhva: Makes sense. Will do.

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: done

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: done

faiss/impl/fast_scan/accumulate_loops.h
alexanderguzhva: done

faiss/impl/fast_scan/simd_result_handlers.h
alexanderguzhva: done

alexanderguzhva:
> Possible test locations and cases (suggested by Claude; I made a pass to remove the complete nonsense ones but not all of these might be relevant)
> 
> test_fast_scan.cpp or a new test_single_query_collect_handler.cpp:
> 
>     * Test SingleQueryResultCollectHandler returns an expected number of results with a certain threshold
> 
>     * Test with an IDSelector to verify filtering works
> 
> 
> Block-skip filtering tests:
> 
>     * Build an index, search with an IDSelectorNot that excludes all vectors in certain blocks, verify results match a non-fast-scan index
> 
>     * Verify that filtering doesn't break when blocks are partially filtered
> 
> 
> Edge cases:
> 
>     * Empty selector (all filtered out)
> 
>     * ntotal not a multiple of 32 (partial last block)

If suggested by Claude, then implemented by Claude as well. But I've checked the uts manually

alexanderguzhva:
is it possible that 1 faiss unit test is currently flaky? an AVX2 test seems to fail

mnorris11:
> is it possible that 1 faiss unit test is currently flaky? an AVX2 test seems to fail

Tests are passing, trying to address internal review comments atm

mnorris11:
Heads up @alexanderguzhva , I had to make some small refactors on top of this, but I cannot export to your PR, so the change commit will be slightly different from this PR. (Check for `for_each_block` in the coming commit)

PR #4952, opened 2026-03-18
enable specifying a subset of SIMD levels that are implemented at dispatching time.

mdouze:
Summary:
So far the `with_simd_level` function would assume that all  implementations are available for each function. This is not the case, and  is effectively not the case when we enable AVX512_SPR.

This diff has a specialized `with_selected_simd_levels` that takes a bitmask of the available implementations as template parameter.

The previous `with_simd_level` remains, it assumes AVX2, NEON and AVX512 implementations are available.

Differential Revision: D96943554


mnorris11:
Abandoned internally, so closing this one

Issue #4890, opened 2026-03-09
C++20 features, while only C++17 may be implied


alexanderguzhva:
using templated lambdas is a C++20 feature. As a result, a compilation may trigger the following compiler warning (`warning: explicit template parameter list for lambdas is a C++20 extension [-Wc++20-extensions]`) at the following lines:
https://github.com/facebookresearch/faiss/blob/bb298a2323c1df5fd4ee4cdf8ddae23d6614e45c/faiss/IndexIVFPQ.cpp#L1250
https://github.com/facebookresearch/faiss/blob/bb298a2323c1df5fd4ee4cdf8ddae23d6614e45c/faiss/IndexIVFPQ.cpp#L1252
https://github.com/facebookresearch/faiss/blob/bb298a2323c1df5fd4ee4cdf8ddae23d6614e45c/faiss/IndexIVFPQ.cpp#L1270

check https://github.com/facebookresearch/faiss/blob/bb298a2323c1df5fd4ee4cdf8ddae23d6614e45c/CMakeLists.txt#L56-L60

mdouze:
Right we should move all code to C++20...

mnorris11:
I am trying to finalize it in https://github.com/facebookresearch/faiss/pull/4881. cuVS was the last holdout because we were on an older version, but that landed this week, so we can move to 20 more officially.

Anarion-zuo:
Ran into a similar issue. Posted a PR here. https://github.com/facebookresearch/faiss/pull/4962

mnorris11:
We're on C++20 now, so I will close.

PR #4962, opened 2026-03-19
fix: C++ 20 build failure

Anarion-zuo:
**Summary**
Issue: #4890 
- Replace C++20 templated lambdas in `IndexIVFPQ::get_InvertedListScanner` with C++17-compatible functor structs.
- Preserve existing SIMD/decoder dispatch behavior while avoiding `-Wc++20-extensions` warnings.

**Changes**
- Add namespace-scope functors to handle scanner creation, decoder selection, and SIMD dispatch.
- Keep logic for metric type, PQ nbits, and selector handling unchanged.

**Tests**
- `ctest --test-dir build-docker-tests --output-on-failure -j2`
- `pytest -q tests/test_fast_scan_ivf.py tests/test_index.py tests/test_index_accuracy.py tests/test_search_params.py`

**Notes**
- C++17-compatible refactor to align with `FAISS_ENABLE_CUVS` CMake setting.


mnorris11:
We would rather just move to C++ 20. It will get done in https://github.com/facebookresearch/faiss/pull/4881.

mnorris11:
I will close this because we have landed the C++20 for cuVS, but thank you for the contribution!

Issue #4927, opened 2026-03-14
[BUG] IndexHNSWCagra with Base layer only return 0 results for range_search


navneet1v:
## Description
I was running HNSWCagra Index range_search and comparing it with HNSWIndex. What I realized was when CagraIndex contains only base layer, then in that case number of results returned is 0. On doing further deep-dive what I can see was since, CagraIndex with base layer only doesn't have an entry point, the code exists with 0 results.

Code ref:
https://github.com/facebookresearch/faiss/blob/df0dea6c6d8951056763dc03528b3973c6ba26e2/faiss/impl/HNSW.cpp#L845-L847

For normal topK search with base layer only, there is a special code for searching in the base layer only. Ref: https://github.com/facebookresearch/faiss/blob/df0dea6c6d8951056763dc03528b3973c6ba26e2/faiss/IndexHNSW.cpp#L966-L977

### Reproduction Code

I created this below code to repro the bug. In the code I have 3 indices, 
1. CagraIndex with all layers.
2. HNSWIndex
3. CagraIndex with base layer only copied from 1 using copyBaseLevelOnly function.

For index 3, we are not seeing any results.

```
/**
* Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

#include <cstdio>
#include <iostream>
#include <random>
#include <vector>

#include <faiss/IndexFlat.h>
#include <faiss/IndexHNSW.h>
#include <faiss/impl/AuxIndexStructures.h>
#include <faiss/impl/HNSW.h>

/// Copy the base layer graph from src to a new IndexHNSWCagra (dst)
/// that only has level 0. Modeled after GpuIndexCagra::copyTo().
void copyBaseLevelOnly(
        const faiss::IndexHNSWCagra& src,
        faiss::IndexHNSWCagra& dst) {
    auto n = src.ntotal;
    auto d = src.d;
    auto M = src.hnsw.nb_neighbors(0) / 2;
    auto graph_degree = src.hnsw.nb_neighbors(0);

    // Setup storage
    if (dst.storage && dst.own_fields) {
        delete dst.storage;
    }
    dst.storage = new faiss::IndexFlatL2(d);
    dst.own_fields = true;
    dst.d = d;
    dst.metric_type = src.metric_type;
    dst.is_trained = true;
    dst.keep_max_size_level0 = true;

    // Reset and reconfigure HNSW structure
[...]

PR #4963, opened 2026-03-19
fix: cagra bug

Anarion-zuo:
**Summary**
Resolved https://github.com/facebookresearch/faiss/issues/4927.
- Fixes base-level-only `IndexHNSWCagra::range_search` to return results without an entry point.
- Adds a C++ unit test reproducing the base-level-only range search case and a GPU interop Python test.

**Changes**
- Implement base-level-only range search path in `IndexHNSWCagra` using per-query range handlers.
- Add C++ helper to copy level-0 graph only and a new test for base-level-only range search.
- Add GPU interop Python test for base-level-only range search.

**Tests**
- `docker build -f docker/Dockerfile.dev -t faiss-dev:local --build-arg UID="$(id -u)" --build-arg GID="$(id -g)" .`
- `docker run --rm -v "$PWD:/workspace" -w /workspace faiss-dev:local bash -lc "cmake -B build-docker -G Ninja -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=ON -DBUILD_TESTING=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo . && cmake --build build-docker -j --target faiss faiss_test swigfaiss && ./build-docker/tests/faiss_test --gtest_filter=HNSW.Test_IndexHNSWCagra_BaseLevelOnly_RangeSearch"`

**Notes**
- The Python GPU test is skipped without CUVS/GPU support; it is included for GPU-enabled CI.


navneet1v:
@Anarion-zuo thanks for raising the PR. This is way better solution than mine. I had to make changes at multiple places. But this far more cleaner

faiss/IndexHNSW.cpp
```diff
@@ -285,6 +283,7 @@ void hnsw_search(
                 ndis += stats.ndis;
                 nhops += stats.nhops;
                 res.end();
+                vt.advance();
```
navneet1v, line 287: why this is needed? just for understanding?

Anarion-zuo: `VisitedTable vt` is reused across queries inside the same parallel region for performance (to avoid reallocating/clearing a large visited array for every query). `vt.advance()` is what “resets” the visited state between queries in O(1) by bumping an epoch counter (or clearing the hashset variant), so each query starts with an empty visited set.

Without `vt.advance()`, nodes visited while processing query *i* would remain marked as visited for query *i+1*, which can cause the search to skip almost everything and return incorrect / empty results.

This is the same pattern used elsewhere in HNSW search code, e.g. in the main search loop that shares a `VisitedTable` across queries and calls `advance()` after each query.

Anarion-zuo:
@navneet1v  Hi. It's been some time. Do I need a maintainer's approval to trigger the CI tests?

Anarion-zuo:
@navneet1v Hi. Are you a maintainer? Someone should trigger the CI on this one.

navneet1v:
> @navneet1v Hi. Are you a maintainer? Someone should trigger the CI on this one.

I am not the maintainer

navneet1v:
taggin @mdouze

mnorris11:
cc @cjnolet @tarang-jain

mnorris11:
@Anarion-zuo would you be able to fix the unit tests? Are the logs visible to you?

Anarion-zuo:
@mnorris11 I fixed the visible part of the CI.

Anarion-zuo:
@mnorris11 gentle ping

faiss/impl/simdlib/simdlib_avx512.h
```diff
@@ -394,4 +394,44 @@ inline float horizontal_add(const simd16float32_tpl<SIMDLevel::AVX512>& a) {
     return _mm512_reduce_add_ps(a.f);
 }
 
+template <>
+struct simd512bit_tpl<SIMDLevel::AVX512_SPR>
```
mnorris11: Why did the template have to change for AVX512 if the issue is related to CAGRA / HNSW? #5015 added SIMD dispatch for HNSW.

Anarion-zuo: It was meant to bypass certain issues in x86 spr machine building & running. Removing it would cause the CI to fail.

Anarion-zuo: So adopting https://github.com/facebookresearch/faiss/pull/5015 can resolve the CI issue?

Anarion-zuo:
I can't see the facebook internal linter failure. Otherwise it's all fixed up.

faiss/impl/simdlib/simdlib_avx512.h
mnorris11: Yes, it should pass on latest code

Issue #4742, opened 2025-12-28
GpuIndexCagra fails search if the source IndexHNSWCagra is destroyed after copyFrom


mageirakos:
My Goal: 
---
Happy New Year!

I want to be able to persist CAGRA GPU indexes.

The standard Faiss workflow for this as I understand it, and please correct me if I am wrong, is:

1 GpuIndexCagra (GPU) $\rightarrow$ copyTo $\rightarrow$ IndexHNSWCagra (CPU).
2. write_index (CPU Index) to disk.
3. read_index (CPU Index) from disk.
4. IndexHNSWCagra(CPU) $\rightarrow$ copyFrom or index_cpu_to_gpu $\rightarrow$ GpuIndexCagra (GPU).

Problem Description:
---
In DGX-Spark :

When cloning a CPU `IndexHNSWCagra` to a GPU CAGRA index using `copyFrom/index_cpu_to_gpu`, the resulting GPU index appears to maintain a shallow dependency on the host memory of the source index. If the source CPU index is destroyed, subsequent searches on the GPU index fail with a CUDA illegal memory access. I also get invalid  device ordinal but you can see on the output that it should be 0. I tried to follow this pattern from `index_cpu_to_gpu` internally : [GpuCloner.cpp#L236-L244](https://github.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuCloner.cpp#L236-L244)

Interesting:
- In a different H100 machine: The same script works, and I do not get the above problem. Which might hint to either a unified memory issue, or some mis-configuration? 
- Also if you reduce nb, so that we're using less memory in the DGX-Spark the crash does not happen.

Setup
----
Tested on both machines in a container with library versions: 
```
-- Faiss version:  1.13.0
-- cuVS version:   25.08.0
-- RAFT version:   25.08.0
-- CUDA version:   12.8.93
```

Minimal script for reproducibility:
``` cpp
#include <cuda_runtime.h>
#include <faiss/gpu/GpuIndexCagra.h>
#include <faiss/gpu/StandardGpuResources.h>
#include <faiss/IndexHNSW.h>
#include <iostream>
#include <memory>
#include <vector>

void print_available_devices() {
    int deviceCount = 0;
    if (cudaGetDeviceCount(&deviceCount) != cudaSuccess || deviceCount == 0) {
        std::cerr << "No CUDA devices found." << std::endl;
        return;
    }
    std::cout << "--- Available CUDA Devices ---" << std::endl;
[...]

PR #4968, opened 2026-03-20
fix: GPU CAGRA copyFrom host-memory lifetime dependency and add regression test

Anarion-zuo:
## Summary
This PR fixes https://github.com/facebookresearch/faiss/issues/4742 , where a faiss::gpu::GpuIndexCagra (and binary variant) created via copyFrom / index_cpu_to_gpu could still depend on CPU-side state after the source CPU index was destroyed. In the reported repro, this could lead to CUDA illegal memory access / invalid device ordinal errors during subsequent GPU search.

The PR also fixes an out-of-bounds write in dense kNN-graph construction in GpuIndexCagra::copyFrom_ex and adds a regression test that explicitly deletes the CPU source index before running GPU search.

## Changes
- Rework the CAGRA copyFrom path so the resulting GPU index no longer depends on the lifetime of the source CPU IndexHNSWCagra after the transfer completes.
- Fix dense kNN-graph allocation in GpuIndexCagra::copyFrom_ex by using resize() instead of reserve() before indexed writes.
- Apply the same lifetime fix to GpuIndexBinaryCagra::copyFrom .
- Update CAGRA GPU unit tests to delete the source CPU index immediately after copyFrom and then run search, matching the reported repro scenario.
## Reproduction
Build GPU index -> copyTo CPU -> copyFrom back to GPU -> destroy CPU index -> search on GPU.

## Testing
- Added/updated GPU regression tests in TestGpuIndexCagra.cu to cover the case where the CPU source is deleted immediately after copyFrom .
- Verified that GPU search still succeeds after the source CPU index has been destroyed.
If you want, I can also make this a bit more concise and “merge-ready” in Meta/Faiss PR style.


mnorris11:
cc @cjnolet @tarang-jain

faiss/gpu/GpuIndexCagra.cu
```diff
@@ -331,28 +336,33 @@ void GpuIndexCagra::copyFrom_ex(
         auto base_index = dynamic_cast<IndexFlat*>(index->storage);
         FAISS_ASSERT(base_index);
         auto dataset = base_index->get_xb();
+        auto total = static_cast<size_t>(index->ntotal) * index->d;
+        host_storage_f32_.assign(dataset, dataset + total);
```
tarang-jain: This might not be the ideal strategy and I think the idea was intentional to not create an owning copy. We ideally don't want to be copying the entire dataset to the cagra index because it can be huge. We can instead add a warning while doing the copyFrom to inform the user to keep the HNSW index in scope.

Anarion-zuo: Changed a bit of the code to resolve this. Added a warning in `GpuIndexBinaryCagra::copyFrom` that the GPU index keeps a non‑owning pointer to the CPU dataset, so the source `IndexBinaryHNSWCagra` must remain alive for as long as the GPU index is used.

Anarion-zuo:
@mnorris11 Fixed the format and x86 building CI errors, at least on my machine.

faiss/gpu/GpuIndexCagra.cu
mnorris11: @tarang-jain how does the current approach look?

Anarion-zuo:
@mnorris11 It seems the CIs are all resolved.

tarang-jain:
@Anarion-zuo can you update the PR description. We are no longer doing the copy of the dataset.

faiss/gpu/GpuIndexCagra.cu
tarang-jain: Looks good. I am approving this. So basically this PR is just adding a warning it seems?

tarang-jain:
In the meantime I am approving the diff.

Anarion-zuo:
@tarang-jain PR description updated.

PR #4974, opened 2026-03-24
Fix HNSW Panorama Perf Bug

AlSchlo:
For some reason a refactor in my last commit of the Panorama HNSW PR removed the update of the heap threshold, so nothing was getting pruned (and hence there was no speedup)... 😓 

This PR also adds PanoramaStats to HNSW and a unit test to avoid this from happening again.


faiss/impl/HNSW.cpp
```diff
@@ -976,6 +983,7 @@ int search_from_candidates_panorama(
         for (size_t i = 0; i < batch_size; i++) {
             idx_t idx = index_array[i];
             if (res.add_result(exact_distances[i], idx)) {
+                threshold = res.threshold;
```
AlSchlo, line 986: This line was accidentally removed.

PR #5003, opened 2026-03-29
reuse unpack buffer in multibit IVFRaBitQFastScan refinement 

lyang24:
This removes two pieces of avoidable overhead from multibit `IndexIVFRaBitQFastScan` refinement.

  In `compute_full_multibit_distance()`, we were allocating a temporary unpack buffer and reacquiring the current list codes for every refined candidate. The handler already has the list pointer, so this patch just reuses that and keeps a reusable unpack buffer on
  the handler.

  I only measured this on multibit workloads. On my AVX2 machine (`d=128`, `METRIC_L2`), the gain is small but consistent:
  - `127us -> 126us`
  - `1335us -> 1292us`
  - `6414us -> 6084us`
  - `17955us -> 16800us`

  No change to 1-bit logic or search semantics.

### DIFF
--- a/benchs/bench_cppcontrib_sa_decode.cpp
+++ b/benchs/bench_cppcontrib_sa_decode.cpp
@@ -12,7 +12,6 @@
 #include <iostream>
 #include <memory>
 #include <random>
-#include <thread>
 #include <tuple>
 #include <vector>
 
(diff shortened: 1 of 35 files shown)


mnorris11:
cc @alibeklfc

PR #5005, opened 2026-03-30
use `level` but not `0` in neighbor_range(i, level, &begin, &end) (line 198)

nameearly:
just like line 95-103 in same file (\faiss\impl\HNSW.cpp):
```
void HNSW::clear_neighbor_tables(int level) {
    for (size_t i = 0; i < levels.size(); i++) {
        size_t begin, end;
        neighbor_range(i, level, &begin, &end);  // ✅ right：use `level` 
        // ...
    }
}
```


mnorris11:
Thanks @nameearly for the contribution! It has been merged.

PR #5016, opened 2026-03-31
read-heap-buffer-overflow (size 32) in float faiss::fvec_norm_L2sqr<>()

limqiying:
Differential Revision: D98879864


mnorris11:
Internal diff already closed

PR #5046, opened 2026-04-06
Fix Windows ARM64 (MSVC) build: disable NEON SIMD path

### LINKED ISSUE #4993: Windows ARM64 (MSVC) build broken by NEON SIMD templatization (D95392150)
## Summary

The NEON SIMD templatization landed via D95392150 (commit `8d8268c8a`, GitHub PR #4866) breaks the build on **Windows ARM64 with MSVC**. The previous code (`simdlib_neon.h` before templatization) compiled successfully.

## Environment

- **Compiler:** MSVC (Visual Studio 2022), targeting `arm64`
- **OS:** Windows 11 ARM64
- **faiss version:** main branch at commit `8d8268c8a` and later

## Errors

Three categories of errors occur:

### 1. NEON intrinsics cannot be used as non-type template parameters

MSVC implements NEON intrinsics (`vsubq_u16`, `vandq_u16`, `vorrq_u16`, etc.) as **compiler builtins**, not as real functions. Their address cannot be taken, so the `binary_func_impl::call<&vsubq_u16>()` pattern in `simdlib_neon.h` fails:

```
faiss/impl/simdlib/simdlib_neon.h(540): error C2065: 'vsubq_u16': undeclared identifier
faiss/impl/simdlib/simdlib_neon.h(540): error C2975: 'F': invalid template argument, expected compile-time constant expression
```

This affects all lines using `.call<&intrinsic>()` (lines 540, 552, 564, etc.).

### 2. Duplicate function definitions due to MSVC ARM64 type aliasing

On MSVC ARM64, `uint8x16x2_t` and `uint16x8x2_t` (and other NEON struct types) resolve to the same underlying `__n128x2` type. This causes overload ambiguity:

```
faiss/impl/simdlib/simdlib_neon.h(34): error C2084: function 'uint8x16_t faiss::detail::reinterpret_u8(const uint8x16x2_t &)' already has a body
```

### 3. Missing members/operators on NEON-specialized template types

Several members and operators are not defined for the NEON specializations:

```
faiss/impl/fast_scan/simd_result_handlers.h(184): error C2039: 'get_scalar_0': is not a member of 'faiss::simd16uint16_tpl<faiss::SIMDLevel::ARM_NEON>'
faiss/IndexIVFPQFastScan.cpp(206): error C2039: 'store': is not a member of 'faiss::simd8float32_tpl<faiss::SIMDLevel::ARM_NEON>'
```

Operators `>>` and `&` are also missing for NEON types in `kernels_simd256.h`.

## Suggested Fixes

1. **Wrap NEON builtins in real functions** so their addresses can be taken (or replace the function-pointer template pattern with a different dispatch mechanism for MSVC).
2. **Guard overloaded `reinterpret_*` functions** with `#ifndef _MSC_VER` or use SFINAE / `if constexpr` to avoid ambiguity from collapsed NEON types on MSVC.
3. **Ensure all required members** (`get_scalar_0`, `store`, operators `>>`, `&`) are defined for the ARM_NEON specializations.

Alternatively, falling back to `SIMDLevel::NONE` (emulated) on MSVC ARM64 would be a quick workaround until proper MSVC NEON support is added.

## Related

- #4857 — MSVC `__restrict` mismatch from the same templatization effort (x86, different issue)
- #4890 — C++20 features in templatized code while only C++17 is required

## Reproducer

Build faiss with CMake on Windows 11 ARM64 using MSVC:

```
cmake -B build -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=OFF -DBUILD_TESTING=OFF -A ARM64
cmake --build build
```

CI logs showing the failure: https://github.com/faiss-wheels/faiss-wheels/actions/runs/23570080174/job/68630487844?pr=158

Chessing234:
## Summary

- Fixes the Windows ARM64 (MSVC) build failure caused by the NEON SIMD templatization (D95392150 / PR #4866)
- MSVC ARM64 now falls back to the scalar (`SIMDLevel::NONE`) emulated implementation, which compiles and runs correctly
- Three root causes are addressed by this workaround: (1) MSVC NEON builtins cannot be used as non-type template parameters, (2) MSVC collapses distinct NEON struct types to the same `__n128x2`, and (3) missing members/operators on NEON template specializations

## Changes

- **`faiss/CMakeLists.txt`**: Added `AND NOT MSVC` guard to the ARM64 NEON block so `COMPILE_SIMD_ARM_NEON` is not defined and NEON source files are not compiled on MSVC ARM64
- **`cmake/link_to_faiss_lib.cmake`**: Same guard for the Dynamic Dispatch (DD) mode ARM64 path
- **`faiss/impl/simdlib/simdlib_dispatch.h`**: Added `_M_ARM64` recognition (MSVC's equivalent of `__aarch64__`) with explicit `!defined(_MSC_VER)` exclusion, plus explanatory comments

## Test plan

- [ ] Verify the build succeeds on Windows ARM64 with MSVC: `cmake -B build -DFAISS_ENABLE_GPU=OFF -DFAISS_ENABLE_PYTHON=OFF -DBUILD_TESTING=OFF -A ARM64 && cmake --build build`
- [ ] Verify existing Linux/macOS aarch64 builds (GCC/Clang) are unaffected and still use the NEON path
- [ ] Verify x86_64 builds are unaffected

Fixes #4993


algoriddle:
Thanks for the fix! The CMakeLists.txt and link_to_faiss_lib.cmake changes look correct.

One simplification: the `simdlib_dispatch.h` change is redundant. Since the CMakeLists.txt change already ensures `COMPILE_SIMD_ARM_NEON` is never defined on MSVC ARM64, the `#elif defined(COMPILE_SIMD_ARM_NEON)` branch can never be reached on MSVC. Adding `&& !defined(_MSC_VER)` and `|| defined(_M_ARM64)` there has no effect — both are dead code. I'd drop that file from the PR and keep just the two CMake changes. A comment in CMakeLists.txt (which you already have) is sufficient to document the workaround.

alibeklfc:

Chessing234:
Thanks both! @algoriddle fair point — the `simdlib_dispatch.h` guard is redundant once the CMakeLists.txt change guarantees `COMPILE_SIMD_ARM_NEON` is never defined on MSVC ARM64. Happy to drop that hunk to keep the diff minimal.

@alibeklfc understood — I'll wait for the SIMD refactor to land before revising, so the fix lines up with the new layout instead of having to be rewritten twice. Ping me if you'd like me to rebase/trim earlier.

PR #5057, opened 2026-04-07
Fix: static SIMD dispatch falls to scalar for avx512_spr/avx512/arm_sve builds

mulugetam:
While running `benchs/bench_rabitq.py`, I observed that the execution path is using scalar instructions instead of AVX-512, despite FAISS being built with a static dispatch target of `avx512_spr`. Below is a summary and a fix.

**Summary**
 
Static (non-DD) builds with `avx512_spr`, and partially `avx512` and `arm_sve`, silently fall back to scalar (`SIMDLevel::NONE`) for all SIMD-dispatched functions. This is because the static dispatch path in `with_selected_simd_levels` lacks the fallthrough logic that the DD path implements via `switch`/`[[fallthrough]]`.
 
**Problem**
 
The static dispatch in `faiss/impl/simd_dispatch.h` performs a single check:
 
```cpp
if constexpr (available_levels & (1 << int(SINGLE_SIMD_LEVEL))) {
    return action.template operator()<SINGLE_SIMD_LEVEL>();
} else {
    return action.template operator()<SIMDLevel::NONE>();
}
```

This is a binary decision and none of the predefined masks (`A0`, `A1`, `A2`, `AVX2_NEON`, `MINIMAX_HEAP_SIMD_LEVELS`) include `AVX512_SPR`, so in a static `avx512_spr` build, it resolves to scalar.
 
For example, `RaBitQuantizer` dispatches via `with_selected_simd_levels<AVAILABLE_SIMD_LEVELS_A0>(...)`.
`AVAILABLE_SIMD_LEVELS_A0` is defined as:

```cpp   
constexpr int AVAILABLE_SIMD_LEVELS_A0 =
       AVAILABLE_SIMD_LEVELS_AVX2_NEON | (1 << int(SIMDLevel::AVX512));
```
       
This mask includes bits for `NONE (0)`, `AVX2 (1)`, `AVX512 (2`), and `ARM_NEON (4)` but it does NOT include `AVX512_SPR (3)`.

The DD path handles this correctly because its `switch` statement falls through from `AVX512_SPR` --> `AVX512` --> `AVX2` --> `NONE`, picking the best available implementation. The static path had no equivalent mechanism.
 
**Affected configurations**
 
| Static build | Mask | Dispatches to (current) | Should dispatch to |
|---|---|---|---|
| `avx512_spr` | A0, A1, MINIMAX | `NONE` | `AVX512` |
| `avx512_spr` | A2, AVX2_NEON | `NONE` | `AVX2` |
| `avx512` | A2, AVX2_NEON | `NONE` | `AVX2` |
| `arm_sve` | A0, AVX2_NEON | `NONE` | `ARM_NEON` |
 
DD builds are not affected.
 
**Fix**
 
Add a compile-time fallthrough chain in the static dispatch path that mirrors the DD runtime behavior:
 
- `AVX512_SPR` --> try `AVX512` --> try `AVX2` --> `NONE`
- `AVX512` --> try `AVX2` --> `NONE`
- `ARM_SVE` --> try `ARM_NEON` --> `NONE`
 
This would fix broken (level x mask) combinations across distances, RaBitQ, scalar/product quantizer, HNSW, IndexFlat, IVF, and fused distances.

### DIFF
--- a/faiss/impl/simd_dispatch.h
+++ b/faiss/impl/simd_dispatch.h
@@ -47,6 +47,37 @@ constexpr int AVAILABLE_SIMD_LEVELS_A2 = AVAILABLE_SIMD_LEVELS_NONE |
 
 constexpr int AVAILABLE_SIMD_LEVELS_ALL = -1;
 
+constexpr SIMDLevel get_simd_fallback(SIMDLevel level) {
+    switch (level) {
+        case SIMDLevel::AVX512_SPR:
+            return SIMDLevel::AVX512;
+        case SIMDLevel::AVX512:
+            return SIMDLevel::AVX2;
+        case SIMDLevel::ARM_SVE:
+            return SIMDLevel::ARM_NEON;
+        case SIMDLevel::AVX2:
+        case SIMDLevel::ARM_NEON:
+        case SIMDLevel::RISCV_RVV:
+            return SIMDLevel::NONE;
+        default:
+            return SIMDLevel::NONE;
+    }
+}
+
+template <int available_levels, SIMDLevel current_level, typename LambdaType>
+inline auto dispatch_with_fallback(LambdaType&& action) {
+    if constexpr (available_levels & (1 << int(current_level))) {
+        return action.template operator()<current_level>();
+    } else if constexpr (current_level != SIMDLevel::NONE) {
+        return dispatch_with_fallback<
+                available_levels,
+                get_simd_fallback(current_level)>(
+                std::forward<LambdaType>(action));
+    } else {
+        return action.template operator()<SIMDLevel::NONE>();
+    }
+}
+
 /** The complete dispatching function. It takes into account:
  * - the currently selected SIMD level
  * - the compiled in SIMD levels (given by COMPILE_SIMD_XXX)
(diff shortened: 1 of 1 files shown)


alibeklfc:

faiss/impl/simd_dispatch.h
```diff
@@ -114,11 +114,32 @@ inline auto with_selected_simd_levels(LambdaType&& action) {
...
+            (SINGLE_SIMD_LEVEL == SIMDLevel::AVX512_SPR ||
+             SINGLE_SIMD_LEVEL == SIMDLevel::AVX512) &&
+            (available_levels & (1 << int(SIMDLevel::AVX2)))) {
+        // AVX512_SPR/AVX512 -> AVX2 fallthrough
+        return action.template operator()<SIMDLevel::AVX2>();
+    } else if constexpr (
+            SINGLE_SIMD_LEVEL == SIMDLevel::ARM_SVE &&
+            (available_levels & (1 << int(SIMDLevel::ARM_NEON)))) {
+        // ARM_SVE -> ARM_NEON fallthrough.
+        return action.template operator()<SIMDLevel::ARM_NEON>();
```
mnorris11: @mulugetam Thanks for this fix; is it ok if we change this around slightly?

We can define these first (around line 50)
```
constexpr SIMDLevel get_simd_fallback(SIMDLevel level) {
    switch (level) {
        case SIMDLevel::AVX512_SPR:
            return SIMDLevel::AVX512;
        case SIMDLevel::AVX512:
            return SIMDLevel::AVX2;
        case SIMDLevel::ARM_SVE:
            return SIMDLevel::ARM_NEON;
        case SIMDLevel::AVX2:
        case SIMDLevel::ARM_NEON:
        case SIMDLevel::RISCV_RVV:
            return SIMDLevel::NONE;
        default:
            return SIMDLevel::NONE;
    }
}

template <int available_levels, SIMDLevel current_level, typename LambdaType>
inline auto dispatch_with_fallback(LambdaType&& action) {
    if constexpr (available_levels & (1 << int(current_level))) {
        return action.template operator()<current_level>();
    } else if constexpr (current_level != SIMDLevel::NONE) {
        return dispatch_with_fallback<
                available_levels,
                get_simd_fallback(current_level)>(
                std::forward<LambdaType>(action));
    } else {
        return action.template operator()<SIMDLevel::NONE>();
    }
}
```

then we can change this section to 
```
    return dispatch_with_fallback<available_levels, SINGLE_SIMD_LEVEL>(
            std::forward<LambdaType>(action));
```

mulugetam:
@mnorris11 rebased with the recommend changes. Thanks!

PR #5067, opened 2026-04-09
Optimize the DistanceComputerByte::compute_code_distance with AVX512F, AVX512BW, AVX512VL instructions

baohongmin:
Using AVX-512 F/BW/VL instructions to optimize the DistanceComputerByte::compute_code_distance to improve vector load.

Test with one million 128 dimensional vectors: L2 and Inner Product calculation: >1.3X performance improvement.
Significant performance gains are achieved when dimensionality exceeds 16.


alibeklfc:

faiss/impl/scalar_quantizer/sq-avx512.cpp
```diff
@@ -677,24 +677,54 @@ struct DistanceComputerByte<Similarity, SIMDLevel::AVX512>
...
-
-            __m512i v;
+        constexpr int kLanes = 32;
+        int i = 0;
+        for (; i + kLanes <= d; i += kLanes) {
+            __m256i c1 = _mm256_loadu_epi8(code1 + i);
+            __m256i c2 = _mm256_loadu_epi8(code2 + i);
+            __m512i c1i16 = _mm512_cvtepu8_epi16(c1);
+            __m512i c2i16 = _mm512_cvtepu8_epi16(c2);
+#ifdef __AVX512VNNI__
```
mnorris11: @baohongmin thanks for this contribution. Our refactor for the SIMD in Faiss has completed, so we are working through the backlog. 

it looks like VNNI is supported on Cascade Lake onwards on Intel. https://www.intel.com/content/www/us/en/support/articles/000058341/processors/intel-xeon-processors.html

The concern I have is that we are deliberately trying to avoid the `#ifdef` inside files, preferring just once at the top of relevant files. However, we don't have a VNNI specialization right now. (i.e. we have sq-avx512-spr.cpp, but not sq-avx512-vnni.cpp). What we would need to do is something like:

- utils/simd_levels.h — insert AVX512_VNNI in the enum between AVX512 and AVX512_SPR.
- utils/simd_levels.cpp — detect VNNI via CPUID leaf 7, ECX=0, EBX bit 11, gated inside the existing has_avx512f block, before the SPR check.
- impl/simd_dispatch.h — add a case SIMDLevel::AVX512_VNNI between the SPR and AVX512 cases (with [[fallthrough]]), and add it to get_simd_fallback so the chain is AVX512_SPR → AVX512_VNNI → AVX512 → AVX2 → NONE.
- New TU impl/scalar_quantizer/sq-avx512-vnni.cpp — mirror sq-avx512-spr.cpp: wrap in #ifdef COMPILE_SIMD_AVX512_VNNI, #define THE_LEVEL_TO_DISPATCH SIMDLevel::AVX512_VNNI, include sq-dispatch.h, and provide the DistanceComputerByte<Sim, SIMDLevel::AVX512_VNNI> specialization using _mm512_dpwssd_epi32 — unguarded, no #ifdef.
- add new file to CMakeLists.txt, taking the spr version as an example

Net behavior at runtime:
- Skylake-X → AVX512 tier → madd_epi16
- Cascade Lake / Ice Lake / Zen 4 → AVX512_VNNI tier → dpwssd (your change)
- SPR → goes to SPR version

baohongmin: @mnorris11 thanks for your suggestion. AVX512_SPR does support VNNI compilation, so only minor changes should be needed based on sq-avx512-spr.cpp.
https://github.com/facebookresearch/faiss/blob/b7cd013065b41d8f3b1a303591f0e64bec14eb70/faiss/CMakeLists.txt#L482

mnorris11: SPR CPUs support VNNI, but there are CPUs which support VNNI which are pre-SPR (cascade lake and cooper lake).  So this would also be useful for those machines. That is why a new file and SIMD level like `impl/scalar_quantizer/sq-avx512-vnni.cpp` could be useful. 

Alternatively to keep it simpler, we could just move it to `sq-avx512-spr.cpp` and keep the optimization there, and add the flags as you suggest. 

when thinking about it, maybe the second is best for now. Feel free to add the cmakelists flags.

baohongmin: The PR at [5173](https://github.com/facebookresearch/faiss/pull/5173) has already applied optimizations to the vector load in sq-avx512-spr.cpp similar to mine — it likely referenced my PR. The bottleneck of this function is the vector load. Testing shows that replacing with the _mm512_dpwssd_epi32 instruction brings less than 1% performance improvement, so there is no need to make the replacement.

mnorris11: Got it. We will not make the change in SPR then. Would you be interested in adding the `impl/scalar_quantizer/sq-avx512-vnni.cpp` kind of flow then, for those pre-SPR CPUs which support VNNI?

baohongmin: @mnorris11 
Without using VNNI instructions, we can avoid creating a new sq-avx512-vnni.cpp file and instead make minor modifications directly in sq-avx512.cpp to significantly improve performance, while remaining compatible with pre-SPR CPUs which support AVX-512 instructions. Please take a look at my updated code. What do you think of this approach? 
"By removing the _mm512_dpwssd_epi32 VNNI instruction and using native AVX-512 computation instead — since the performance bottleneck lies in the vector load, upgrading the original 128-bit load to a 256-bit load yields a significant performance improvement. Using VNNI instructions increases performance by less than 1%, so there is no need to introduce VNNI and cause extensive code changes."

mnorris11: Ah OK, so avoid VNNI entirely, just improve vector load part, which has no VNNI instructions at all. This works.

Edit: I have updated the title and description, feel free to update further though. I will try to merge today

mnorris11:
Merged, thanks for the contribution!

PR #5084, opened 2026-04-13
Add byte-packed IVF RaBitQ fast-scan layout for 4+ bit indexes

lyang24:
## Summary

  This PR adds an optional byte-packed multibit layout for `IndexIVFRaBitQFastScan` when `nb_bits >= 4`.

  The new layout stores `ex_code` as one byte per dimension and uses dedicated byte-packed multibit distance kernels in the FastScan refinement path. Lower-bit configurations keep the existing bit-packed layout, so `nbits=1/2/3` behavior is unchanged.

  ## Format / compatibility

  - existing fast-scan formats `Iwrn` and `Iwrf` continue to read through the old bit-packed path
  - new byte-packed indexes use a new `Iwrp` fourcc
  - `Iwrp` roundtrip serialization was verified

  ## Performance

  On Cohere 1M, single-threaded search, `nbits=4`:

  | nprobe | bit-packed | byte-packed | change |
  |-------:|-----------:|------------:|-------:|
  | 32 | 237 | 332 | +40% |
  | 64 | 169 | 234 | +38% |
  | 128 | 118 | 138 | +17% |

  Recall is unchanged:
  - `nprobe=32`: `0.8847`
  - `nprobe=64`: `0.9281`
  - `nprobe=128`: `0.9501`

  `nbits=2` is unchanged.

  ## Notes

  - the byte-packed path is only used for `nbits >= 4`
  - `distance_to_code()` is explicitly rejected for byte-packed IVF FastScan indexes, rather than silently using the wrong multibit decoder

### DIFF
--- a/faiss/IndexIVFRaBitQFastScan.cpp
+++ b/faiss/IndexIVFRaBitQFastScan.cpp
@@ -47,7 +47,8 @@ IndexIVFRaBitQFastScan::IndexIVFRaBitQFastScan(
         MetricType metric,
         int bbs_in,
         bool own_invlists_in,
-        uint8_t nb_bits)
+        uint8_t nb_bits,
+        bool byte_packed_excodes_in)
         : IndexIVFFastScan(
                   quantizer_in,
                   d_in,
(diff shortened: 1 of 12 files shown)


alexanderguzhva:
@lyang24 so, you're trading size for QPS, correct?

lyang24:
> @lyang24 so, you're trading size for QPS, correct?

yes i think a lot of this work is superseded by https://github.com/facebookresearch/faiss/pull/5097 5097 is better

PR #5103, opened 2026-04-15
RaBitQ: simplify compute_query_factors preprocessing hot path

lyang24:
This cleans up a few small but unnecessary pieces of work in `rabitq_utils::compute_query_factors()`.

 When the IVF residual path calls this function with `centroid == nullptr`, we were still recomputing the query norm even though the same value had already been computed a few lines earlier. We were also filling `rotated_q` by subtracting zero element-by-element instead of just copying the query, and carrying a few defensive empty-array branches that should not be hit for valid RaBitQ dimensions.

The math does not change here, and neither does the index format or search behavior. This is just a small preprocessing hot-path cleanup so `compute_query_factors()` does less obviously redundant work.


alibeklfc:
Merged, closing PR

PR #5123, opened 2026-04-19
Reuse per-thread query buffers in IVFRaBitQFastScan compute_LUT

lyang24:
The goal is to remove per-(query, probe) vector allocations from the float LUT path without changing distance math or the newer mainline interfaces.


alibeklfc:
Merged, closing the PR

PR #5128, opened 2026-04-20
introduce RVV platform and its ScalarQuantizer

alexanderguzhva:


faiss/impl/scalar_quantizer/sq-rvv.cpp
```diff
@@ -0,0 +1,311 @@
...
+#include <cmath>
+
+namespace faiss {
+
+namespace scalar_quantizer {
+
+/*************************************************************************
+ * Marker specializations.
+ *
+ * Unlike x86/NEON sq-*.cpp files that expose a fixed 8-wide / 16-wide codec
```
mdouze, line 26: cool that could be a model on how to implement SVE that is also variable-width

faiss/impl/scalar_quantizer/sq-rvv.cpp
```diff
@@ -0,0 +1,311 @@
...
+template <class Similarity>
+struct DistanceComputerByte<Similarity, SIMDLevel::RISCV_RVV>
+        : DistanceComputerByte<Similarity, SIMDLevel::NONE> {
+    using Base = DistanceComputerByte<Similarity, SIMDLevel::NONE>;
+    using Base::Base;
+};
+
+/*************************************************************************
+ * Fast path — QT_4bit_uniform + L2
+ *
```
mdouze, line 144: That's a cool optimization that I never came to implement. We should implement is as well for AVX and friends

mdouze (commented):
What about the feasibility of extending CI to this platform? 
IIUC it's quite popular in China.

faiss/impl/scalar_quantizer/sq-rvv.cpp
alexanderguzhva: @mdouze yep, same can be done for SVE

mnorris11:
> What about the feasibility of extending CI to this platform? IIUC it's quite popular in China.

@mdouze We would need a self-hosted runner as the Github Actions runners don't seem to support it.

mnorris11:
This got merged in https://github.com/facebookresearch/faiss/commit/a31dad31df54cb932b0152be98de7ea781d6b43c. I don't know why the PR didn't close.

PR #5135, opened 2026-04-21
Introduce early stop facilities for IVF

alexanderguzhva:
# Introduce early-stop controls for IVF search

This PR adds optional early-stop controls to IVF search through `SearchParametersIVF`.

IVF search looks in a sequence of buckets. Normally it keeps looking until it has checked all requested buckets. This PR adds a few ways to stop earlier:

- stop after enough candidate vectors have been checked,
- stop FastScan after enough buckets have been checked,
- stop range search after several nearby buckets produce no results.

All new fields default to disabled, so existing searches keep their current behavior unless callers opt in.

## New SearchParametersIVF fields

Default values:

| Field | Default | Effect |
|---|---:|---|
| `ensure_topk_full` | `false` | disabled; keep the existing hard `max_codes` behavior |
| `max_lists_num` | `0` | disabled; FastScan is bounded only by `nprobe` |
| `max_empty_result_buckets` | `0` | disabled; range search scans all requested probes |

`ensure_topk_full`

For k-NN search, `max_codes` limits how many candidate vectors are scanned. With `ensure_topk_full=false`, the search stops when the budget is reached.

With `ensure_topk_full=true`, very small budgets become less aggressive: `max_codes` is treated as at least `k` post-filter scans. This helps when `max_codes < k`, or when an `IDSelector` filters out many candidates before a distance is computed.

In the generic IVF path this is supported for `parallel_mode` 0 and 3. In FastScan k-NN it is supported by the per-query implementations 10 and 11.

`max_lists_num`

FastScan k-NN can also stop after visiting a limited number of inverted lists. `0` means disabled. With `ensure_topk_full`, this list budget is treated as at least `k` lists. This is only used by the FastScan per-query path (`search_implem_10`, also used for implementation 11).

When FastScan is left on `implem = 0`, setting an early-stop k-NN option makes the dispatcher choose the per-query implementation automatically. If an incompatible FastScan implementation is selected explicitly, Faiss throws instead of ignoring the option.

`max_empty_result_buckets`

This is for range search. It stops probing a query after that many consecutive probed lists add no in-radius results. The counter resets as soon as a list adds at least one result.

`0` means disabled. Setting this trades recall for less work: once nearby buckets repeatedly produce no hits, farther buckets are less likely to help.

The generic IVF range path supports this in `parallel_mode == 0`. FastScan range search also supports it through implementation 10. With `implem = 0`, FastScan selects implementation 10 automatically when this option is set; explicit incompatible implementations throw.

## Selector-aware scan accounting

This PR adds `InvertedListScannerStats`, returned by inverted-list scanner methods. It currently reports:

- `scan_cnt`: number of distances actually computed,
- `nheap_updates`: number of result updates.

The important behavior change is that `IndexIVFStats::ndis` now uses `scan_cnt`. With an `IDSelector`, candidates rejected by the selector are not counted as computed distances. Without a selector, this matches the previous count because every scanned code reaches distance computation.

This same post-filter count is used by `ensure_topk_full`, so a restrictive selector no longer consumes the search budget with candidates that were never eligible for the result heap.

## FastScan behavior

FastScan k-NN now applies `max_codes`, `max_lists_num`, and `ensure_topk_full` in the per-query implementation. The code keeps separate counters per query, so multi-query batches have independent budgets.

For `max_codes`, FastScan uses the result handler's scanned-row count. This means filtered-out rows do not consume the code budget.

FastScan range search rejects k-NN-only options (`max_codes`, `max_lists_num`, and `ensure_topk_full`). It accepts `max_empty_result_buckets` when the selected implementation can preserve per-query probe order.

## API exposure

The new fields are exposed through:

- C++ `SearchParametersIVF`,
- Python bindings,
- the C API.

Custom subclasses of `InvertedListScanner` or `BinaryInvertedListScanner` need to update their scanner method signatures to return `InvertedListScannerStats` instead of `size_t` or `void`.

## Tests

The tests cover:

- default parameters preserving baseline k-NN and range-search results,
- `ensure_topk_full` filling more result slots under tight budgets,
- selector-aware `IndexIVFStats::ndis`,
- range-search early exit and counter reset behavior,
- unsupported parallel modes and FastScan implementations,
- FastScan k-NN and range-search early-stop behavior,
- Python and C API field exposure.

### DIFF
--- a/faiss/IndexIVF.h
+++ b/faiss/IndexIVF.h
@@ -15,6 +15,7 @@
 #include <faiss/Clustering.h>
 #include <faiss/Index.h>
 #include <faiss/impl/IDSelector.h>
+#include <faiss/impl/InvertedListScannerStats.h>
 #include <faiss/impl/platform_macros.h>
 #include <faiss/invlists/DirectMap.h>
 #include <faiss/invlists/InvertedLists.h>
@@ -68,6 +69,25 @@ struct Level1Quantizer {
 struct SearchParametersIVF : SearchParameters {
     size_t nprobe = 1;    ///< number of probes at query time
     size_t max_codes = 0; ///< max nb of codes to visit to do a query
+
+    /// FastScan k-NN only: maximum number of inverted lists to visit.
+    /// 0 means unlimited, i.e. bounded only by nprobe. When set together
+    /// with max_codes, either budget may stop the scan. With
+    /// ensure_topk_full, this limit is treated as at least k lists.
+    size_t max_lists_num = 0;
+
+    /// For k-NN search, make small early-stop budgets less aggressive:
+    /// max_codes is treated as at least k post-IDSelector scans. Supported
+    /// by generic IVF in parallel_mode 0 and 3, and by FastScan k-NN
+    /// implementations 10 and 11.
+    bool ensure_topk_full = false;
+
+    /// Range-search only: stop after this many consecutive probed lists add
+    /// no in-radius results. 0 disables the heuristic. This trades recall
+    /// for less work. Supported in parallel_mode 0; FastScan range search
+    /// uses implementation 10 for this option.
+    size_t max_empty_result_buckets = 0;
+
     SearchParameters* quantizer_params = nullptr;
     /// context object to pass to InvertedLists
     void* inverted_list_context = nullptr;
@@ -512,9 +532,9 @@ struct InvertedListScanner {
      * @param distances  heap distances (size k)
      * @param labels     heap labels (size k)
      * @param k          heap size
-     * @return number of heap updates performed
+     * @return per-list scan statistics
      */
-    virtual size_t scan_codes(
+    virtual InvertedListScannerStats scan_codes(
             size_t n,
             const uint8_t* codes,
             const idx_t* ids,
@@ -523,18 +543,17 @@ struct InvertedListScanner {
             size_t k) const;
 
     // same as scan_codes, using an iterator
-    virtual size_t iterate_codes(
+    virtual InvertedListScannerStats iterate_codes(
             InvertedListsIterator* iterator,
             float* distances,
             idx_t* labels,
-            size_t k,
-            size_t& list_size) const;
+            size_t k) const;
 
     /** scan a set of codes, compute distances to current query and
      * update results if distances are below radius
      *
      * (default implementation fails) */
-    virtual void scan_codes_range(
+    virtual InvertedListScannerStats scan_codes_range(
             size_t n,
             const uint8_t* codes,
             const idx_t* ids,
@@ -542,14 +561,13 @@ struct InvertedListScanner {
             RangeQueryResult& result) const;
 
     // same as scan_codes_range, using an iterator
-    virtual void iterate_codes_range(
+    virtual InvertedListScannerStats iterate_codes_range(
             InvertedListsIterator* iterator,
             float radius,
-            RangeQueryResult& result,
-            size_t& list_size) const;
+            RangeQueryResult& result) const;
 
     // accumulate results with a ResultHandler
-    virtual size_t scan_codes(
+    virtual InvertedListScannerStats scan_codes(
             size_t n,
             const uint8_t* codes,
             const idx_t* ids,
--- a/faiss/impl/InvertedListScannerStats.h
+++ b/faiss/impl/InvertedListScannerStats.h
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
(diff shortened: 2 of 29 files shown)


alexanderguzhva:
@mdouze would you please comment the PR whenever you have time? Thanks

faiss/IndexIVF.h
```diff
@@ -68,6 +69,45 @@ struct Level1Quantizer {
 struct SearchParametersIVF : SearchParameters {
     size_t nprobe = 1;    ///< number of probes at query time
     size_t max_codes = 0; ///< max nb of codes to visit to do a query
+
+    /// Soft cap on the number of probed inverted lists. Only consulted by
```
mdouze: This explanation is confusing. Maybe summarize it as an algorithm?

faiss/impl/InvertedListScannerStats.h
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Returned by value (rather than mutated on the scanner) so the
+ * thread-safety contract on InvertedListScanner is preserved:
+ * concurrent scan_codes / scan_codes_range / iterate_codes /
+ * iterate_codes_range calls on a single scanner instance each
+ * receive their own stats object.
+ *
+ * Per-search totals across all probed lists accumulate into the
+ * caller-provided IndexIVFStats. Field-name correspondence:
+ *   InvertedListScannerStats::scan_cnt       -> IndexIVFStats::ndis
```
mdouze: Agreed that it is cleaner to return this via a stats object rather than via the global variable. 
We should have a principled way of committing it back to the global stats object. In IndexIVF?

tests/test_fast_scan_ivf.py
```diff
@@ -1019,3 +1019,167 @@ class TestRangeSearchImplem10(TestRangeSearchImplem12):
 @for_all_simd_levels
 class TestRangeSearchImplem110(TestRangeSearchImplem12):
     IMPLEM = 110
+
+
+class TestFastScanEarlyTerminationKnobs(unittest.TestCase):
+    """
+    Phase 0.5 wires three early-termination knobs on SearchParametersIVF
```
mdouze: what is phase 0.5 ?

tests/test_ivf_early_termination.cpp
```diff
@@ -0,0 +1,783 @@
...
+        // early_ids ⊆ full_ids
+        for (auto id : early_ids) {
+            EXPECT_TRUE(full_ids.count(id))
+                    << "early-exit returned id " << id
+                    << " not present in full-scan result for query " << q;
+        }
+    }
+}
+
+// Load-bearing test for the empty-bucket early-exit wiring at
```
mdouze: complicated explanation

mdouze (commented):
I am worried that this change may be too specific. We should think about expanding the ResultHandler with a way to break out of the test. 

https://github.com/facebookresearch/faiss/blob/c3460c980be78afac11f6dc1f88b170fd7e70693/faiss/impl/ResultHandler.h#L28 

Would it be possible to make the comments more intelligible? Ask the coding assistant something like "explain as if I'm 5"

faiss/IndexIVF.h
alexanderguzhva: Let me do that

faiss/impl/InvertedListScannerStats.h
alexanderguzhva: technically, yes. But I wasn't sure I needed to commit these changes, because I wanted to get an initial review first. Adding this to global ivf stats index is trivial.

tests/test_fast_scan_ivf.py
alexanderguzhva: I'll remove redundant comments

tests/test_ivf_early_termination.cpp
alexanderguzhva: I'll take a look

alexanderguzhva:
@mdouze I have updated both the description and the comments

alexanderguzhva:
> I am worried that this change may be too specific. We should think about expanding the ResultHandler with a way to break out of the test.
> 
> https://github.com/facebookresearch/faiss/blob/c3460c980be78afac11f6dc1f88b170fd7e70693/faiss/impl/ResultHandler.h#L28

@mdouze  The problem is in the inconsistency of `InvertedListScanner` ABI. There are custom `InvertedListScanner` variants (`IndexIVFPQFastScanScanner`, `IndexRaBitQFastScanScanner` and for `IndexIVFSpectralHash`) that directly override `scan_codes()` version that does not rely on `ResultHandler`. The problem with FastScan implementations is that they rely on knowing topk number, because it is needed to construct a temporary SIMD heap. So, in this case a `ResultHandler&` can be only `HeapResultHandler&`... etc.
 
I will try to reorganize the code as a new PR, although I still believe that the way I suggested in this PR is the most rational and clear one.

alexanderguzhva:
#5160 as an alternative

mnorris11:
Merged https://github.com/facebookresearch/faiss/pull/5160/ instead

PR #5148, opened 2026-04-27
Fix ODR violation in ScannerMixIn by adding SIMDLevel template parameter

mulugetam:
While benchmarking FastScan on an AVX-512 SPR system with a DD (dynamic dispatch) build, we observed that the generated code was executing scalar (NONE-level) instructions instead of the expected AVX2/AVX-512 kernels. 

**Root cause**

`ScannerMixIn<Handler>` is defined in `dispatching.h`, which is compiled once per SIMD level in DD mode. Each translation unit defines `THE_LEVEL_TO_DISPATCH` to a different `SIMDLevel` value (NONE, AVX2, AVX512, AVX512_SPR), and the class uses this macro in its method bodies to select SIMD code paths.

However, since `ScannerMixIn` is only templated on `Handler`, every TU produces a class with the same mangled name — regardless of the SIMD level it was compiled for. The linker silently picks one, and in practice it chose the NONE version, making all FastScan searches fall back to scalar code.

**Fix**

Add a non-type template parameter `SIMDLevel SL = THE_LEVEL_TO_DISPATCH` to `ScannerMixIn`. This makes each TU's instantiation a distinct template specialization with a unique mangled name, so the linker preserves all versions and runtime dispatch selects the correct one.

**Impact**

- DD builds now correctly dispatch FastScan to the highest available SIMD level at runtime
- No effect on static (non-DD) builds, where only one SIMD level is compiled


mulugetam:
@subhadeepkaran @mdouze  Could you take a look when you have time? Thanks!

PR #5156, opened 2026-04-29
Introduce RVV

alexanderguzhva:
I've verified the successful compilation via docker + qemu combination.

Also, this PR introduces all the needed overrides, which rely on default implementation, but later can be populated with the vectorized code.
Also, fixes a typo in SQ4U.

@algoriddle I would appreciate if you could take a look whenever you have time. Thanks!

### DIFF
--- a/faiss/impl/simd_dispatch.h
+++ b/faiss/impl/simd_dispatch.h
@@ -32,9 +32,9 @@ constexpr int AVAILABLE_SIMD_LEVELS_NONE = (1 << int(SIMDLevel::NONE));
 constexpr int AVAILABLE_SIMD_LEVELS_AVX2_NEON = AVAILABLE_SIMD_LEVELS_NONE |
         (1 << int(SIMDLevel::AVX2)) | (1 << int(SIMDLevel::ARM_NEON));
 
-// A0: same + AVX512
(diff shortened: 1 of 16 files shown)


faiss/impl/simd_dispatch.h
mnorris11, line 159: A0 is the default, but only a few functions are specialized for RVV, so we can move it out.

Matthijs recommended `with_selected_simd_levels<AVAILABLE_SIMD_LEVELS_A0 | (1 << int(SIMDLevel::RISCV_RVV))>` here.

alexanderguzhva: @mnorris11 I'd like to keep it as it, bcz A0 is the first, basic level of SIMD for a platform, as far as I understand. With hopes that RISCV users will correct us anyways if things are wrong.

alexanderguzhva: Also, I hope that RISCV guys will populate various SIMD things for RaBitQ etc. as well.

mdouze: The `with_selected_simd_levels` was introduced precisely for if some functions are not implemented with some SIMD levels. 
Do you think the code will be optimized with RVV anytime soon?

PR #5165, opened 2026-04-30
c_api: fix IndexShards own_indices getter/setter name mismatch

Binozo:
`IndexShards_c.cpp` defines:

```cpp
DEFINE_GETTER(IndexShards, int, own_indices)
DEFINE_SETTER(IndexShards, int, own_indices)
```

while `IndexShards_c.h` defines
```cpp
FAISS_DECLARE_GETTER_SETTER(IndexShards, int, own_fields)
```

this mismatch causes neither `faiss_IndexShards_set_own_indices` and `faiss_IndexShards_set_own_fields` to be exported in the c_api. This patch completes the migration to `indices` instead of `fields` in the c_api.


Enet4 (approved):
Yup, sounds about right. It's named `own_indices`, as declared in `ThreadedIndex`.

PR #5172, opened 2026-05-01
Fix lints associated with early stop facilities for IVF (#5172)

mnorris11:
Summary:

Follow up to D103044128
Fix some lint issues and one stats-tracking issue (and add test for it) in IVF early-stop code:

- **ResultHandler.h**: `RangeResultHandler::add_result` now returns `true` when a result is actually added, fixing `nheap_updates` stat tracking through the `scan_codes_range` / `expanded_scanners.h` path. Previously always returned `false`, making `nheap_updates` a perpetual no-op for range search via the default scanner. This is inconsistent with `iterate_codes_range` and `IndexIVFSpectralHash_impl.h`, which track correctly.
- **IndexIVF.cpp**: Move `effective_max_codes` declaration to its only use site, eliminating a dead-store lint warning.
- **IndexIVFPQFastScan.cpp**: Remove unused `#include <faiss/utils/Heap.h>`.
- **test_fast_scan_ivf.py**: Reformat trailing comment for readability.
- **test_ivf_early_termination.cpp**: Add unit test verifying `RangeResultHandler::add_result` return value. Register test in BUCK (was CMake-only).

Differential Revision: D103438123

### DIFF
--- a/faiss/IndexIVF.cpp
+++ b/faiss/IndexIVF.cpp
@@ -421,7 +421,6 @@ void IndexIVF::search_preassigned(
     const idx_t unlimited_list_size = std::numeric_limits<idx_t>::max();
     idx_t cur_max_codes = params ? params->max_codes : this->max_codes;
     const bool ensure_topk_full = params ? params->ensure_topk_full : false;
-    idx_t effective_max_codes = cur_max_codes;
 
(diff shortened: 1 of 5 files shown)


mnorris11:
cc @alexanderguzhva small lint fixes on top of your change

PR #5173, opened 2026-05-01
Add Sapphire Rapids optimizations for ScalarQuantizer (L2, IP)

mulugetam:
This PR specializes the byte-vector distance path for `AVX512_SPR` on `QT_8bit_direct` and `QT_8bit_direct_signed`, achieving speedups of up to **2.3x** on distance benchmarks and up to **1.21x** on search (`IndexIVFScalarQuantizer`) compared to the existing SPR implementation.

**Inner product:** A 64-byte VNNI loop using `_mm512_dpbusd_epi32` replaces the 16-byte `cvtepu8_epi32` + `mullo_epi32` path. For the unsigned×unsigned case, the operand is biased by −128; a closed-form correction using `sum(a)` and `sum(b)` (accumulated via `_mm512_sad_epu8`) restores the exact result. The signed variant applies the same correction terms.

**L2 distance:** A 64-byte widened-multiply-add loop via `_mm512_cvtepu8_epi16` + `_mm512_madd_epi16` replaces the narrower path. The signed variant is bit-exact because the −128 bias cancels in the difference.

*Improves upon and supersedes #5067.*

### Distance benchmark

```
bench_scalar_quantizer_distance --d=<128|256|768> --n=2000 --iterations=20
```

| Quantizer | d | Baseline (ms) | This PR (ms) | Speedup |
|---|--:|--:|--:|--:|
| `QT_8bit_direct` | 128 | 40.8 | 22.9 | **1.78×** |
| `QT_8bit_direct_signed` | 128 | 35.0 | 23.0 | **1.52×** |
| `QT_8bit_direct` | 256 | 75.3 | 35.7 | **2.11×** |
| `QT_8bit_direct_signed` | 256 | 69.1 | 35.7 | **1.94×** |
| `QT_8bit_direct` | 768 | 215.5 | 92.3 | **2.33×** |
| `QT_8bit_direct_signed` | 768 | 206.4 | 91.0 | **2.27×** |

Raw performance data: https://gist.github.com/mulugetam/72c0960e47bc640f99aa346f363e56fe

### End-to-end search (`IndexIVFScalarQuantizer`)

```
python benchs/bench_scalar_quantizer.py
```

| Range stat | `QT_8bit_direct` | `QT_8bit_direct_signed` |
|---|--:|--:|
| RS_minmax | 1.10× | 1.12× |
| RS_minmax | 1.09× | 1.15× |
| RS_minmax | 1.14× | 1.13× |
| RS_minmax | 1.15× | 1.07× |
| RS_minmax | 1.06× | 1.17× |
| RS_minmax | 1.09× | 1.02× |
| RS_minmax | 1.09× | 1.11× |
| RS_meanstd | 1.21× | 1.13× |
| RS_meanstd | 1.14× | 1.06× |
| RS_meanstd | 1.11× | 1.07× |
| RS_meanstd | 1.12× | 1.07× |
| RS_meanstd | 1.15× | 1.10× |
| RS_meanstd | 1.13× | 1.04× |
| RS_meanstd | 1.04× | 1.07× |
| RS_quantiles | 1.11× | 1.03× |
| RS_quantiles | 1.12× | 1.11× |
| RS_quantiles | 1.11× | 0.98× |
| RS_quantiles | 1.14× | 1.17× |
| RS_optim | 1.06× | 1.07× |

Raw performance data: https://gist.github.com/mulugetam/632e2e08c9358b2184cbaa3397a6c73f

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -40,6 +40,7 @@ set(FAISS_SIMD_AVX512_SRC
 # flag. Not compiled into faiss_avx512 — that target stops at the
 # baseline AVX-512F/CD/VL/DQ/BW feature set.
 set(FAISS_SIMD_AVX512_SPR_SRC
+  impl/scalar_quantizer/sq-avx512-spr.cpp
   utils/simd_impl/rabitq_avx512_spr.cpp
   utils/hamming_distance/hamming_avx512_spr.cpp
 )
@@ -471,7 +472,7 @@ endif()
 if(NOT WIN32)
   # Architecture mode to support AVX512 extensions available since Intel(R) Sapphire Rapids.
   # Ref: https://networkbuilders.intel.com/solutionslibrary/intel-avx-512-fp16-instruction-set-for-intel-xeon-processor-based-products-technology-guide
-  target_compile_options(faiss_avx512_spr PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mavx512vpopcntdq -mpopcnt -mavx512fp16 -mavx512bf16 ${FAISS_BMI2_FLAGS}>)
+  target_compile_options(faiss_avx512_spr PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mavx512f -mavx512cd -mavx512vl -mavx512dq -mavx512bw -mavx512vpopcntdq -mpopcnt -mavx512vnni -mavx512fp16 -mavx512bf16 ${FAISS_BMI2_FLAGS}>)
 else()
   target_compile_options(faiss_avx512_spr PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/arch:AVX512>)
   # we need bigobj for the swig wrapper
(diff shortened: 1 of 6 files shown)


mdouze:
Thanks for the PR. 
Do I understand correctly that the results are not exactly the same since this is based on an integer-integer comparison instead of integer-float? 
In fact this is a tradeoff that we may want to apply to other scalar quantizers as well, especially the QT_x_uniform ones.

mulugetam:
@mdouze Thanks for the review. 

The results are bit-exact for `QT_8bit_direct`: both the existing AVX512 code and this PR operate in the integer domain via `DistanceComputerByte`, which truncates the query to `uint8` in `set_query()`. Thiss PR simply widens the loop from 16 to 64 bytes per iteration using VNNI, with an algebraically exact bias correction for the unsigned IP case.

For `QT_8bit_direct_signed`, my reading of the code is that it has no `DistanceComputerByte` variant for the signed type and it just falls through to the float-domain `DCTemplate<Quantizer8bitDirectSigned>` path, which preserves full float precision in the query. The new `DistanceComputerByteSigned` truncates the query to integer (`uint8_t(int(x[i]) + 128)`), losing the fractional part. As stated above, this truncation loss already exists for `QT_8bit_direct`, but it's new for `QT_8bit_direct_signed`.

I can document this as a known tradeoff (speed vs. precision for `QT_8bit_direct_signed`). Alternatively, I could only enable it for the symmetric `compute_code_distance` path where both inputs are already integer (but this means losing the benefit of VNNI for `query_to_code` path).

Regarding `QT_x_uniform`: the VNNI byte-domain technique doesn't directly apply there since, from my reading of the code, those quantizers require scale/offset denormalization.

Quick question: are the DD changes complete, or still in progress? I'd like to re-review and rebase all my other pending PRs (I believe there are 9) once they land. Thanks!

mdouze:
I think that the QT_8bit_direct_signed can be performed in the integer domain, that's the initial purpose of 8bit_direct_x
Yes the DD is considered complete since yesterday. 
Thanks for all the SIMD optimization efforts! We will try to ingest them as soon as possible.

mulugetam:
@mdouze Rebased with fixes.

mulugetam:
@mnorris11 Conflicts resolved and rebased.

mulugetam:
@mnorris11 CMakeLists.txt conflict resolved and rebased.

PR #5177, opened 2026-05-04
Fix Windows conda build: pytorch package name and setup-miniconda activation

mdouze:
## Problem

Two issues causing Windows conda CI failures:

1. **`pytorch-cpu >=2.7` is only packaged for `linux-64`**, not `win-64`. The conda test environment could not be created on Windows, causing `conda activate "test"` to fail with `EnvironmentNameNotFound`.

2. **`setup-miniconda@v3` default `activate-environment: test`** creates a named `test` conda environment and writes `conda activate "test"` to shell profiles (`.bash_profile`). conda-build on Windows runs test scripts via Git Bash with `--login`, which sources `.bash_profile` and runs `conda activate "test"` — but that environment does not exist in conda-build's isolated context, causing `EnvironmentNameNotFound`.

## Changes

- `conda/faiss/meta.yaml`: use `pytorch-cpu >=2.7  # [not win]` and `pytorch >=2.7  # [win]` in the `faiss-cpu` test requirements (pytorch-cpu is Linux-only; win-64 uses pytorch)
- `.github/actions/build_conda/action.yml`: set `activate-environment: 'base'` in the `setup-miniconda@v3` step. This avoids creating a throwaway `test` env and instead writes `conda activate base` to shell profiles — the base env always exists, so no `EnvironmentNameNotFound`


mdouze:
obsolete

PR #5189, opened 2026-05-07
Support user provided blas library

zhiqiang-hhhh:
Hello Faiss community,

  I am using Faiss as the vector search engine in Apache Doris, where Faiss is integrated as a git submodule. In our
build, both OpenBLAS and Faiss are compiled from source, and I found that it is currently inconvenient to pass a
BLAS/LAPACK implementation maintained by the parent project into Faiss.

  I modified `faiss/CMakeLists.txt` so that a superbuild can directly provide `BLAS_LIBRARIES` and `LAPACK_LIBRARIES`.
When these variables are already set by the parent project, Faiss uses them directly instead of continuing with its own
BLAS/LAPACK autodetection logic.

  This helps in cases where:
  - OpenBLAS is built in the same build system or provided by the parent project
  - the parent project wants deterministic dependency selection
  - `find_package(BLAS)` / `find_package(LAPACK)` may pick a different implementation than the one intended by the
integrator

  The change does not affect the default standalone build behavior: if `BLAS_LIBRARIES` and `LAPACK_LIBRARIES` are not
provided, Faiss still follows the existing logic.


RoanHeNaN:
Hi @mnorris11 , it looks like the Meta CodeSync check says “PR has not been imported”. Is there anything I need to do from my side, or is this waiting for an internal import/review step?

mnorris11:
> Hi @mnorris11 , it looks like the Meta CodeSync check says “PR has not been imported”. Is there anything I need to do from my side, or is this waiting for an internal import/review step?

Hi @zhiqiang-hhhh , nothing needed on your side. We are busy trying to ensure the SIMD dynamic dispatch rollout goes through.

PR #5200, opened 2026-05-08
Fix Panorama segfault

aknayar:
In the current code you can segfault Panorama indexes by loading a dataset with 960 dims and setting `nlevels` to 128.

In the case where `nlevels` does not evenly divide `d`, Panorama currently calculates `level_width = ceil(d / nlevels)`. We cannot change this behavior due to backwards compatibility. However, for our case of 960 dims and 128 levels, this would mean each level is 8 dims. Thus, the last 8 levels have 0 dims each. Since we calculate `actual_level_width` using `size_t` params, it overflows and produces 8 when it should produce 0, leading to out-of-bounds memory accesses.

The fix is to simply truncate `nlevels` in the `Panorama` object when this would happen. A debug log is emitted to the user when this happens. In the aforementioned example, the user would see `WARNING truncating nlevels from 128 to 120`.

This PR also fixes a similar latent bug in `IndexFlatPanorama::reconstruct` where we would perform out-of-bounds memory accesses when `nlevels` doesn't evenly divide `d`.

Existing test cases are strengthened to ensure bugs are fixed and to avoid any future regressions.


faiss/IndexHNSW.cpp
```diff
@@ -707,7 +707,10 @@ IndexHNSWFlat::IndexHNSWFlat(int d_in, int M, MetricType metric)
  **************************************************************/
 
 IndexHNSWFlatPanorama::IndexHNSWFlatPanorama()
-        : IndexHNSWFlat(), cum_sums(), pano(0, 1, 1), num_panorama_levels(0) {}
+        : IndexHNSWFlat(),
+          cum_sums(),
+          pano(sizeof(float), 1, 1),
```
aknayar, line 712: Needed so we don't have a division by zero when we do `size_t n_real_levels = d / level_width_floats`.

aknayar:
@mnorris11 Any advice on these OOMing tests? Is it something on my end?

Edit: My new tests were too beefy

mnorris11:
> In the case where nlevels does not evenly divide d, Panorama currently calculates level_width = ceil(d / nlevels). We cannot change this behavior due to backwards compatibility. 

Thanks for the fix, can you explain this part more?

aknayar:
@mnorris11 Yup! Consider the case where d = 960 and nlevels = 128. Ideally we would have `level_width = floor(960 / 128)` = 7 dims per level. This would allow us to have 127 levels at `level_width` 7 and the 128th level would take the remaining (960 - 127 * 7 = 71) dims, satisfying the user's request for 128 levels.

However, the existing logic in the codebase is `level_width = ceil(d / nlevels)` instead of `floor`. This means `level_width` = 8 for this configuration. Since we transpose the storage layout to be level-major, we need to maintain this `level_width` of 8 for backwards compatibility (we cannot switch `level_width` to use floor). Loading an index which was serialized with a `level_width` of 8 into an index that expects a `level_width` of 7 will lead to bogus results.

Thus, if the user requests a configuration like d = 960, nlevels = 128, we have to truncate nlevels to be something that can actually accommodate the available dims given a `level_width` of `ceil(d / nlevels)`. If backwards compatibility weren't a concern, we would rewrite this logic entirely.

PR #5202, opened 2026-05-08
Add MetalIndexIVFFlat with IVF scan/merge kernels and expanded top-k support

Evandabest:
## Summary                                                                                                                                                                                                       
Adds IVFFlat index support to the Metal GPU backend
                                                                                                                                                                                                                 
- Add `MetalIndexIVFFlat` with full train/add/search/reset/copyFrom/copyTo support                                                                                                                               
- Add IVF scan/merge MSL kernels (`ivf_scan_list`, `ivf_scan_list_small`, `ivf_scan_list_interleaved`, `ivf_merge_lists`) to `MetalDistance.metal`                                                               
- Add `MetalDistance.h/.mm` orchestration layer for IVF search dispatch (coarse quantization + scan + merge in a single command buffer)                                                                          
- Add `compute_norms` and `l2_with_norms` kernels for efficient L2 coarse quantization                                                                                                                           
- Expand top-k support from k <= 256 to k <= 2048 (`topk_threadgroup_{512,1024,2048}`)                                                                                                                           
- Raise dimension limit from d <= 512 to d <= 2048 (supports 768, 1024, 1536 embeddings)                                                                                                                         
- Update `MetalCloner` to support IVFFlat in `index_cpu_to_metal_gpu` / `index_metal_gpu_to_cpu`                                                                                                                 
- Rename `nil` parameter to `n_il` in `InvertedLists.h/.cpp` to fix Objective-C++ keyword conflict
- Add Python bindings (MetalPythonBridge) and SWIG integration for Metal backend
- Add Python test suite (10 tests covering Flat and IVFFlat via Python bindings)                                                                                                                    
                                                                                                                                                                                                                 
## Changes                                                                                                                                                                                                       
- **New:** `MetalIndexIVFFlat.h/.mm` - IVFFlat index class (train, add, search, reset, cloner support)                                                                                                           
- **New:** `impl/MetalIVFFlat.h/.mm` - GPU-resident IVF list storage with segment allocator                                                                                                                      
- **New:** `MetalDistance.h/.mm` - IVF search dispatch (scan + merge orchestration)                                                                                                                              
- **New:** test/TestMetalIndexIVFFlat.cpp - 4 C++ tests (L2, IP, reset, cpu to gpu round trip)                                                                                                             
- **Modified:** `MetalDistance.metal` - IVF kernels, compute_norms, l2_with_norms, topk 512/1024/2048                                                                                                            
- **Modified:** `MetalKernels.h/.mm` - IVF encode methods, expanded topk variants, kMaxK=2048                                                                                                                    
- **Modified:** `MetalIndex.h` - added IVF config fields (indicesOptions, interleavedLayout, useFloat16CoarseQuantizer)                                                                                          
- **Modified:** `MetalCloner.mm` - IVFFlat support in cpu <-> gpu cloning                                                                                                                                            
- **Modified:** `CMakeLists.txt` - added new source files                                                                                                                                                        
- **Modified:** `.github/actions/build_cmake/action.yml` - added TestMetalIndexIVFFlat to CI                                                                                                                     
- **Modified:** `faiss/invlists/InvertedLists.h/.cpp` - renamed `nil` parameter to `n_il` (Objective-C++ keyword conflict)
- **New:** MetalPythonBridge.h/.mm - C++ bridge API for SWIG (resources, cloning, profiler, sync)
- **New:** test/test_metal_python.py - 10 Python tests (Flat L2/IP, IVFFlat L2/IP, reset, roundtrip, profiler, resources)
- **Modified:** faiss/python/swigfaiss.swig - added FAISS_ENABLE_METAL section for Metal bindings
- **Modified:** faiss/python/CMakeLists.txt - added FAISS_ENABLE_METAL compile def and Metal framework linking

## Differences from CUDA IVFFlat                                                                                                                                                                                 
 
**Training:** Delegates to CPU (`IndexIVFFlat::train`). CUDA can train on GPU via k-means or cuVS. GPU training is possible with Metal but low priority since training is a one-time cost.                       
                     
**Add path:** Coarse quantization runs on CPU, then vectors are copied to GPU storage. CUDA does both on GPU. Moving coarse assignment to Metal is doable using the existing flat search kernels; planned for a future optimization. 
                                                                                                                                                                                                                 
**Reconstruct:** Reads from CPU host mirror rather than GPU buffers. On Apple Silicon with unified memory, CPU and GPU see the same physical memory, so this is functionally equivalent.                         
 
**IVF list storage:** Uses a single contiguous buffer with a segment allocator, rather than CUDA's per-list `DeviceVector` allocations. This avoids managing thousands of small `MTLBuffer` objects and reduces fragmentation; a better fit for Apple Silicon's unified memory pool.
                                                                                                                                                                                                                 
## Notes on other changes
**`nil` → `n_il` rename in `InvertedLists.h/.cpp`:** ObjC++ treats `nil` as a keyword (equivalent to `nullptr` for Objective-C objects). Any `.mm` file that transitively includes `InvertedLists.h` fails to compile because the parameter name `int nil` conflicts with the keyword. This is a declaration only rename, no behavior, ABI, or API change. Does not affect CUDA, CPU, or any other build.
                                                                                                                                                                                                                   
**`centroidsAreFP16` parameter in `runMetalIVFFlatFullSearch`:** The Metal-backend branch passes a `useFloat16CoarseQuantizer` config flag through the search dispatch chain. CUDA handles this differently, FP16 centroid storage is a property of the coarse quantizer object (`GpuIndexFlatConfig::useFloat16`), not a function parameter. The parameter is currently a no-op but is included to maintain the correct function signature and avoid a parameter alignment bug where a trailing `bool` would be misinterpreted as `waitForCompletion`.   
                                                           
## Build and test                                                                                                                                                                                                
```bash                                                                                                                                                                                                          
cmake -B build \
  -DFAISS_ENABLE_GPU=OFF \
  -DFAISS_ENABLE_PYTHON=ON \
  -DBUILD_TESTING=ON \
  -DCMAKE_BUILD_TYPE=Release \
  .
cmake --build build --target faiss faiss_metal swigfaiss TestMetalIndexFlat TestMetalIndexIVFFlat -j$(sysctl -n hw.logicalcpu)
cd build && ctest -R TestMetalIndex --output-on-failure 

#Python Tests
pip install .
pytest faiss/gpu_metal/test/test_metal_python.py -v
```


mdouze:
Please add tests in Python to exercise the whole call stack.

Evandabest:
@mdouze I have added python tests that excercise the whole call stack. I also updated the CI to enable Python for Metal and run the tests.

mnorris11:
Hi @Evandabest , I double checked the review below and it seems like we could update these 3 things. The first one is probably the most important, 2nd one is most debatable. What do you think?

## Code Review — MetalIndexIVFFlat

### 1. Double coarse assignment in `add()` / `add_with_ids()`

**`MetalIndexIVFFlat.mm:588`** and **`MetalIndexIVFFlat.mm:628`**

`quantizer->assign()` at line 588 computes list assignments, but `cpuIndex_->add()` at line 596 calls `IndexIVF::add_with_ids()` which calls `quantizer->assign()` again internally (`IndexIVF.cpp:195`). The quantizer runs twice per batch with identical inputs. Same pattern in `add_with_ids()` at lines 628/630.

Fix: call `cpuIndex_->add_core()` instead, which accepts pre-computed assignments and skips the internal `assign()`:

```cpp
// In add():
std::vector<idx_t> list_nos(n);
cpuIndex_->quantizer->assign(n, x, list_nos.data());

cpuIndex_->add_core(n, x, nullptr, list_nos.data());
ntotal = cpuIndex_->ntotal;
```

```cpp
// In add_with_ids():
std::vector<idx_t> list_nos(n);
cpuIndex_->quantizer->assign(n, x, list_nos.data());

cpuIndex_->add_core(n, x, xids, list_nos.data());
ntotal = cpuIndex_->ntotal;
```

Neither CPU `IndexIVF` nor CUDA `GpuIndexIVF` compute assignments twice — CPU calls `assign()` once in `add_with_ids()` and passes the result to `add_core()`; CUDA calls `searchCoarseQuantizer_()` once in `IVFBase::addVectors()`.

---

### 2. Latent pointer invalidation in `search_preassigned()` chunked path

**`MetalIndexIVFFlat.mm:1702-1719`**

`fullCoarse` at line 1702 points into `searchCoarseBuf_` contents. At line 1718, `ensureSearchBuf_(searchCoarseBuf_, ...)` may reallocate that buffer, invalidating `fullCoarse`. On the next loop iteration at line 1713, `fullCoarse + srcBase` would be a dangling pointer.

Currently safe because `coarseChunkBytes <= searchCoarseCap_` (chunk is always ≤ full buffer), but one refactor away from a use-after-free. The equivalent path in `search()` (line 1009) avoids this by allocating separate per-pass buffers.

Fix: copy to a local vector before the loop. Replace lines 1701-1703:

```cpp
std::vector<int32_t> fullCoarseVec(
        reinterpret_cast<const int32_t*>([searchCoarseBuf_ contents]),
        reinterpret_cast<const int32_t*>([searchCoarseBuf_ contents])
                + (size_t)qCount * nprobe);

std::vector<int32_t> coarseChunk;
const int32_t* fullCoarse = fullCoarseVec.data();
```

The CUDA path avoids this class of bug entirely — coarse assignments are allocated as a single `DeviceTensor`, written once, and consumed read-only.

---

### 3. O(k) insertion sort for top-k merge

**`MetalIndexIVFFlat.mm:367-402`**

`insertTopKCandidate` does a linear scan + shift for each candidate, making each insertion O(k). Called k × chunks × queries times in the merge paths (lines 1213, 1305, 1337 in `search()`; lines 1857, 1949 in `search_preassigned()`), giving O(chunks × queries × k²) total. With k=1024 this is significant.

CPU Faiss uses a binary heap (O(log k) per insertion) via `HeapArray::addn_with_ids`. CUDA does the merge entirely on GPU via `BlockSelect` bitonic merge networks.

Fix: replace with a max-heap (L2) / min-heap (IP) and heap-sort the final result:

```cpp
inline void heapInsertIfBetter(
        float candDist, int64_t candIdx, bool isL2,
        float* heapDist, int64_t* heapIdx, int k) {
    if (candIdx < 0 || k <= 0) return;
    bool dominated = isL2 ? (candDist >= heapDist[0])
                          : (candDist <= heapDist[0]);
    if (dominated) return;
    heapDist[0] = candDist;
    heapIdx[0] = candIdx;
    int i = 0;
    while (true) {
        int left = 2*i+1, right = 2*i+2, worst = i;
        if (isL2) {
            if (left < k && heapDist[left] > heapDist[worst]) worst = left;
            if (right < k && heapDist[right] > heapDist[worst]) worst = right;
        } else {
            if (left < k && heapDist[left] < heapDist[worst]) worst = left;
            if (right < k && heapDist[right] < heapDist[worst]) worst = right;
        }
        if (worst == i) break;
        std::swap(heapDist[i], heapDist[worst]);
        std::swap(heapIdx[i], heapIdx[worst]);
        i = worst;
    }
}
```

Heap-sort the result before writing to `distances`/`labels`. Drops the merge from O(chunks × k²) to O(chunks × k log k) per query.

Evandabest:
Thanks for the code review @mnorris11. First one was a good catch, I will switch both `add()` and `add_with_ids()` to use `add_core()` with the precomputed assignments so the quantizer only runs once. As for the pointer invalidation issue, although it is currently safe since the chunk is always ≤ than the full buffer, copying the coarse assignments into a local vector before the loop is the right call. I will replace the insertion sort with a heap based merge. I am planning to add a Metal merge kernel so that this whole process is done on GPU in a future optimization pass.

mnorris11:
We are merging this now @Evandabest , thanks for the contribution!

The landed commit will have a slight change: we added a new string to the compile_options in MetalKernels.mm, just so a Python user knows which they are using.

PR #5211, opened 2026-05-13
Fix IDSelectorBitmap conversion to cuVS bitset (#5211)

mnorris11:
Summary:

`convert_to_bitset_bitmap` in `CuvsFilterConvert.cu` assumed that `IDSelectorBitmap.n` was the bit/vector count, but it can also be the byte count of the bitmap array (which is how the Python wrapper constructs it via `faiss.IDSelectorBitmap(numpy_array)`). This caused two bugs:

1. The `RAFT_EXPECTS(bitset.size() == n)` check failed when `n` was the byte count (e.g., 6.25M bytes vs 50M bits), throwing a `raft::logic_error` that propagated up and caused filtered search to silently fall back to unfiltered results.

2. `n_elements = (selector.n + 7) / 8` divided the byte count by 8 again, copying only 1/8 of the bitmap data when `n` was already the byte count.

The fix handles both conventions by computing the actual number of bytes to copy as `min(n, ceil(bitset_size / 8))`. This works regardless of whether `n` represents bytes or bits:
- Python wrapper (n = byte count): copies all n bytes
- C++ convention (n = bit count): copies ceil(bit_count / 8) bytes (the minimum needed)

Reviewed By: pankajsingh88

Differential Revision: D104935985

### DIFF
--- a/faiss/gpu/test/TestGpuFilterConvert.cu
+++ b/faiss/gpu/test/TestGpuFilterConvert.cu
@@ -78,12 +78,12 @@ void run_complex() {
 
     auto or_selector = faiss::IDSelectorOr(&range_selector, &array_selector);
 
-    auto bitmap_faiss_cpu = std::vector<uint8_t>((spec.bitset_len + 8) / 8);
+    auto bitmap_faiss_cpu = std::vector<uint8_t>((spec.bitset_len + 7) / 8);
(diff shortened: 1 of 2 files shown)


mnorris11:
@tarang-jain  @lowener  Does this change look right?

PR #5212, opened 2026-05-14
ci: declare workflow-scope permissions on nightly, publish-docs, update-doxygen

arpitjain099:
Three workflows currently don't declare a `permissions:` block, so the workflow `GITHUB_TOKEN` falls back to whatever the repo default grants. This patch pins each one to the minimum it actually needs:

- `.github/workflows/nightly.yml` -- `contents: read`. The job only checks out, invokes the local `./.github/actions/build_conda` composite action, and uses `ANACONDA_API_TOKEN` (an external secret) for the upload. No GitHub API write.
- `.github/workflows/publish-docs.yml` -- `contents: write`. The job re-builds Sphinx HTML and pushes a commit to `gh-pages` via `$GITHUB_TOKEN`.
- `.github/workflows/update-doxygen.yml` -- `contents: write`. Same shape, pushes regenerated Doxygen XML to `gh-pages`.

Style matches `autoclose.yml`, `build-pip.yml`, and `retry_build.yml`, all of which already carry per-job `permissions:` blocks.

`build.yml` is deliberately left out. It's a small caller that fans out to three reusable workflows (`build-pull-request.yml`, `build-release.yml`, `build-pip.yml`). Adding a `permissions:` block to the caller intersects with the callees' grants (`build-pip.yml` needs `id-token: write` + `contents: read`), so it deserves a separate, more careful change.

No behavioural impact for the three workflows in this PR.


alibeklfc:
Hi. Thanks for your contributions. We have decided to not proceed with this PR. Since workflows works correctly as is

PR #5213, opened 2026-05-14
Open SIFT demo data in binary mode

massy-o:
## Summary
- open SIFT demo vector files with binary mode

## Why
The SIFT demo reads .fvecs/.ivecs data as fixed-width binary records. Opening the file with rb avoids platform text-mode translation, especially on Windows, and keeps the byte-size checks aligned with the actual file contents.

## Testing
- git diff --check


demos/demo_sift1M.cpp
```diff
@@ -31,7 +31,7 @@
  *****************************************************/
 
 float* fvecs_read(const char* fname, size_t* d_out, size_t* n_out) {
-    FILE* f = fopen(fname, "r");
+    FILE* f = fopen(fname, "rb");
```
massy-o, line 34: Self-review: The demo treats .fvecs/.ivecs files as fixed-width binary records, so rb avoids platform text-mode translation before the size and fread assertions run.

PR #5231, opened 2026-05-22
Add DGX Spark support and update documentation

zbrad:
- Added FAISS_OUTPUT_NAME and FAISS_CUVS_SPARK_LIBRARY cmake options
- Added FAISS_C_OUTPUT_NAME cmake option in c_api
- Added scripts for building Spark library, Python bindings, and wheel packaging
- Updated branch changes summary and various documentation files
- Removed VS Code settings from repository
- Updated environment file and build scripts


mnorris11:
@zbrad  
Can you give an overview of the benefit of this support? We already build for up through blackwell, any reason Faiss GPU or cuVS-through-Faiss is not working on DGX Spark? 

It seems like there are a lot of cmake, binary files, GPU dependency that do not belong in the PR. Also we have a build-release workflow already that also has the GPU packages, you can refer to those for examples.

zbrad:
Damn, I thought i got all the bad binaries, I'll  rescrub.  Sorry about that.

 I did find a bug on the arm, which caused it to fail during build, but also was trying to build a spark only binary set, to minimize size and build time.  I had put this project aside, but was now circling back to cleanup to contribute back.



________________________________

mnorris11:
@zbrad no worries, but before you do more work, is it the case that there was a gap in functionality, and the current Faiss GPU or cuVS-through-Faiss indeed does not work on DGX Spark platform? I have not tested it but it appears to run CUDA, so I thought Faiss would be working there.

PR #5237, opened 2026-05-26
Improve ranklist intersection size

jbajic:
This PR changes the logic of the `ranklist_intersection_size` function to use an std::unordered_set instead of a bit-flag-on-payload trick and of a sort + binary-search. Now it works with ints bigger than 2^60 since some ids might not be incrementally generated, and we could encounter them.


mdouze:
Thanks. 
Can you explain a bit better what this does? Does it allow to convert optimal search parameters into a SearchParameters object rather than setting it directly in the index object ?

jbajic:
> Thanks. Can you explain a bit better what this does? Does it allow to convert optimal search parameters into a SearchParameters object rather than setting it directly in the index object ?

Hi, not sure if you are asking me about this PR, since the questions seem more like it is about the [other PR](https://github.com/facebookresearch/faiss/pull/5236).
But here the idea is to just replace the bit-changing trick with the  usage of unordered_map, so we can use the whole space of int64

Issue #4465, opened 2025-07-28
In v1.11.0, the recall rate of the ivfflat index's 1% filtering query is very low (33.2%). Why?


qinhaoranaaa:
下面是十次1%过滤查询的召回率以及平均召回率
<img width="264" height="187" alt="Image" src="https://github.com/user-attachments/assets/eda1395f-d4ec-4635-9a52-28b7a0c6fd73" />

gtwang01:
Hello. Can you read this issue reporting guide and provide more details on your issue?
https://github.com/facebookresearch/faiss/wiki/Troubleshooting#reporting-issues

Issue #4182, opened 2025-02-10
Hard to achieve high recall rate(recall@R) for huge workload


EvagelineFEI:
Hello team,
I am using faiss.IndexHNSWFlat and my dataset is sift1b from http://corpus-texmex.irisa.fr/. 
I add 10M data into the index and then I adjust the parameters(M, efc, efs)to improve the recall rate(recall@100).
I have tried many pairs,(M from 5 to 500) but my best result stuck at 0.93.
It doesn't seem that my precision calculation code has error. And it goes better for smaller dataset(like 1M,2M, where the precision can reach 0.96 or higher). What can be the problem?
Some of my machine info:
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         46 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  80
  On-line CPU(s) list:   0-79
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Xeon(R) CPU E5-2673 v4 @ 2.30GHz
    CPU family:          6
    Model:               79
    Thread(s) per core:  2
    Core(s) per socket:  20
    Socket(s):           2
    Stepping:            1
    CPU max MHz:         3600.0000
    CPU min MHz:         1200.0000
    BogoMIPS:            4600.00

satymish:
Hi @EvagelineFEI what's the data distribution like for the smaller datasets as well as the larger datasets? Is the data randomly sampled and provide uniform data distribution?

mnorris11:
@EvagelineFEI how many query vectors are you using? All 10k from sift1b?

Issue #3344, opened 2024-04-04
Low recall with SPACEV1B dataset on GPU


karthik86248:
# Summary

I have downloaded 1 billion vectors (subset) from the original SPACEV1B dataset hosted on the [SPTAG repo](https://github.com/microsoft/SPTAG/tree/main/datasets/SPACEV1B) . The groundtruth was computed manually.

I'm using a slightly modified version of the [bench_gpu_1bn.py ](https://github.com/facebookresearch/faiss/blob/main/benchs/bench_gpu_1bn.py) script file to run ANN on SPACEV1B.

The recall values reported are relatively low (around 0.3). the index used is : OPQ24_96,IVF262144,PQ24.  
Tried experimenting higher PQ values like 38, 32 etc but no significant improvement.

From the big-ann-benchmarks competition [baseline](https://github.com/harsha-simhadri/big-ann-benchmarks/tree/main/neurips21/t3), the suggested index is : IVF1048576,SQ8. Planning to try this index next. The GPU benchmark scripts in [FAISS repo](https://github.com/facebookresearch/faiss/tree/main/benchs) don't seem to support this SQ index.

# Platform

OS:  Ubuntu 22.04.1 LTS

Faiss version:  1.7.4

Installed from:  Conda

Faiss compilation options: 

Running on:
- [ ] CPU
- [x] GPU

Interface: 
- [ ] C++
- [x] Python

# Reproduction instructions

mdouze:
Space1B should not be too problematic, see tab2 in https://proceedings.mlr.press/v176/simhadri22a/simhadri22a.pdf
Support for SQ8 was added to Faiss after the GPU bench script. However, if you run on a single GPU the dataset may not fit in the GPU ram.

Issue #3019, opened 2023-08-22
I found an issue during the IndexIVFPQ query process in version 1.7.2, and I'm not sure if it's a bug. I hope you can help me solve it.


yangshubin2023:
# Summary
```
//.........

int main() {
    int d = 64;      // dimension
    int nb = 100000; // database size
    int nq = 10000;  // nb of queries

    std::mt19937 rng;
    std::uniform_real_distribution<> distrib;

    float* xb = new float[d * nb];
    float* xq = new float[d * nq];

    // .......

    int nlist = 100;
    int k = 4;
    int m = 8;                       // bytes per vector
    faiss::IndexFlatIP quantizer(d); // the other index
    faiss::IndexIVFPQ index(&quantizer, d, nlist, m, 8, METRIC_INNER_PRODUCT);

    index.train(nb, xb);
    index.add(nb, xb);

    { // sanity check
        idx_t* I = new idx_t[k * 5];
        float* D = new float[k * 5];

        index.search(5, xb, k, D, I);

        //........

        delete[] I;
        delete[] D;
    }

    delete[] xb;
    delete[] xq;

    return 0;
}
```

# Platform
Operating System: Ubuntu 20.04.3 LTS
Kernel: Linux 5.4.0-122-generic
Architecture: x86-64

Running on:
-  CPU

Interface: 
- C++

# Reproduction instructions
After reading the Feiss code, I found that IndexIVFPQ used residual calculation during both training and data addition processes. The residual data is used to calculate fine-grained centroids, and the fine-grained centroid data is also stored as residual data.

However, during the query process, the query vector x was not subjected to residual processing and compared with the fine-grained centroid to calculate the distance. Is this correct?

In my understanding, the query quantity x should also be calculated based on the residual vector x ', and then use x' and fine-grained centroid comparison techniques based on distance to make sense.

Comparing the fine-grained centroids formed by the original vector x and residual data makes me a bit confused.

mdouze:
You are correct that the query vector is also compared based on the residual vectors. What makes you think this is not the case?

PR #5254, opened 2026-05-29
docs(ivfpq): clarify recall depends on clustered data structure

SMC17:
## Summary

Add a comment block to `tutorial/python/3-IVFPQ.py` that explains the most common confusion users hit when they first run IVFPQ on synthetic data: recall on uniform random vectors is far below the numbers reported on real embedding datasets such as SIFT1M, because IVFPQ depends on the data carrying cluster structure that the inverted-file step and the residual product quantizer can exploit.

## The doc gap

The IVFPQ tutorial currently builds an index over `np.random.random((nb, d))` plus a single coordinate ramp. A new user who copies the tutorial, swaps in their own uniform synthetic data, and measures recall will often see something an order of magnitude below the wiki numbers and conclude either that the index is broken or that their parameters are wrong. The actual cause is upstream of any parameter: uniform data has no exploitable structure. Nothing in the in-source tutorial, the class doxygen for `IndexIVFPQ`, or the README spells this out.

The issue tracker has a long tail of "low IVFPQ recall" reports, including:

- #4465 "the recall rate of the ivfflat index's 1% filtering query is very low"
- #4182 "Hard to achieve high recall rate(recall@R) for huge workload"
- #2971 "Why is the recall rate of IP lower than L2 for IVFPQ"
- #3019 IndexIVFPQ query process question
- #3344 "Low recall with SPACEV1B dataset"

Many of these resolve to the same underlying principle.

## The clarification

The added comment block:

1. States that IVFPQ is lossy in two ways (inverted-file pruning and PQ residual coding) and that both steps assume cluster structure.
2. Names the canonical real datasets (SIFT1M, sentence embeddings, learned representations) where the published recall numbers come from.
3. Warns explicitly that uniform random vectors will give recall an order of magnitude lower, and that this is a property of the data not a bug in the index.
4. Cites Jegou, Douze, Schmid 2011 (the IVFPQ paper) so a reader who wants the math has the entry point.

The tutorial code itself is unchanged. The light coordinate ramp on `xb[:, 0]` already in the example injects enough structure for a runnable demo; the comment makes that explicit.

## Why the tutorial and not the FAQ

The official FAQ lives on the GitHub wiki and is not in this repo, so a tutorial-side note is the only path that ships with the source and reaches every user who reads `tutorial/python/3-IVFPQ.py`. The IVFPQ tutorial is the highest-traffic in-source surface for this class of confusion.

## Test plan

- [x] Python syntax check passes (`python3 -m ast` parse).
- [x] No code path is changed; runtime behaviour of the tutorial is unaffected.
- [x] Comment block uses only ASCII, no em-dashes.
- [ ] Maintainer review of phrasing and citation form.

### DIFF
--- a/tutorial/python/3-IVFPQ.py
+++ b/tutorial/python/3-IVFPQ.py
@@ -3,6 +3,27 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+# Note on recall for IVFPQ
+# ------------------------
+# IVFPQ is a lossy index: the inverted-file step prunes the search to a few
+# Voronoi cells, and the product quantizer replaces each residual with its
+# nearest codebook entry. Both steps assume the database has exploitable
+# cluster structure, so that (a) most true neighbours of a query fall in
+# the few cells visited, and (b) residuals concentrate around a small set
+# of prototypes. Real-world embedding datasets (image features such as
+# SIFT1M, sentence embeddings, learned representations) carry that
+# structure and give the recall numbers reported in the Faiss benchmarks.
+#
+# Uniform random vectors do not. With no clusters to exploit, nprobe=10
+# out of nlist=100 cells visits roughly 10% of the database and per-cell
+# quantization noise is large, so top-k recall is typically an order of
+# magnitude below what is seen on clustered data. The synthetic data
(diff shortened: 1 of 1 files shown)


SMC17:
Sorry — been testing some things and this leaked out. Closing.

PR #5275, opened 2026-06-04
Validate zero-copy vector sizes when reading indexes

dfgvaetyj3456356-hash:
## Summary

`read_vector_base()` handles mmap and zero-copy vector reads during index deserialization, but it did not apply the vector byte-limit and overflow checks used by the normal `READVECTOR`/`READXBVECTOR` paths before mapping serialized data.

This updates the mmap/zero-copy fast path to validate the multiplied element count before creating a view, so it matches the existing deserialization guardrails for owned vectors.

## Test plan

- Added `ReadIndexDeserialize.ZeroCopyReadXBVectorByteLimit`, which builds a small serialized `IndexFlatL2` payload and verifies `ZeroCopyIOReader` rejects an `READXBVECTOR` payload at the strict byte-limit boundary.
- Ran `git diff --check` locally.

I could not run the C++ test binary in this checkout because `cmake` is not installed in the local environment.

### DIFF
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -158,8 +158,11 @@ template <typename VectorT>
 size_t checked_deserialization_vector_size(
         size_t size,
         size_t size_multiplier) {
+    FAISS_THROW_IF_NOT_MSG(
+            size_multiplier > 0,
+            "deserialization vector size multiplier must be positive");
     FAISS_THROW_IF_NOT_FMT(
-            size_multiplier == 0 || size <= SIZE_MAX / size_multiplier,
+            size <= SIZE_MAX / size_multiplier,
             "deserialization vector size %zu would overflow with multiplier %zu",
             size,
             size_multiplier);
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -155,6 +155,29 @@ void set_deserialization_lattice_r2_limit(size_t value) {
 // * if `size_multiplier` is defined, then a size will be multiplied by it.
 // * returns true is the case was handled; otherwise, false
 template <typename VectorT>
+size_t checked_deserialization_vector_size(
+        size_t size,
+        size_t size_multiplier) {
+    FAISS_THROW_IF_NOT_FMT(
+            size_multiplier == 0 || size <= SIZE_MAX / size_multiplier,
+            "deserialization vector size %zu would overflow with multiplier %zu",
+            size,
+            size_multiplier);
+
+    size *= size_multiplier;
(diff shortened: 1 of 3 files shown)


dfgvaetyj3456356-hash:
Strengthened this after another pass. The update now:\n\n- removes the unnecessary zero-multiplier escape in the shared size check by making the positive-multiplier invariant explicit\n- reuses a crafted minimal IndexFlatL2 READXBVECTOR payload helper\n- keeps the ZeroCopyIOReader byte-limit regression\n- adds the matching MappedFileIOReader regression, since both readers use the same read_vector_base() fast path and both need to reject before returning a view\n\nLocal validation: git diff --check passes; C++ test execution is blocked on my Windows checkout because cmake is not installed locally.

dfgvaetyj3456356-hash:
Follow-up validation update:

- Installed the missing local build tooling in WSL and built the CPU-only `faiss_test` target with CMake/Ninja.
- While validating, I noticed `tests/test_read_index_deserialize.cpp` was not included in `FAISS_TEST_SRC`, so the new regressions were not being built by the CMake test binary. I added that wiring in `ead8da56a`.
- Rebuilt `faiss_test` successfully.
- Ran the focused regressions successfully:

```
./tests/faiss_test --gtest_filter='ReadIndexDeserialize.ZeroCopyReadXBVectorByteLimit:ReadIndexDeserialize.MappedFileReadXBVectorByteLimit'
[==========] Running 2 tests from 1 test suite.
[ RUN      ] ReadIndexDeserialize.ZeroCopyReadXBVectorByteLimit
[       OK ] ReadIndexDeserialize.ZeroCopyReadXBVectorByteLimit (1 ms)
[ RUN      ] ReadIndexDeserialize.MappedFileReadXBVectorByteLimit
[       OK ] ReadIndexDeserialize.MappedFileReadXBVectorByteLimit (0 ms)
[  PASSED  ] 2 tests.
```

faiss/impl/index_read.cpp
```diff
@@ -173,8 +199,8 @@ bool read_vector_base(
                 READANDCHECK(&size, 1);
             }
 
-            // perform the size multiplication
-            size *= size_multiplier.value_or(1);
+            size = checked_deserialization_vector_size<VectorT>(
+                    size, size_multiplier.value_or(1));
```
mnorris11, line 203: Can we use `size = mul_no_overflow(size, size_multiplier, "deserialization vector size");`?

faiss/impl/index_read.cpp
```diff
@@ -209,8 +235,8 @@ bool read_vector_base(
...
```
mnorris11, line 239: ditto

PR #5293, opened 2026-06-09
Validate codes.size() in RaBitQ flat index deserialization

AlexMelanFromRingo:
The flat RaBitQ index readers (`fourcc("Ixrq")` and `fourcc("Ixrr")` in `read_index`, `faiss/impl/index_read.cpp`) assign `code_size` straight from the file and read the `codes` vector without checking that its length matches `ntotal * code_size`.

Every other flat index validates this on load — `IndexFlat` (`index_read.cpp:1561`), `IndexLSH` (`:1600`), `IndexPQ` (`:1611`), `IndexResidualQuantizer` (`:1639`) — and the IVF RaBitQ readers (`Iwrq`/`Iwrr`) recompute `code_size` and go through `read_InvertedLists`’ size checks. The flat RaBitQ path does neither, so an index whose header `ntotal`/`code_size` do not match the stored `codes` buffer loads without error and then reads past the buffer during `search()` / `sa_decode()` (the kernel in `IndexRaBitQ.cpp` strides `codes.data() + i * code_size` for `i` in `0..ntotal`).

This PR adds the same `codes.size() == ntotal * code_size` assertion the other flat indexes already use, in both the `Ixrq` and `Ixrr` arms.

### Reproduce (current `main` / `faiss-cpu` 1.14.2)
```python
import faiss, numpy as np, struct
d=8; idx=faiss.IndexRaBitQ(d)
xb=np.random.RandomState(0).randn(4,d).astype("float32"); idx.train(xb); idx.add(xb)
raw=bytearray(faiss.serialize_index(idx))   # Ixrq, ntotal=4, code_size=9, 36-byte codes
struct.pack_into("<q", raw, 8, 100_000_000) # edit ONLY the ntotal header field
open("m.faiss","wb").write(raw)
idx2=faiss.read_index("m.faiss")            # before: succeeds (ntotal=100000000, 36-byte codes)
idx2.search(np.zeros((1,d),"float32"), 5)   # before: out-of-bounds read / segfault in IndexRaBitQ::search
```
With this patch, `read_index` raises `codes.size() == ntotal * code_size` instead of loading a self-inconsistent index.


faiss/impl/index_read.cpp
```diff
@@ -2546,6 +2546,12 @@ std::unique_ptr<Index> read_index_up(IOReader* f, int io_flags) {
 
         // rabitq.nb_bits is already set to 1 by read_RaBitQuantizer
         idxq->code_size = idxq->rabitq.code_size;
+        // Reconcile the file-supplied ntotal/code_size with the actual codes
+        // length, like the other flat indexes (IndexFlat/LSH/PQ/...), so a
+        // crafted file can't make search() read past the codes buffer.
+        FAISS_THROW_IF_NOT(
+                idxq->codes.size() ==
+                static_cast<size_t>(idxq->ntotal) * idxq->code_size);
```
mnorris11: We can re-use the existing utility `validate_code_size_match`, and remove the comment since it is pretty self evident

faiss/impl/index_read.cpp
```diff
@@ -2564,6 +2570,9 @@ std::unique_ptr<Index> read_index_up(IOReader* f, int io_flags) {
                 idxq->qb);
 
         idxq->code_size = idxq->rabitq.code_size;
+        FAISS_THROW_IF_NOT(
+                idxq->codes.size() ==
+                static_cast<size_t>(idxq->ntotal) * idxq->code_size);
```
mnorris11: (same here)

faiss/impl/index_read.cpp
AlexMelanFromRingo: Thanks for the review, and apologies for the slow turnaround. Rebased onto current `main`, dropped both comments as suggested, and added regression tests.

On re-using `validate_code_size_match`: I tried wiring it in, but I don't think it fits this particular check, and I'd rather flag that than quietly use it wrong.

Every existing call site compares a *scalar* `code_size` field read from the file against the value derived from the quantizer (`idxl->code_size` vs `(nbits + 7) / 8`, `idxr->code_size` vs `idxr->rq.code_size`, and so on), and it reports `"%s code_size mismatch: stored %zd vs derived %zd"`. What this check validates is the *length of the codes buffer* against `ntotal * code_size`, so that message would print byte counts and call them a `code_size`.

`IndexRaBitQ` also has no separately-stored `code_size` to validate: it is assigned straight from the quantizer (`idxq->code_size = idxq->rabitq.code_size`), so there is no stored-vs-derived pair for the helper to compare here.

For a codes-buffer check the established idiom across the flat readers is the `mul_no_overflow` assertion, so I matched it exactly:

```cpp
FAISS_THROW_IF_NOT(
        idxq->codes.size() ==
        mul_no_overflow(
                (size_t)idxq->ntotal,
                idxq->code_size,
                "IndexRaBitQ codes"));
```

This is byte-for-byte what `IndexFlat`, `IndexLSH`, `IndexPQ` and `IndexResidualQuantizer` already do, and it keeps the existing `"codes.size()"` assertion tests consistent.

It also closes a real hole in my original version. `static_cast<size_t>(ntotal) * code_size` can wrap: `code_size` is odd in the default single-bit case (`(d + 7) / 8 + 8` = 9 for d=8), so it is invertible mod 2^64 and an attacker can solve for an `ntotal` that makes the raw product wrap to exactly `codes.size()`. That would sail through my original assertion and restore the out-of-bounds read. `mul_no_overflow` throws instead of wrapping.

Added `RaBitQCodesSizeMismatch_Ixrq` / `_Ixrr` to `tests/test_read_index_deserialize.cpp`, mirroring the existing `IndexPQCodesSizeMismatch`.

Happy to switch to `validate_code_size_match(codes.size(), ...)` if you would still prefer it here, just say the word.

faiss/impl/index_read.cpp
AlexMelanFromRingo: Same treatment applied to this arm: `mul_no_overflow` assertion matching the sibling flat readers, comment dropped. See the reasoning in the thread above.

PR #5297, opened 2026-06-10
Speed up Top1 and Reservoir add_results with AVX2/AVX512

mulugetam:
This PR replaces the scalar `add_results` loops in `Top1BlockResultHandler` and `ReservoirBlockResultHandler` with runtime-dispatched SIMD kernels.

### Top1 (k=1)

- **AVX2:** 8-wide branchless argmin/argmax via `_mm256_blendv_ps` / `_mm256_blendv_epi8`, with horizontal reduction and scalar tail.
- **AVX512:** 16-wide via `_mm512_mask_blend_ps` / `_mm512_mask_blend_epi32`.

### Reservoir (k>1)

- **NONE:** Scalar path with `res.i` and `res.threshold` hoisted into local variables, eliminating per-iteration struct field reloads caused by `float*` / `int64_t*` aliasing.
- **AVX512:** `VPCOMPRESSPS` + `VPCOMPRESSD` bulk-insert all elements beating the threshold in a single pass. Indices are widened from int32 to int64 via `_mm512_cvtepi32_epi64`. Falls back to the scalar path when reservoir capacity < 32 (2× lane width), where compress setup cost outweighs throughput benefit.

### Performance

Benchmarking using `benchs/bench_index_flat.py` (on Xeon) gives the below gains.

```
+-----+------+------+------+-------+
|  d  | k=10 | k=32 | k=64 | k=100 |
+-----+------+------+------+-------+
|  16 | +10% | +15% | +28% |  +31% |
|  32 |  +7% | +12% | +25% |  +28% |
|  64 |  +7% | +11% | +21% |  +25% |
| 128 |  +7% | +10% | +18% |  +19% |
+-----+------+------+------+-------+
```
k=1 is neutral — distance compute dominates at all tested dimensions.

Raw results (old and new): https://gist.github.com/mulugetam/82ef99c25f532f7439070c802b746865

### DIFF
--- a/faiss/impl/ResultHandler.h
+++ b/faiss/impl/ResultHandler.h
@@ -18,6 +18,7 @@
 #include <faiss/impl/InvertedListScannerStats.h>
 #include <faiss/utils/Heap.h>
 #include <faiss/utils/partitioning.h>
+#include <faiss/utils/simd_levels.h>
 #include <algorithm>
 #include <iostream>
 
@@ -234,23 +235,9 @@ struct Top1BlockResultHandler : TopkBlockResultHandler<C, use_sel> {
     }
 
     /// add results for query i0..i1 and j0..j1
-    void add_results(size_t j0, size_t j1, const T* dis_tab_2) final {
-        for (size_t i = i0; i < i1; i++) {
-            const T* dis_tab_i = dis_tab_2 + (j1 - j0) * (i - i0) - j0;
-
-            auto& min_distance = this->dis_tab[i];
-            auto& min_index = this->ids_tab[i];
-
-            for (size_t j = j0; j < j1; j++) {
-                const T distance = dis_tab_i[j];
-
-                if (C::cmp(min_distance, distance)) {
-                    min_distance = distance;
-                    min_index = j;
-                }
-            }
-        }
-    }
+    /// Implemented in result_handler.cpp; dispatches via
+    /// with_selected_simd_levels to top1_add_results_tpl<C, use_sel, SL>.
+    void add_results(size_t j0, size_t j1, const T* dis_tab_2) final;
 
     void add_result(const size_t i, const T dis, const TI idx) {
         auto& min_distance = this->dis_tab[i];
@@ -553,20 +540,9 @@ struct ReservoirBlockResultHandler : TopkBlockResultHandler<C, use_sel> {
     }
 
     /// add results for query i0..i1 and j0..j1
-    void add_results(size_t j0, size_t j1, const T* dis_in) {
-#pragma omp parallel for
-        for (int64_t i = static_cast<int64_t>(i0); i < static_cast<int64_t>(i1);
-             i++) {
-            ReservoirTopN<C>& reservoir =
-                    reservoirs[i - static_cast<int64_t>(i0)];
-            const T* dis_tab_i =
-                    dis_in + (j1 - j0) * (i - static_cast<int64_t>(i0)) - j0;
-            for (size_t j = j0; j < j1; j++) {
-                T dis = dis_tab_i[j];
-                reservoir.add_result(dis, j);
-            }
-        }
-    }
+    /// Implemented in result_handler.cpp; dispatches via
+    /// with_selected_simd_levels to reservoir_add_results_tpl<C, use_sel, SL>.
+    void add_results(size_t j0, size_t j1, const T* dis_in);
 
     /// series of results for queries i0..i1 is done
     void end_multiple() final {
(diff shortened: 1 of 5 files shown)


mulugetam:
@mnorris11 Could you take a look?

faiss/impl/result_handler/ResultHandler.cpp
```diff
@@ -0,0 +1,235 @@
...
+// Reservoir: scalar fallback + AVX512 compress path.
+// VPCOMPRESSPS/VPCOMPRESSD require AVX512F so there is no AVX2 path.
+// On non-AVX512 hosts the dispatch falls back to NONE automatically.
+constexpr int RESERVOIR_SIMD_LEVELS =
+        (1 << int(SIMDLevel::NONE)) | (1 << int(SIMDLevel::AVX512));
+
+// ----------------------------------------------------------------
+// Scalar (NONE) helper implementations
+// ----------------------------------------------------------------
+
```
mdouze, line 34: put this in an anonymous namespace

faiss/impl/result_handler/ResultHandler.cpp
```diff
@@ -0,0 +1,235 @@
...
+        }
+        res.i = ri;
+    }
+}
+
+// ----------------------------------------------------------------
+// SIMDLevel::NONE explicit specialisations
+// ----------------------------------------------------------------
+
+template <>
```
mdouze: perhaps make a macro to trigger the instantiations for all combinatoins

faiss/impl/ResultHandler.h
```diff
@@ -234,23 +235,9 @@ struct Top1BlockResultHandler : TopkBlockResultHandler<C, use_sel> {
...
-                const T distance = dis_tab_i[j];
-
-                if (C::cmp(min_distance, distance)) {
-                    min_distance = distance;
-                    min_index = j;
-                }
-            }
-        }
-    }
+    /// Implemented in result_handler.cpp; dispatches via
```
mdouze, line 238: the initial idea was to keep this in the compile unit so that it could be inlined, but indeed it is probably not worth it for multiple results in one run.

mdouze (commented):
Thanks for this PR. 

There is another AVX512 implementation of the reservoir here 

https://github.com/facebookresearch/faiss/blob/20afed0fa3aefe2d4f5bf6a5f6ab1d1459efec05/benchs/avx512_result_handlers/faiss_avx512_result_handler.h

but I did not find time to properly integrate it into the main code...

mulugetam:
@mnorris11 @mdouze Thank you. I have addressed your comments. Could you take another look?

mulugetam:
@mnorris11 @mdouze The current PR uses the existing `shrink_fuzzy`. I noticed the version you started implementing uses `argpartition`.

Below are the results from running this [microbenchmark](https://gist.github.com/mulugetam/335b40c49d9be27364c4c6b2ff772e6d) to compare the performance of `shrink_fuzzy` against `argpartition`. Do we want to use `argpartition` instead (on this PR or a followup PR)? Also, I noticed that `argpartition` uses `int32_t`.

```text
Compaction only (one call on a full reservoir of 2k elements)
```

```
+-------+-------------+--------------+-------------+----------+
|   k   | ShrinkFuzzy | ArgPartition |   Winner    |  Ratio   |
|       |    (ns)     |     (ns)     |             |          |
+-------+-------------+--------------+-------------+----------+
|    10 |         216 |          231 | ShrinkFuzzy |  1.07x   |
|    50 |         288 |          379 | ShrinkFuzzy |  1.32x   |
|   100 |         391 |          516 | ShrinkFuzzy |  1.32x   |
|   500 |        1285 |         2052 | ShrinkFuzzy |  1.60x   |
|  1000 |        3092 |         3052 | tie         |  1.01x   |
|  5000 |       11475 |        15412 | ShrinkFuzzy |  1.34x   |
| 10000 |       59780 |        35155 | ArgPartition|  1.70x   |
+-------+-------------+--------------+-------------+----------+
```

```text
Full loop (end-to-end: 100k distances, compactin on fill)
```

```
+-------+-------------+--------------+--------------+
|   k   | ShrinkFuzzy | ArgPartition | ArgPartition |
|       |    (ns)     |     (ns)     |   speedup    |
+-------+-------------+--------------+--------------+
|    10 |      52833  |       45384  |     1.16x    |
|    50 |     254569  |       49275  |     5.17x    |
|   100 |     441378  |       53232  |     8.29x  * |
|   500 |     534447  |       80331  |     6.65x    |
|  1000 |     534787  |      120961  |     4.42x    |
|  5000 |     617403  |      286266  |     2.16x    |
| 10000 |     538920  |      383021  |     1.41x    |
+-------+-------------+--------------+--------------+
* peak speedup
```

PR #5326, opened 2026-06-19
Fix invalid L2 distances from IndexHNSWFlat with stale cached_l2norms

### LINKED ISSUE #5320: IndexHNSWFlat can return invalid L2 results after IndexFlatL2 sync_l2norms() followed by add()
## Current Behavior

When using `IndexHNSWFlat` with `METRIC_L2`, calling `sync_l2norms()` on the underlying `IndexFlatL2` storage after a partial `add()`, and then adding more vectors, can make HNSW return invalid L2 distances and inconsistent Top-k results.

This violates the dimension permutation invariance of L2 distance.

For L2 distance, if the same dimension permutation is applied to every database vector and every query vector, all pairwise distances should remain unchanged:

```text
d(P(x), P(q)) = d(x, q)
```

Therefore, the squared L2 distances and Top-k labels should remain unchanged.

However, when the underlying `IndexFlatL2` storage has stale `cached_l2norms`, `IndexHNSWFlat` can return different Top-k results before and after a common dimension permutation. It can also return negative squared L2 distances, which are invalid.

This is not normal ANN variance. The exact flat oracle confirms that the permutation preserves all expected L2 distances and Top-k IDs. The negative L2 distances indicate that invalid distance values are being computed.

## Actual output on the affected path:

```text
Expected exact Top-k for query row 0:

IDs: [128, 151, 157, 11, 145, 52, 33, 20, 109, 42]
D:   [3134.0, 3149.0, 3287.0, 3966.0, 4222.0, 5085.0, 5375.0, 5673.0, 6191.0, 6278.0]

Actual HNSW with stale norms before permutation:

IDs: [157, 128, 151, 163, 11, 145, 140, 148, 172, 191]
D:   [-7465.0, -4375.0, -3405.0, -1575.0, -629.0, 259.0, 289.0, 533.0, 567.0, 637.0]

Actual HNSW with stale norms after permutation:

IDs: [157, 128, 151, 163, 178, 11, 145, 140, 42, 148]
D:   [-7465.0, -4375.0, -3405.0, -1575.0, -913.0, -629.0, 259.0, 289.0, 387.26953125, 533.0]
```

Order comparison:

```text
expected exact labels = [128, 151, 157, 11, 145, 52, 33, 20, 109, 42]
actual original labels = [157, 128, 151, 163, 11, 145, 140, 148, 172, 191]
actual permuted labels = [157, 128, 151, 163, 178, 11, 145, 140, 42, 148]
```

For example, squared L2 distances should never be negative, but the returned distances include values such as:

```text
-7465.0
-4375.0
-3405.0
-1575.0
```

This happens after the following sequence:

1. Add a small first batch of vectors into `IndexHNSWFlat`.
2. Downcast the underlying storage to `IndexFlatL2`.
3. Call `storage.sync_l2norms()`.
4. Add more vectors into the HNSW index.
5. Search with HNSW.

After step 4, `cached_l2norms.size()` remains smaller than `ntotal`, but HNSW search can still use the cached-norm distance computer.

## Steps to Reproduce

The following script creates two `IndexHNSWFlat` indexes:

1. one with the original vectors;
2. one with the same dimension permutation applied to every database vector and query vector.

Both indexes call `sync_l2norms()` after a partial add, then add the remaining vectors. This leaves the underlying `IndexFlatL2::cached_l2norms` stale.

```python
import faiss
import numpy as np

np.set_printoptions(precision=6, suppress=False)

seed = 0
d = 8
nb = 200
nq = 20
k = 10
first_batch = 10
M = 16
perm = np.array([3, 0, 7, 1, 6, 2, 5, 4])

rng = np.random.default_rng(seed)
xb = rng.integers(-50, 51, (nb, d)).astype("float32")
xq = rng.integers(-50, 51, (nq, d)).astype("float32")


def build_hnsw_with_stale_l2norms(vectors):
    index = faiss.IndexHNSWFlat(d, M, faiss.METRIC_L2)
    index.hnsw.efConstruction = 128
    index.hnsw.efSearch = 256

    index.add(vectors[:first_batch])

    storage = faiss.downcast_index(index.storage)
    storage.sync_l2norms()

    # This leaves cached_l2norms stale: size is still first_batch.
    index.add(vectors[first_batch:])
    return index


def flat_search(vectors, queries):
    index = faiss.IndexFlatL2(d)
    index.add(vectors)
    return index.search(queries, k)


D_gt, I_gt = flat_search(xb, xq)
D_gt_p, I_gt_p = flat_search(xb[:, perm], xq[:, perm])

assert np.array_equal(I_gt, I_gt_p)
assert np.array_equal(D_gt, D_gt_p)

idx = build_hnsw_with_stale_l2norms(xb)
idx_p = build_hnsw_with_stale_l2norms(xb[:, perm])

D, I = idx.search(xq, k)
D_p, I_p = idx_p.search(xq[:, perm], k)

row = int(np.where((I != I_p).any(axis=1))[0][0])

print("trigger_config:", {
    "index": "IndexHNSWFlat",
    "metric": "METRIC_L2",
    "d": d,
    "nb": nb,
    "nq": nq,
    "k": k,
    "M": M,
    "efConstruction": 128,
    "efSearch": 256,
    "first_batch_before_sync_l2norms": first_batch,
    "cached_l2norms_size_after_second_add": faiss.downcast_index(idx.storage).cached_l2norms.size(),
    "ntotal": idx.ntotal,
    "permutation": perm.tolist(),
})

print("violating_query_row:", row)
print("expected_topk_ids:", I_gt[row].tolist())
print("expected_distances:", D_gt[row].tolist())
print("actual_original_ids:", I[row].tolist())
print("actual_original_distances:", D[row].tolist())
print("actual_permuted_ids:", I_p[row].tolist())
print("actual_permuted_distances:", D_p[row].tolist())

if not np.array_equal(I, I_p):
    print("BUG REPRODUCED")
else:
    raise SystemExit("bug did not reproduce")
```

## Expected Behavior

For squared L2 distance, applying the same dimension permutation to all stored vectors and all query vectors should preserve all pairwise distances.

Example:

```text
q = [q0, q1, q2, q3]
x = [x0, x1, x2, x3]

perm = [2, 0, 3, 1]

P(q) = [q2, q0, q3, q1]
P(x) = [x2, x0, x3, x1]
```

The squared L2 distance remains unchanged:

```text
||x - q||^2 = ||P(x) - P(q)||^2
```

Therefore, the Top-k labels and squared L2 distances should remain unchanged.

For the reproduction above, the exact flat oracle gives the expected Top-k for query row 0:

```text
IDs: [128, 151, 157, 11, 145, 52, 33, 20, 109, 42]
D:   [3134.0, 3149.0, 3287.0, 3966.0, 4222.0, 5085.0, 5375.0, 5673.0, 6191.0, 6278.0]
```

The expected HNSW behavior is:

1. It may return approximate neighbors, but the returned L2 distances should be valid.
2. Squared L2 distances should never be negative.
3. A common dimension permutation should not change the actual L2 distance values.
4. If `cached_l2norms` is used, the cache should cover all vectors in the index.

## Actual Behavior

`IndexHNSWFlat` returns invalid negative squared L2 distances and inconsistent Top-k results when `cached_l2norms` is stale.

The stale cache is created by:

```text
index.add(first_batch)
storage.sync_l2norms()
index.add(remaining_vectors)
```

After the second add:

```text
cached_l2norms.size() = 10
ntotal = 200
```

However, HNSW search can still use the cached-norm distance computer. For vectors added after `sync_l2norms()`, the distance computer may access norm entries beyond the cached norm array.

This produces invalid distances such as:

```text
-7465.0
-4375.0
-3405.0
-1575.0
```

The exact flat oracle confirms that the dimension permutation itself preserves all expected L2 distances and Top-k IDs. Therefore, this is not caused by the metamorphic transformation.

notandruu:
Fixes #5320

## Root cause

`IndexFlatL2::get_FlatCodesDistanceComputer()` selects the cached-norms distance computer (`FlatL2WithNormsDis`) whenever `cached_l2norms` is non-empty. That computer reads `cached_l2norms[i]` for every vector up to `ntotal`. However `add()` after a `sync_l2norms()` does not extend or invalidate `cached_l2norms`, so its size can be smaller than `ntotal`. The computer then reads stale / out-of-range norms and returns invalid squared L2 distances (including negative values), with inconsistent and wrong top-k results. This is reachable from `IndexHNSWFlat` (METRIC_L2), whose storage is an `IndexFlatL2`.

## Fix

Take the cached path only when `cached_l2norms.size() == ntotal`, i.e. when the cache covers every vector; otherwise fall back to the on-the-fly L2 computer. The fast path is unchanged for the normal usage (add all vectors, then `sync_l2norms()` once). One line, plus a comment.

## Reproduction and validation

Repro: add a partial batch to an `IndexHNSWFlat`, `sync_l2norms()` on the downcast `IndexFlatL2`, add the rest, then search. Validated with a from-source build (BLAS=Accelerate, libomp) using a C++ negative control:

| build | negative squared-L2 distances | min distance |
|---|---|---|
| before fix | 46 / 50 | -7628.0 |
| after fix | 0 | 739.0 |

Regression test `TestSyncL2Norms.test_indexflat_l2_sync_norms_stale_after_add` added next to the existing `sync_l2norms` test. It reproduces the scenario and asserts no negative distances and that each query's nearest neighbour matches an exact `IndexFlatL2` oracle. Run against the released wheel (which has the bug) the test fails (min distance -11114); with this fix the underlying cause is removed.

Note: I could not build the Python bindings locally (no swig), so the C++ fix was validated via a from-source C++ build and the Python regression test was validated to fail on the unpatched build; CI will exercise it against the patched bindings.


faiss/IndexFlat.cpp
```diff
@@ -438,7 +438,11 @@ void IndexFlatL2::clear_l2norms() {
 
 FlatCodesDistanceComputer* IndexFlatL2::get_FlatCodesDistanceComputer() const {
     if (metric_type == METRIC_L2) {
-        if (!cached_l2norms.empty()) {
+        // Only use the cached-norms distance computer when the cache covers
+        // every vector. add() after a sync_l2norms() leaves cached_l2norms
+        // shorter than ntotal; using it then would read stale/out-of-range
+        // norms and yield invalid (e.g. negative) L2 distances.
```
mnorris11, line 408: Adding the check seems correct, but we can trim the AI comment a bit:
```
    // Only use the cached-L2norm path when the cache covers every vector.
    // add() does not maintain it.
```

tests/test_index.py
```diff
@@ -231,6 +231,42 @@ def test_indexflat_l2_sync_norms_1(self):
         #  np.testing.assert_equal(Iref, I1)
         np.testing.assert_equal(D3, D1)
 
+    def test_indexflat_l2_sync_norms_stale_after_add(self):
+        # Regression test for #5320: calling sync_l2norms() after a partial
+        # add() and then adding more vectors left cached_l2norms shorter than
+        # ntotal. The cached-norms distance computer then read stale/out-of-
+        # range norms, returning invalid (e.g. negative) squared L2 distances
+        # and inconsistent top-k results.
```
mnorris11, line 266: ditto for this comment

tests/test_index.py
```diff
@@ -231,6 +231,42 @@ def test_indexflat_l2_sync_norms_1(self):
...
+        index.add(xb[first_batch:])
+        D, I = index.search(xq, k)
+
+        # Squared L2 distances must never be negative.
+        self.assertGreaterEqual(D.min(), 0.0)
+        # The exact nearest neighbour for each query must be found with the
+        # correct (non-negative) distance.
+        for row in range(nq):
+            self.assertEqual(I[row, 0], Iref[row, 0])
+            np.testing.assert_almost_equal(D[row, 0], Dref[row, 0], decimal=4)
```
mnorris11, line 295: I think this tests a slightly different issue.
Right now it is saying, "did HNSW find the true nearest neighbor?". But HNSW is ANN, so it will not always find it.
Instead, we should make sure the returned distances are valid.
```
        for row in range(nq):
            for j in range(k):
                true_d = ((xb[I[row, j]] - xq[row]) ** 2).sum()
                np.testing.assert_almost_equal(D[row, j], true_d, decimal=4)
```
This tests, "for the returned results from HNSW, recompute the distance, and make sure we got the right value"

PR #5335, opened 2026-06-22
Fix DoS in range_search with empty SearchParameters (fixes #5332)

### LINKED ISSUE #5332: IndexIDMap::range_search crashes/hangs when called with an empty SearchParameters object
## How to reproduce this bug?

This bug can be reproduced with an `IndexIDMap` wrapping an exact `IndexFlatL2` index. The issue is triggered when `range_search` is called with an explicitly provided but empty `SearchParameters()` object.

```python
import subprocess
import sys
import textwrap
import platform

import faiss
import numpy as np

child_code = r'''
import faiss
import numpy as np

index = faiss.IndexIDMap(faiss.IndexFlatL2(2))

xb = np.array(
    [
        [0, 0],
        [1, 0],
        [0, 2],
    ],
    dtype="float32",
)
ids = np.array([10, 20, 30], dtype="int64")

index.add_with_ids(xb, ids)

xq = np.array([[0, 0]], dtype="float32")

print("before")
print(index.range_search(xq, 5.0, params=faiss.SearchParameters()))
print("after")
'''

print("FAISS version:", faiss.__version__)
print("Python version:", sys.version.split()[0])
print("OS:", platform.platform())

index = faiss.IndexIDMap(faiss.IndexFlatL2(2))

xb = np.array(
    [
        [0, 0],
        [1, 0],
        [0, 2],
    ],
    dtype="float32",
)
ids = np.array([10, 20, 30], dtype="int64")

index.add_with_ids(xb, ids)

xq = np.array([[0, 0]], dtype="float32")

print("Expected result without params:")
print(index.range_search(xq, 5.0))

try:
    p = subprocess.run(
        [sys.executable, "-u", "-c", child_code],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        text=True,
        timeout=5,
    )

    print("Actual child return code:", p.returncode)
    print("Actual child stdout:", p.stdout)
    print("Actual child stderr:", p.stderr)

except subprocess.TimeoutExpired as e:
    print("Actual result: child process did not return within 5 seconds")
    print("Child stdout before timeout:", e.stdout)
    print("Child stderr before timeout:", e.stderr)
```

## What is the expected behavior?

`SearchParameters::sel` is optional. If `sel == nullptr`, no ID filtering should be applied.

Therefore, the following two calls should behave the same:

```python
index.range_search(xq, 5.0)
index.range_search(xq, 5.0, params=faiss.SearchParameters())
```

Expected result:

```text
lims      = [0, 3]
distances = [0.0, 1.0, 4.0]
labels    = [10, 20, 30]
```

An empty `SearchParameters()` object should be equivalent to not passing `params`.

## What is the actual behavior?

On Windows with `pip faiss-cpu 1.14.2`, the child process prints:

```text
before
```

and then never returns from the native call.

## Server Version / FAISS Version

FAISS version tested:

```text
1.14.2
```

## Setup

```text
OS: Windows 11
Python: 3.13.9
Installation: pip faiss-cpu
Index type: IndexIDMap(IndexFlatL2)
Metric: L2
Search type: exact range_search
```

FERASSYYU:


mnorris11:
Apologies, second one today I didn't see the corresponding PR for, was going through our backlog and saw the issue first. It is fixed already.

PR #5341, opened 2026-06-24
Add fast scan variants of SQ indexes - IndexSQFastScan and IndexIVFSQFastScan

mulugetam:
This PR introduces `IndexSQFastScan` and `IndexIVFSQFastScan`, fast scan variants of scalar quantized indexes.

**IndexSQFastScan** (33f2cb639bd7a888b2ab7578083b73753c2fd3d3)

Introduces `IndexSQFastScan`, a fast scan variant of `IndexScalarQuantizer` that leverages the PQ4 FastScan SIMD infrastructure (`vpshufb`) for accelerated search over scalar-quantized codes.

Three code paths depending on quantizer type:

- Native 4-bit (`QT_4bit`, `QT_4bit_uniform`): codes map directly onto the 16-level PQ4 fast scan. ~11x speedup, no precision loss.

- Rerank (`QT_6bit`, `QT_8bit`, `QT_8bit_uniform`, `QT_8bit_direct`, `QT_8bit_direct_signed`): re-quantize to 4-bit for a coarse fast scan first pass, then rerank top-k*rerank_factor candidates with exact original-precision distances. ~7-9x speedup, near-zero precision loss at rerank_factor=4.

- Fallback (`QT_fp16`, `QT_bf16`, `TurboQuant`, etc.): delegates to the ScalarQuantizer's own SIMD-optimized scanner. Same speed as IndexScalarQuantizer — no fast scan benefit, but provides a unified interface for all quantizer types.

**benchmark**: `benchs/bench_index_sq_fastscan.py`

```text
+---------------------+----------------------+---------+---------+--------+--------+
| Category            | QType                | Speedup | ms/q SQ | R@1    | R@10   |
+---------------------+----------------------+---------+---------+--------+--------+
| Native 4-bit (SIMD) | QT_4bit              | 11.9x   | 0.240   | 0.7714 | 0.9985 |
|                     | QT_4bit_uniform      | 11.7x   | 0.235   | 0.7180 | 0.9952 |
+---------------------+----------------------+----------+---------+--------+--------+
| Rerank (rf=4)       | QT_6bit              |  9.3x   | 0.202   | 0.9371 | 1.0000 |
|                     | QT_8bit              |  8.4x   | 0.173   | 0.9783 | 1.0000 |
|                     | QT_8bit_uniform      |  6.7x   | 0.138   | 0.9728 | 1.0000 |
|                     | QT_8bit_direct       |  4.7x   | 0.101   | 0.9914 | 1.0000 |
|                     | QT_8bit_direct_signed|  4.9x   | 0.101   | 0.5999 | 0.6591 |
+---------------------+----------------------+----------+---------+--------+--------+
| Fallback (~1.0x)    | QT_fp16              |  0.9x   | 0.373   | 0.9914 | 1.0000 |
|                     | QT_bf16              |  1.0x   | 0.365   | 0.9914 | 1.0000 |
|                     | QT_8bit_tqmse        |  1.0x   | 0.261   | 0.6033 | 0.9561 |
|                     | QT_5bit_tq           |  1.0x   | 3.013   | 0.2680 | 0.6248 |
|                     | QT_4bit_tq           |  1.0x   | 2.416   | 0.1511 | 0.3994 |
|                     | QT_3bit_tq           |  1.0x   | 0.411   | 0.0487 | 0.1691 |
|                     | QT_2bit_tq           |  1.1x   | 0.231   | 0.0376 | 0.1532 |
|                     | QT_4bit_tqmse        |  1.0x   | 0.290   | 0.0012 | 0.0064 |
|                     | QT_3bit_tqmse        |  1.0x   | 0.345   | 0.0000 | 0.0001 |
|                     | QT_2bit_tqmse        |  1.0x   | 0.266   | 0.0000 | 0.0001 |
|                     | QT_1bit_tqmse        |  1.0x   | 0.280   | 0.0000 | 0.0001 |
+---------------------+----------------------+----------+---------+--------+--------+

Key: Speedup = IndexScalarQuantizer ms / IndexSQFastScan ms
     R@1, R@10 = IndexSQFastScan recall (matches SQ for all fallback types)
     Rerank shows rf=4 (rf=2 and rf=8 yield same recall, ~0.1x speedup variation)
     Dataset: SIFT1M, k=32
```

**IndexIVFSQFastScan** (ebfa072816df2fbd80a6b463a913866f862587f6)

Similar to the above, this commit introduces `IndexIVFSQFastScan`, an IVF variant of scalar quantized fast scan that extends the PQ4 FastScan SIMD infrastructure (vpshufb) to inverted-list indexes with scalar quantization. Uses BlockInvertedLists for packed 4-bit codes scanned via SIMD. Will continue to update the PR until it's feature complete.

Three code paths depending on quantizer type:

- Native 4-bit (QT_4bit, QT_4bit_uniform): codes pack directly into the SIMD block layout with no precision loss. ~12x speedup.

- Rerank (QT_6bit, QT_8bit, QT_8bit_uniform, QT_8bit_direct, QT_8bit_direct_signed): re-quantized to 4-bit for the SIMD coarse pass, then reranked using exact original-precision codes stored in a parallel ArrayInvertedLists. ~8-10x speedup, near-zero precision loss at rerank_factor=2.

- Fallback (QT_fp16, QT_bf16, TurboQuant, etc.): delegates to IndexIVF's InvertedListScanner path. Same behavior as IndexIVFScalarQuantizer — no fast scan benefit, but provides a unified interface for all quantizer types.

**benchmark (SIFT1M, d=128, k=32, SapphireRapids)**: `benchs/bench_index_ivf_sq_fastscan.py`

```text
+---------------------+----------------------+--------+-------------------------------------------+
| Category            | QType                | nprobe | Speedup | ms/q SQ | R@1 FS | R@10 FS     |
+---------------------+----------------------+--------+---------+---------+--------+-------------+
| Native 4-bit (SIMD) | QT_4bit              |      1 |   7.1x  |  0.007  | 0.4582 | 0.5456      |
|                     |                      |      4 |  10.4x  |  0.007  | 0.6903 | 0.8667      |
|                     |                      |     16 |  12.9x  |  0.022  | 0.7668 | 0.9885      |
|                     |                      |     64 |  12.1x  |  0.070  | 0.7725 | 0.9983      |
|                     |                      |    256 |  11.3x  |  0.230  | 0.7725 | 0.9983      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_4bit_uniform      |      1 |  15.8x  |  0.006  | 0.4397 | 0.5449      |
|                     |                      |      4 |  10.5x  |  0.007  | 0.6419 | 0.8650      |
|                     |                      |     16 |  10.4x  |  0.018  | 0.7113 | 0.9855      |
|                     |                      |     64 |  10.9x  |  0.063  | 0.7162 | 0.9951      |
|                     |                      |    256 |  11.0x  |  0.222  | 0.7162 | 0.9951      |
+---------------------+----------------------+--------+---------+---------+--------+-------------+
| Rerank (rf=4)       | QT_6bit              |      1 |   4.9x  |  0.007  | 0.4582 | 0.5456      |
|                     |                      |      4 |   4.8x  |  0.007  | 0.6903 | 0.8667      |
|                     |                      |     16 |   7.9x  |  0.018  | 0.7668 | 0.9885      |
|                     |                      |     64 |  10.4x  |  0.067  | 0.7725 | 0.9983      |
|                     |                      |    256 |  10.2x  |  0.221  | 0.7725 | 0.9983      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_8bit              |      1 |  17.1x  |  0.007  | 0.4582 | 0.5456      |
|                     |                      |      4 |  12.5x  |  0.008  | 0.6903 | 0.8667      |
|                     |                      |     16 |   8.9x  |  0.015  | 0.7668 | 0.9885      |
|                     |                      |     64 |   8.4x  |  0.049  | 0.7725 | 0.9983      |
|                     |                      |    256 |   9.3x  |  0.188  | 0.7725 | 0.9983      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_8bit_uniform      |      1 |  18.2x  |  0.007  | 0.4397 | 0.5449      |
|                     |                      |      4 |  11.1x  |  0.008  | 0.6419 | 0.8650      |
|                     |                      |     16 |   8.3x  |  0.014  | 0.7113 | 0.9855      |
|                     |                      |     64 |   7.8x  |  0.045  | 0.7162 | 0.9951      |
|                     |                      |    256 |   7.8x  |  0.164  | 0.7162 | 0.9951      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_8bit_direct       |      1 |   8.6x  |  0.007  | 0.4083 | 0.5426      |
|                     |                      |      4 |   7.0x  |  0.008  | 0.5987 | 0.8587      |
|                     |                      |     16 |   6.8x  |  0.013  | 0.6550 | 0.9756      |
|                     |                      |     64 |   6.1x  |  0.035  | 0.6581 | 0.9850      |
|                     |                      |    256 |   7.4x  |  0.150  | 0.6581 | 0.9850      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_8bit_direct_signed|      1 |  16.5x  |  0.006  | 0.4257 | 0.5447      |
|                     |                      |      4 |  10.1x  |  0.007  | 0.6286 | 0.8633      |
|                     |                      |     16 |   7.1x  |  0.012  | 0.6919 | 0.9831      |
|                     |                      |     64 |   5.7x  |  0.033  | 0.6963 | 0.9923      |
|                     |                      |    256 |   7.1x  |  0.145  | 0.6963 | 0.9923      |
+---------------------+----------------------+--------+---------+---------+--------+-------------+
| Fallback (~1.0x)    | QT_fp16              |      1 |   1.0x  |  0.007  | 0.5434 | 0.5462      |
|                     |                      |      4 |   1.1x  |  0.013  | 0.8607 | 0.8678      |
|                     |                      |     16 |   0.9x  |  0.035  | 0.9814 | 0.9902      |
|                     |                      |     64 |   0.9x  |  0.115  | 0.9912 | 1.0000      |
|                     |                      |    256 |   0.9x  |  0.481  | 0.9912 | 1.0000      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_bf16              |      1 |   1.0x  |  0.008  | 0.5427 | 0.5462      |
|                     |                      |      4 |   1.1x  |  0.013  | 0.8581 | 0.8678      |
|                     |                      |     16 |   1.0x  |  0.035  | 0.9778 | 0.9902      |
|                     |                      |     64 |   1.0x  |  0.115  | 0.9876 | 1.0000      |
|                     |                      |    256 |   1.0x  |  0.487  | 0.9876 | 1.0000      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_5bit_tq           |      1 |   1.0x  |  0.016  | 0.1866 | 0.3827      |
|                     |                      |     64 |   1.0x  |  0.794  | 0.2680 | 0.6248      |
|                     |                      |    256 |   1.0x  |  2.885  | 0.2680 | 0.6248      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_4bit_tq           |      1 |   1.0x  |  0.013  | 0.1182 | 0.2841      |
|                     |                      |     64 |   0.9x  |  0.618  | 0.1511 | 0.3994      |
|                     |                      |    256 |   0.9x  |  2.298  | 0.1511 | 0.3994      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_3bit_tq           |      1 |   1.1x  |  0.008  | 0.0526 | 0.1599      |
|                     |                      |     64 |   0.8x  |  0.092  | 0.0488 | 0.1696      |
|                     |                      |    256 |   0.8x  |  0.357  | 0.0487 | 0.1691      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_2bit_tq           |      1 |   1.2x  |  0.008  | 0.0450 | 0.1526      |
|                     |                      |     64 |   1.1x  |  0.072  | 0.0377 | 0.1534      |
|                     |                      |    256 |   1.1x  |  0.224  | 0.0376 | 0.1532      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_4bit_tqmse        |      1 |   1.0x  |  0.007  | 0.0619 | 0.1851      |
|                     |                      |     64 |   1.0x  |  0.072  | 0.0002 | 0.0010      |
|                     |                      |    256 |   0.9x  |  0.275  | 0.0000 | 0.0000      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_3bit_tqmse        |      1 |   1.0x  |  0.007  | 0.0621 | 0.1852      |
|                     |                      |     64 |   1.0x  |  0.090  | 0.0006 | 0.0032      |
|                     |                      |    256 |   1.0x  |  0.350  | 0.0000 | 0.0000      |
|                     +----------------------+--------+---------+---------+--------+-------------+
|                     | QT_2bit_tqmse        |      1 |   1.1x  |  0.007  | 0.0621 | 0.1852      |
|                     |                      |     64 |   1.0x  |  0.068  | 0.0015 | 0.0080      |
|                     |                      |    256 |   1.0x  |  0.244  | 0.0000 | 0.0000      |
+---------------------+----------------------+--------+---------+---------+--------+-------------+

Key: Speedup = IndexIVFScalarQuantizer ms / IndexIVFSQFastScan ms
     R@1, R@10 = IVFSQFastScan recall (matches IVF-SQ for all fallback types)
     Rerank shows rf=4 only (rf=2,8 yield similar results)
     Fallback shows nprobe=1,64,256 only (abbreviated for low-recall types)
     Dataset: SIFT1M, nlist=256, k=32
```


mulugetam:
@mnorris11 @mdouze Could you please take a look at this when you have a chance? Thanks!

faiss/IndexSQFastScan.h
```diff
@@ -0,0 +1,148 @@
...
+ * SIMD accumulators (safe for d <= 257).
+ *
+ * For reranked types, both original codes (for reranking)
+ * and packed 4-bit codes (for the SIMD scan) are stored.  The
+ * overselection ratio is controlled by rerank_factor (default 4).
+ */
+struct IndexSQFastScan : IndexFastScan {
+    ScalarQuantizer sq;
+
+    /// Original full-precision codes (ntotal * sq.code_size bytes).
```
mdouze: since this will be mainly useful for SQ4 indices, why keep these codes ? It doubles the storage while SQ4 is used at least in part to reduce mem usage

faiss/IndexSQFastScan.h
```diff
@@ -0,0 +1,148 @@
...
+    /// Populated for ALL types: used for reranking, reconstruction,
+    /// distance computation (HNSW), and range search.
+    std::vector<uint8_t> codes_8bit;
+
+    /// Overselection ratio for reranking.  The 4-bit fast scan
+    /// retrieves k * rerank_factor candidates, then exact
+    /// original-precision distances select the final top-k.
+    /// Higher = better recall, slower.  Default 4.
+    int rerank_factor = 4;
```
mdouze: could be a float, it is a float for IndexRefine 
https://github.com/facebookresearch/faiss/blob/main/faiss/IndexRefine.h#L36

tests/test_ivf_sq_fastscan.py
```diff
@@ -0,0 +1,434 @@
...
+import numpy as np
+import unittest
+import tempfile
+import os
+
+import faiss
+
+
+def make_data(n, d, seed=42):
+    rs = np.random.RandomState(seed)
```
mdouze: please use faiss.contrib.datasets.SyntheticDataset

tests/test_ivf_sq_fastscan.py
```diff
@@ -0,0 +1,434 @@
...
+        )
+        ref.train(xt)
+        ref.add(xb)
+        ref.nprobe = nprobe
+        D_ref, I_ref = ref.search(xq, k)
+
+        # Test: IndexIVFSQFastScan
+        quantizer_test = faiss.IndexFlatL2(d)
+        test = faiss.IndexIVFSQFastScan(
+            quantizer_test, d, nlist, qtype
```
mdouze: can reuse the same coarse quantizer and copy the scalar quantizer. 
Then the results should be the almost the same

tests/test_ivf_sq_fastscan.py
```diff
@@ -0,0 +1,434 @@
...
+        idx = faiss.IndexIVFSQFastScan(
+            quantizer, d, nlist, faiss.ScalarQuantizer.QT_4bit
+        )
+        idx.train(xt)
+        idx.add(xb)
+        idx.nprobe = nlist
+
+        D1, I1 = idx.search(xq, 5)
+
+        with tempfile.NamedTemporaryFile(suffix=".faissindex", delete=False) as f:
```
mdouze: use `index2 = deserialize_index(serialize_index(index))`

faiss/IndexSQFastScan.cpp
```diff
@@ -0,0 +1,734 @@
...
+    // IndexFastScan::search does not support params, pass nullptr.
+    idx_t k_rerank = std::min((idx_t)((int64_t)k * rerank_factor), ntotal);
+    k_rerank = std::max(k_rerank, k);
+
+    std::vector<float> coarse_dis(n * k_rerank);
+    std::vector<idx_t> coarse_ids(n * k_rerank);
+    IndexFastScan::search(
+            n, x, k_rerank, coarse_dis.data(), coarse_ids.data(), nullptr);
+
+    // Pass 2: rerank candidates with exact 8-bit distances,
```
mdouze: Since the storage is separate for the first 4 bits and the subsequent ones, why do the reranking within this class as opposed to using an IndexRefine ?

mdouze (commented):
Thanks for this contribution! My main comment is that it is not clear why the reranking must take place inside the class since the storage of the codes is not mutualized between the 4 first bits and the subsequent ones.

faiss/IndexSQFastScan.h
mulugetam: I have now completely removed codes_8bit from IndexSQFastScan. The class now only supports native 4-bit types and stores codes exclusively in the packed SIMD layout. reconstruct() and get_distance_computer() unpack on demand from the blocked layout via CodePacker::unpack_1().

faiss/IndexSQFastScan.cpp
mulugetam: The entire rerank path removed from IndexSQFastScan. The class is now scoped to native 4-bit only.

faiss/IndexSQFastScan.h
mulugetam: rerank_factor removed from IndexSQFastScan entirely (no longer relevant). In IndexIVFSQFastScan.h where it's still used, it was changed from int to float with a default value of 1 (see benchmark result).

tests/test_ivf_sq_fastscan.py
mulugetam: Done. Replaced make_data() helper with SyntheticDataset throughout the file.

tests/test_ivf_sq_fastscan.py
mulugetam: _compare_recall() now uses a shared quantizer object for both ref and test, and copies ref.sq to test.sq so both use identical coarse assignment and SQ parameters.

tests/test_ivf_sq_fastscan.py
mulugetam: Both IO test methods now use faiss.deserialize_index(faiss.serialize_index(idx)) instead of writing to a temp file and reading back.

mulugetam:
> Thanks for this contribution! My main comment is that it is not clear why the reranking must take place inside the class since the storage of the codes is not mutualized between the 4 first bits and the subsequent ones.

Thanks, @mdouze. I've updated the code to address all of your comments. Could you take a look when you have a chance?

Below is a summary of the benchmark results. The full details can be found [here](https://gist.github.com/mulugetam/fd2ccdd54f8156f5d7cefcbadc49fbe8).

```text
IndexSQFastScan (flat, native 4-bit only)
==========================================
qtype              SQ ms/q   FS ms/q   Speedup   SQ R@1   FS R@1   SQ R@10   FS R@10
-------------------------------------------------------------------------------------
QT_4bit              0.260     0.021     12.6x    0.775    0.771    0.999     0.999
QT_4bit_uniform      0.247     0.020     12.4x    0.720    0.718    0.996     0.995
```

```text
IndexIVFSQFastScan (nprobe=16)
===============================
qtype                   Mode       IVFSQ ms/q   FS ms/q   Speedup   IVFSQ R@1   FS R@1
---------------------------------------------------------------------------------------
QT_4bit                 native       0.024       0.002     13.8x      0.738      0.767
QT_4bit_uniform         native       0.020       0.003      7.7x      0.683      0.711
QT_6bit                 rerank       0.019       0.002      9.2x      0.917      0.928
QT_8bit                 rerank       0.016       0.004      3.6x      0.964      0.969
QT_8bit_uniform         rerank       0.015       0.002      7.4x      0.960      0.963
QT_8bit_direct          rerank       0.012       0.002      6.0x      0.102      0.981
QT_8bit_direct_signed   rerank       0.012       0.002      5.9x      0.938      0.602
QT_fp16                 fallback     0.035       0.042      0.8x      0.981      0.981
QT_bf16                 fallback     0.037       0.043      0.9x      0.978      0.978
```

mdouze:
You asked a question about HNSW that I cant' find back. 
FastScan variants or reranking do not make much sense with HNSW since  HSNW is based on distance computations with randomly accessed elements in a database. So for the HSNW storage_index it is better to keep the "flat" SQ layout.

mulugetam:
> You asked a question about HNSW that I cant' find back. FastScan variants or reranking do not make much sense with HNSW since HSNW is based on distance computations with randomly accessed elements in a database. So for the HSNW storage_index it is better to keep the "flat" SQ layout.

Thanks! I realized that shortly after asking the question (which I've since removed). I have also now deleted six leftover lines from index_factory.cpp to completely remove the "HNSW32,SQ4fs" wiring. Let me know if everything looks good now so I can squash/rebase, or if you have any further comments.

While I have your time, could you also take a look at #5297 and comment on my latest comment there? Thanks!

mulugetam:
@mnorris11 @mdouze squashed and rebased.

mnorris11:
@mulugetam  sorry for delay on several PRs; how does this compare to RaBitQFS / IVFRaBitQFS in speed/recall with the same bit widths? Say both at 4 and 6 bit if that is what is supported here?

mulugetam:
@mnorris11 Thank you!

Below are the results comparing **SQFastScan** and **RaBitQFastScan** on **SIFT1M** using **32 threads** with `rerank_factor` values of `1`, `2`, and `4`.

The benchmark was run using `benchs/bench_index_ivf_rabitq_fastscan.py`. Based on the results below, I've changed the default `rerank_factor` for `faiss/IndexIVFSQFastScan` from `1` to `2`.

### Summary

- **Speed:** SQFastScan is consistently 4-5x faster at every bit-width and nprobe. Its per-vector cost is near-zero beyond the SIMD shuffle; RaBitQ has additional distance correction overhead.
- **Recall:** At 4-bit, RaBitQ produces slightly better distance approximations (+3pts R@1). Once SQfs uses 6/8-bit reranking (rerank_factor >= 2), it matches or exceeds RaBitQ's recall.
- **Memory:** SQfs 65 MB reflects packed 4-bit codes only (actual footprint for 6/8-bit includes original codes in a parallel inverted list). RaBitQ stores auxiliary normalization data, hence 84-149 MB.
- **Rerank factor:** 1→2 gives the big R@10 jump (e.g., 0.83→0.95 at 6-bit); 2→4 adds only marginal improvement.

### Details

*Raw data:* https://gist.github.com/mulugetam/150af7cd68630744661412c03799292f 

### Flat (4-bit)

    Index                  ms/q    R@1    R@10    MB
    ---------------------  ------  -----  -----  ----
    SQ4fs                   0.06   0.77   0.83    64
    RaBitQfs4 (qb=8)       0.30   0.73   0.80    84

SQ4fs: 5x faster, +4pts recall, 24% less memory.

### IVF 4-bit (nprobe=64, SQfs rerank=1)

    Index                  ms/q    R@1    R@10    MB
    ---------------------  ------  -----  -----  ----
    IVF256,SQ4fs           0.017   0.77   0.83    65
    IVF256,RaBitQfs4       0.093   0.80   0.84    85

RaBitQfs4 has ~3pts better recall but is 5x slower.

### IVF 6-bit (nprobe=64, SQfs rerank=4)

    Index                  ms/q    R@1    R@10    MB
    ---------------------  ------  -----  -----  ----
    IVF256,SQ6fs           0.020   0.94   0.95    65
    IVF256,RaBitQfs6       0.091   0.95   0.95   117

Near-identical recall, SQ6fs is 4.5x faster and uses 44% less memory.

### IVF 8-bit (nprobe=64, SQfs rerank=4)

    Index                  ms/q    R@1    R@10    MB
    ---------------------  ------  -----  -----  ----
    IVF256,SQ8fs           0.020   0.98   0.99    65
    IVF256,RaBitQfs8       0.094   0.98   0.97   149

SQ8fs matches or beats recall while being 4.7x faster and using 56% less memory.

mdouze:
Hi 
Sorry for the delay... 
Let me do a new pass on the code...

faiss/impl/fast_scan/simd_result_handlers.h
```diff
@@ -410,7 +410,8 @@ struct SingleResultHandler : ResultHandlerCompare<C, with_id_map, SL> {
             if (!normalizers) {
                 dis[q] = idis[q];
             } else {
-                float one_a = 1 / normalizers[2 * q];
+                float a = normalizers[2 * q];
```
mdouze, line 413: what is the rationale of inverting the normalizer and adding a division FLOP?

faiss/IndexSQFastScan.h
```diff
@@ -0,0 +1,117 @@
...
+ *     PQ4 FastScan SIMD infrastructure (vpshufb / equivalent).
+ *     No precision loss.
+ *
+ * For higher-precision types (QT_8bit, QT_6bit, etc.), use
+ * IndexRefine(IndexSQFastScan(QT_4bit), IndexScalarQuantizer(QT_8bit))
+ * to get fast-scan with reranking.  For fallback types (QT_fp16,
+ * QT_bf16), use IndexScalarQuantizer directly.
+ *
+ * M = d subquantizers with 16 levels and uint16 SIMD accumulators
+ * (safe for d <= 257).
```
mdouze, line 32: safe = no overflow?

faiss/IndexIVFSQFastScan.h
```diff
@@ -0,0 +1,131 @@
...
+ *
+ *   - QT_6bit, QT_8bit, QT_8bit_uniform, QT_8bit_direct,
+ *     QT_8bit_direct_signed: re-quantised to 4-bit for the SIMD scan,
+ *     then reranked with exact original-precision distances.
+ *
+ *   - All other types (QT_fp16, QT_bf16, TurboQuant): fall back to the
+ *     ScalarQuantizer's own SIMD-optimised InvertedListScanner (same
+ *     behavior as IndexIVFScalarQuantizer, no fast-scan acceleration).
+ *
+ * For reranked types, both the original full-precision codes and the
```
mdouze, line 32: this comment is outdated no?

faiss/IndexSQFastScan.cpp
```diff
@@ -0,0 +1,504 @@
...
+    sq.train(n, x);
+    is_trained = true;
+}
+
+void IndexSQFastScan::add(idx_t n, const float* x) {
+    FAISS_THROW_IF_NOT(is_trained);
+    IndexFastScan::add(n, x);
+}
+
+void IndexSQFastScan::reset() {
```
mdouze, line 99: why redefine?

faiss/IndexSQFastScan.cpp
```diff
@@ -0,0 +1,504 @@
...
+    FAISS_THROW_IF_NOT(is_trained);
+
+    const IDSelector* sel = params ? params->sel : nullptr;
+
+    if (!sel) {
+        IndexFastScan::search(n, x, k, distances, labels, nullptr);
+        return;
+    }
+
+    // IDSelector path: unpack codes and use SQ scanner
```
mdouze, line 124: this is very inefficient. Better fail with FAISS_THROW so that the user has a clue

faiss/IndexSQFastScan.cpp
```diff
@@ -0,0 +1,504 @@
...
+    std::vector<float> recon(n * d);
+    sq.decode(code, recon.data(), n);
+    IndexFastScan::add(n, recon.data());
+}
+
+// -----------------------------------------------------------------------
+// permute_entries
+// -----------------------------------------------------------------------
+
+void IndexSQFastScan::permute_entries(const idx_t* perm) {
```
mdouze, line 298: can this be factorized with other FastScan implementations?

faiss/IndexSQFastScan.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+// -----------------------------------------------------------------------
+// get_distance_computer / get_FlatCodesDistanceComputer
+// -----------------------------------------------------------------------
+
+namespace {
+
+/// Wrapper that owns an unpacked code buffer and delegates distance
+/// computation to an underlying SQDistanceComputer.
+struct OwningFlatCodesDistanceComputer : FlatCodesDistanceComputer {
```
mdouze, line 327: what's the point of supporting this if the whole index needs to be decoded?

faiss/IndexIVFSQFastScan.cpp
```diff
@@ -0,0 +1,899 @@
...
+                    idx_t id = coarse_ids[i * k_coarse + j];
+                    if (id < 0) {
+                        continue;
+                    }
+
+                    idx_t lo = direct_map.get(id);
+                    idx_t list_no = lo_listno(lo);
+                    idx_t offset = lo_offset(lo);
+
+                    InvertedLists::ScopedCodes codes(
```
mdouze, line 542: So IIUC here the added value compared to an external reranker is that the reranking codes are in the same invlist layout as the 4-bit codes correct?

mdouze (commented):
Hi @mulugetam 

Sorry for the long delay... 
I still think that the code could be simplified: 
- I still don't see why the reranking logic is duplicated in IVFSQFastScan,  there is a compelling reason?
- do not decompress the packed codes before search, this makes the whole fast scan pointless. Better throw.
- some method implementations seem redundant

I hope these changes can be made and we'll be more reactive in the coming weeks.

faiss/impl/fast_scan/simd_result_handlers.h
mulugetam: To clarify: the inversion and division aren't new; main already had float one_a = 1 / normalizers[2 * q]. normalizers[2*q] is the LUT quantization scale a, stored un-inverted so it round-trips exactly through (de)serialization and pairs with the offset b as dis = b + idis * (1/a); the accumulator is in quantized units, so recovering the real distance needs that reciprocal once per query regardless.

The only change here is the a != 0 ? 1/a : 0 guard, which avoids inf/nan when a query yields a degenerate zero-range LUT — possible for SQ but not PQ, which is why the original didn't need it. Since it runs per query (not per database vector) it's off the hot path. I'll add a comment explaining this.

faiss/IndexSQFastScan.h
mulugetam: Yes; the M=d subquantizers accumulate uint8 LUT entries into uint16, so d*255 must fit in 65535, i.e. d ≤ 257. I will clarify the wording and a FAISS_THROW_IF_NOT(d ≤ 257) will be added in all constructors (flat + IVF) so it fails instead of silently overflowing.

faiss/IndexIVFSQFastScan.h
mulugetam: I will update the comment match the implementation. If the in-class rerank/fallback paths are removed (see below), it will be trimmed accordingly.

faiss/IndexSQFastScan.cpp
mulugetam: Agreed! It only forwards to IndexFastScan::reset(), so the override will be removed.

faiss/IndexSQFastScan.cpp
mulugetam: I will remove whole-index-unpack; the class will inherit IndexFastScan::search, which throws on any SearchParameters (including a selector), matching IndexPQFastScan. I will add a test asserting the throw.

faiss/IndexSQFastScan.cpp
mulugetam: Yes. The reconstruction-table build and LUT fill will be factored into a shared sq_fastscan_lut.h used by both IndexSQFastScan and IndexIVFSQFastScan, and will remove the duplicated local helpers.

faiss/IndexSQFastScan.cpp
mulugetam: Agreed! get_distance_computer/get_FlatCodesDistanceComputer will be removed (along with search1/range_search, which have the same decode-everything problem). The class will mirror IndexPQFastScan and inherit the base Index behavior (i.e. L2 gets a lazy GenericDistanceComputer via the packed-aware reconstruct; the rest throw).

faiss/IndexIVFSQFastScan.cpp
mulugetam: Correct. The original codes live in a parallel ArrayInvertedLists keyed by direct_map, so rerank reuses the coarse assignment rather than re-quantizing. But I benchmarked in-class rerank vs an external IndexRefine(IndexIVFSQFastScan(4bit), IndexScalarQuantizer(qtype)) on SIFT1M: recall is identical and in-class is not faster (usually slower). Given that, I'm inclined to drop the in-class rerank (and the fallback path) and just recommend IndexRefine. Does that work for you?

mulugetam:
> Hi @mulugetam
> 
> Sorry for the long delay... I still think that the code could be simplified:
> 
> * I still don't see why the reranking logic is duplicated in IVFSQFastScan,  there is a compelling reason?
> * do not decompress the packed codes before search, this makes the whole fast scan pointless. Better throw.
> * some method implementations seem redundant
> 
> I hope these changes can be made and we'll be more reactive in the coming weeks.

Thanks for the review @mdouze. If you concur with my reply on the reranking comment, I'll go ahead and trim IndexSQFastScan and IndexIVFSQFastScan down to native QT_4bit/QT_4bit_uniform only, mirroring IndexPQFastScan/IndexIVFPQFastScan. For reranking and higher-precision types, users can wrap with IndexRefine, exactly as the flat index already documents.

mulugetam:
@mdouze I've trimmed `IndexSQFastScan` and `IndexIVFSQFastScan` down to the native 4-bit types only (`QT_4bit`/`QT_4bit_uniform`, enforced at construction), mirroring `IndexPQFastScan`/`IndexIVFPQFastScan`. The factory exposes this as `SQ4fs`, and other SQ types are no longer accepted by these indexes.

mnorris11:
@mulugetam I have merged this one. I deleted the duplicated .cpp tests though, or merged them into the python unit tests.

Issue #5327, opened 2026-06-19
IndexFlatL2 range_search returns extra hits after zero-padding vectors


mmnhgo:
## Current Behavior

When using `IndexFlatL2` with `METRIC_L2`, appending all-zero dimensions to both stored vectors and query vectors can change the `range_search` result from the correct hit set containing only `id=3` to an incorrect hit set containing `id=2` and `id=3`.

This violates the zero-padding invariance of L2 distance.

For squared L2 distance, if the same number of zero dimensions are appended to all database vectors and all query vectors, all pairwise differences remain unchanged in the original dimensions, and the appended dimensions contribute exactly zero:

```text
For squared L2 distance, d([x, 0], [q, 0]) = d(x, q).
```

Therefore, the squared L2 distances, Top-k ordering, and `range_search` hit sets should remain unchanged.

However, zero-padding increases the dimensionality `d` enough that `IndexFlatL2` switches from the direct L2 accumulation path to the BLAS-based L2 path. The BLAS path computes L2 distance as:

```text
||x||^2 + ||y||^2 - 2 * <x, y>
```

In this case, numerical cancellation changes the true squared L2 distance of `id=2` from `5.0` to `0.0`. As a result, `range_search(radius=1.0)` incorrectly returns `id=2` as an extra hit, even though it should remain outside the radius.

This is not caused by approximate search. The reproduction uses `IndexFlatL2`, which is an exact flat index. It is also not caused by a ground-truth tie: `id=3` has a true squared L2 distance of `0.0`, while `id=2` has a true squared L2 distance of `5.0`.

This is related to #5285, but the trigger is different. #5285 is triggered by applying a large uniform translation to all vectors. This issue is triggered by appending trailing zero dimensions to all vectors. The original coordinates are unchanged, and the added dimensions are all zero.

## Actual Output on the Affected Path

```text
faiss_version: 1.14.3
blas_threshold_used: 3
original_dim: 2
padded_dim: 64001
radius: 1.0

exact_squared_l2:
[20000.0, 52.0, 5.0, 0.0]

exact_padded_squared_l2:
[20000.0, 52.0, 5.0, 0.0]
```

Original `range_search` result:

```text
lims      = [0, 1]
labels    = [3]
distances = [0.0]
```

Zero-padded `range_search` result:

```text
lims      = [0, 2]
labels    = [2, 3]
distances = [0.0, 0.0]
```

[...]

PR #5365, opened 2026-06-30
Fix BLAS L2 cancellation with zero padding (#5327)

codes-by-vamshi:
## Summary
This PR fixes a numerical cancellation issue in the BLAS L2 path where
`||x||^2 + ||y||^2 - 2<x,y>` can lose precision for large norms and small true distances
(e.g., zero-padded vectors), producing incorrect distances and range-search hits.

## Changes
- Added near-cancellation detection in BLAS-based L2 computation paths.
- For suspicious results (negative or very small relative to norm scale), recompute exact L2 with `fvec_L2sqr`.
- Kept tiny-negative clamping to `0`.
- Added regression test `test_zero_padding_invariance` reproducing #5327.

## Validation
- `pytest tests/test_index_accuracy.py -k zero_padding_invariance -q` passed.
- Full Python test suite passed locally with this patch.

### DIFF
--- a/faiss/utils/distances.cpp
+++ b/faiss/utils/distances.cpp
@@ -442,6 +442,13 @@ void exhaustive_L2sqr_blas_default_impl(
         y_norms = y_norms2;
     }
 
+    // Heuristic threshold used to detect catastrophic cancellation in the
+    // L2-from-IP decomposition:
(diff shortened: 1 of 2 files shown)


mnorris11:
See comment on https://github.com/facebookresearch/faiss/issues/5327#issuecomment-4911194346

PR #5370, opened 2026-07-02
Filter out-of-radius results in IndexRefine::range_search

### LINKED ISSUE #5367: IndexRefine range_search can return results outside the radius after refinement
## Summary

`IndexRefine::range_search` can return vectors whose final reported distance is outside the requested radius.

The public contract of `Index::range_search` says that it should return vectors with distance `< radius`. However, when using `IndexRefineFlat`, the final result can contain a vector whose returned refined distance is greater than the radius.

This leads to an incorrect range-search result: the result set contains a label that should have been excluded.

## Bug Type

Incorrect result

## Environment

```text
FAISS version: 1.14.2
Python version: 3.13.9
OS: Windows 11
Build method: pip faiss-cpu
```

## Reproducer

```python
import faiss
import numpy as np
import platform
import sys

print("faiss version:", getattr(faiss, "__version__", "unknown"))
print("python:", sys.version.split()[0])
print("os:", platform.platform())

xtrain = np.array([[0.0], [1.0]], dtype="float32")
xb = np.array([[0.49]], dtype="float32")
xq = np.array([[0.0]], dtype="float32")
radius = np.float32(0.23769901)

base = faiss.IndexScalarQuantizer(
    1, faiss.ScalarQuantizer.QT_6bit, faiss.METRIC_L2
)
base.train(xtrain)

index = faiss.IndexRefineFlat(base)
index.add(xb)

exact_distance = np.sum((xq[0] - xb[0]) ** 2).astype("float32")
print("radius:", float(radius))
print("exact squared L2 distance:", float(exact_distance))
print("expected labels because exact distance >= radius:", [])

lims, D, I = index.range_search(xq, float(radius))
print("actual lims:", lims.tolist())
print("actual labels:", I.tolist())
print("actual distances:", D.tolist())
print("returned distance > radius:", bool(len(D) and D[0] > radius))
```

## Actual Behavior

The only database vector is returned even though its final reported squared L2 distance is greater than the radius.

Example output:

```text
faiss version: 1.14.2
python: 3.13.9
os: Windows-11-10.0.26100-SP0
radius: 0.2376990169286728
exact squared L2 distance: 0.24010001122951508
expected labels because exact distance >= radius: []
actual lims: [0, 1]
actual labels: [0]
actual distances: [0.24010001122951508]
returned distance > radius: True
```

## Expected Behavior

No labels should be returned.

The exact squared L2 distance between the query and the only database vector is:

```text
0.24010001122951508
```

The radius is:

```text
0.2376990169286728
```

Since the exact distance is greater than the radius, the result should be empty.

Expected:

```text
lims: [0, 0]
labels: []
distances: []
```

## Why This Is a Bug

`range_search` is expected to return only vectors whose distance satisfies the radius predicate.

In this reproducer, the final returned distance is:

```text
0.24010001122951508
```

but the radius is:

```text
0.2376990169286728
```

Therefore, the returned vector does not satisfy `distance < radius`.

This is not caused by invalid inputs. The reproducer uses finite `float32` values, no NaN or Inf, no ties, a single query, and a single database vector.

It is also not just a display issue: the returned label remains in the result set even though the returned distance itself is outside the radius.

## Why This Is Not Expected Quantization Behavior

The base index is a scalar-quantized index, so it may produce approximate candidates. However, `IndexRefineFlat` reports the refined distance in the final result.

Once the final result reports a refined distance that is greater than the radius, the corresponding label should not remain in the range-search result.

In other words, the final distance and the final membership should be consistent with the `range_search` radius contract.

Vinay152003:
## Summary

`IndexRefine::range_search` recomputes exact distances for the candidates returned by the base index, but it never re-checks those distances against the radius. Since the base index selects candidates using *approximate* distances, a candidate can be within the radius according to the base index yet outside it once the exact distance is computed. As a result `range_search` can return entries whose reported (correct) distance is larger than the requested radius, which breaks the range-search contract.

Fixes #5367.

## Reproduction

```python
import numpy as np, faiss

d = 1
xb = np.array([[0.49]], dtype="float32")
xq = np.array([[0.0]], dtype="float32")
radius = 0.2376990169286728          # just below the exact L2 distance (0.2401)

base = faiss.IndexScalarQuantizer(d, faiss.ScalarQuantizer.QT_6bit, faiss.METRIC_L2)
base.train(np.linspace(0, 1, 256, dtype="float32").reshape(-1, 1))
base.add(xb)
flat = faiss.IndexFlat(d, faiss.METRIC_L2); flat.add(xb)

index = faiss.IndexRefine(base, flat)
lims, D, I = index.range_search(xq, radius)
print(D)   # [0.24010001] -> returned even though 0.2401 > radius
```

Before the fix the vector is returned with a distance that exceeds the radius; after the fix it is correctly excluded.

## Fix

After refinement, drop candidates whose exact distance falls outside the radius and rebuild `lims`, using the same comparison the base index applies (`CMax`/`CMin::cmp`), so L2 and similarity metrics are handled consistently. The compaction runs serially because the results are packed into a single contiguous per-query array; the parallel distance-recompute loop is unchanged.

## Tests

Extended `TestIndexRefineRangeSearch` to assert that every distance returned by the refined `range_search` is within the radius (the previous test only checked the distances were correct, not that they were in range), and added an `SQ6` case whose coarser quantization is more likely to produce near-boundary mismatches.

## Note on behavior

Results can only shrink: candidates that were previously returned in violation of the radius are now excluded. Counts from `range_search` may therefore be smaller than before for indexes wrapped in `IndexRefine`.

### DIFF
--- a/faiss/IndexRefine.cpp
+++ b/faiss/IndexRefine.cpp
@@ -156,7 +156,24 @@ void IndexRefine::range_search(
     SearchParameters* base_index_params =
             (params != nullptr) ? params->base_index_params : nullptr;
 
-    base_index->range_search(n, x, radius, result, base_index_params);
+    float k_factor = (params != nullptr) ? params->k_factor : this->k_factor;
+    FAISS_THROW_IF_NOT(k_factor >= 1);
+
+    const bool is_similarity = is_similarity_metric(metric_type);
+
+    // The base index selects candidates from approximate (e.g. PQ/PCA)
+    // distances, so the ideal radius to query it with differs from the
+    // refinement radius. Widen the base radius by k_factor (which plays
+    // the same role as in the k-NN search). Results are still filtered at
+    // the requested radius below, so k_factor affects recall, not
+    // correctness. k_factor == 1 queries the base index at `radius`,
+    // reproducing the previous behavior.
+    float base_radius = radius;
+    if (k_factor != 1) {
+        base_radius = is_similarity ? radius / k_factor : radius * k_factor;
+    }
+
+    base_index->range_search(n, x, base_radius, result, base_index_params);
 
 #pragma omp parallel if (n > 1)
     {
@@ -178,12 +195,11 @@ void IndexRefine::range_search(
         }
     }
 
-    // The base index selects candidates by approximate distance, so a
-    // candidate that looked within the radius can fall outside it once refined
-    // above. Drop those and rebuild lims, using the same comparison the base
-    // index applies. Done serially: the results are packed into one contiguous
-    // per-query array, so removing an entry shifts the ones after it.
-    const bool is_similarity = is_similarity_metric(metric_type);
+    // A candidate that looked within the radius under the base index's
+    // approximate distance can fall outside it once refined above. Drop those
+    // and rebuild lims, using the same comparison the base index applies. Done
+    // serially: the results are packed into one contiguous per-query array, so
+    // removing an entry shifts the ones after it.
     const std::vector<size_t> prev_lims(result->lims, result->lims + n + 1);
     size_t wp = 0;
     for (idx_t i = 0; i < n; i++) {
--- a/faiss/IndexRefine.cpp
+++ b/faiss/IndexRefine.cpp
@@ -177,6 +177,30 @@ void IndexRefine::range_search(
             }
         }
     }
+
+    // The base index selects candidates by approximate distance, so a
+    // candidate that looked within the radius can fall outside it once refined
+    // above. Drop those and rebuild lims, using the same comparison the base
+    // index applies. Done serially: the results are packed into one contiguous
+    // per-query array, so removing an entry shifts the ones after it.
+    const bool is_similarity = is_similarity_metric(metric_type);
+    const std::vector<size_t> prev_lims(
+            result->lims, result->lims + n + 1);
+    size_t wp = 0;
+    for (idx_t i = 0; i < n; i++) {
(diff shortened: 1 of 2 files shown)


mdouze:
It is actually an interesting question, because the optimal threshold on the base_index is probably different from the one on the refine_index. The base_index could use PQ or PCA compression which tend to need smaller distances. 

Therefore, there are a few possible design choices: 

1. as the code did before: do nothing but update the distances so that the caller can do something

2. as this PR does: re-filter the distances with the same threshold

3. find a way to pass the new threshold. This could be via k_factor that plays the same role for regular search 

My preference would be 3. then the default behavior with default k_factor=1 is the same as the current version of the PR, but the radius for refinement can be adjusted.

Vinay152003:
Thanks, that makes sense. I've updated the PR to option 3.

`range_search` now uses `k_factor` (the same field as the k-NN path) to widen the radius it queries the base index with, then still filters the refined distances at the requested radius:

- `base_radius = radius * k_factor` for distance metrics, `radius / k_factor` for similarity metrics
- `k_factor == 1` queries the base index at exactly `radius`, so the default behavior is identical to the previous version of the PR
- the final filter is always at the requested radius, so `k_factor` only affects recall, not correctness

I added a test that recall does not decrease as `k_factor` grows while every returned distance stays within the radius. On an SQ4 base, `k_factor == 1` reproduces the current behavior and larger values recover the recall lost to the base index's approximate distances.

Happy to adjust the similarity-metric scaling if you'd prefer a different convention there.

Issue #3908, opened 2024-10-04
Implement add_preassigned for GPU 


mdouze:
GPU search supports `search_preassigned` but not `add_preassigned` which is required when an arbitrary quantizer is used.

https://github.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuIndexIVF.h#L101

It would be useful to implement it.

asadoughi:
Do you have a specific use case in mind that this functionality would unblock?

mlomeli1:
Hi @asadoughi , I have created a toy example to repro and to showcase how our work in the research team could benefit from this functionality:
````
# Example: add_preassigned() not supported for gpu indexes
import os
import time

from faiss.contrib import ivf_tools

os.environ["CUDA_VISIBLE_DEVICES"] = "3,4,5"
import faiss
import faiss.contrib.torch_utils
import numpy as np
import torch

current_cache_index = 0
head_dim = 128
nlist = 1024
nprobe = 512
factory_str = f"IVF{nlist},Flat"
faiss_gpu_resources = faiss.StandardGpuResources()
co = faiss.GpuClonerOptions()
total_samples = 10000000
co.use_raft = False

xk = torch.rand(total_samples, head_dim).to(device="cuda", dtype=torch.float32)
faiss_index = faiss.index_factory(head_dim, factory_str, faiss.METRIC_INNER_PRODUCT)
faiss_index.by_residual = False
faiss.extract_index_ivf(faiss_index).nprobe = nprobe
faiss_index.is_trained = True
km = faiss.Kmeans(20, 50)
km.train(xk[:, :20].cpu())
alt_quantizer = km.index
a = alt_quantizer.search(xk[:, :20].cpu(), 1)[1].ravel()
# (optional) fake coarse quantizer
fake_centroids = np.zeros((nlist, head_dim), dtype="float32")
faiss_index.quantizer.add(fake_centroids)
gpu_index = faiss.index_cpu_to_gpu(
    faiss_gpu_resources,
    xk.device.index,
    faiss_index,
    co,
)
ivf_tools.add_preassigned(gpu_index, xk.cpu().numpy(), a.cpu().numpy())
````
Error:
````
AttributeError:'GpuIndexIVFFlat' object has no attribute 'add_core'
--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_1833376/3474978841.py in ?()
     39     xk.device.index,
     40     faiss_index,
     41     co,
[...]

PR #5372, opened 2026-07-03
Implement GPU IVF add_preassigned via add_core (implement #3908)

weiwch:
## Summary

This implements GPU IVF support for `ivf_tools.add_preassigned`, addressing #3908.

GPU IVF already supports `search_preassigned`, but `ivf_tools.add_preassigned` previously only worked for CPU IVF indexes because GPU IVF did not expose the corresponding `add_core` path. This implements the follow-up I mentioned in #3908 by refactoring my earlier [faiss_preassigned](https://github.com/weiwch/faiss_preassigned) work into the mainline-compatible `add_core` path. This PR adds `GpuIndexIVF::add_core(...)`, wires precomputed IVF list ids through the existing GPU add paging/copy path, and makes the existing `ivf_tools.add_preassigned` helper work for GPU IVF indexes without any change of `ivf_tools.py`.

The implementation follows the CPU IVF design:

- add `GpuIndexIVF::add_core(...)` with the same naming and argument order as `IndexIVF::add_core(...)`
- pass `precomputed_idx` through the existing GPU add paging/copy path
- add `IVFBase::addVectorsPreassigned(...)` for appending vectors to precomputed IVF lists

`contrib.ivf_tools.add_preassigned` can now call `index_ivf.add_core(...)` for both CPU and GPU IVF indexes. Compared with my earlier fork implementation, this does not introduce separate GPU-specific `add_preassigned*` APIs; the naming and entry point are aligned with the existing CPU IVF API.

## Implementation note

`GpuIndex::addPaged_` and `addPaged_ex_` were moved from `private` to `protected` so `GpuIndexIVF::add_core(...)` can reuse the existing GPU add paging path.

## Tests

Added a GPU IVFFlat add-preassigned test that:

- builds CPU and GPU IVF indexes
- assigns vectors with an external quantizer
- adds vectors through `ivf_tools.add_preassigned`
- compares CPU/GPU results using `search_preassigned`


mnorris11:
Thanks for this change!
Would you be able to add 1 small test that exercises the GpuIndexIVFPQ flow? Reason: it has `by_residual = true` by default. 
Then it looks good to me to merge.

weiwch:
I hadn’t used preassigned GpuIndexIVFPQ in my own project before, so I hadn’t tested it or added a test covering that flow. I’ll add a small test soon.

mnorris11:
Merged, thanks for implementing it!

PR #5373, opened 2026-07-03
Support IndexFlat range search for extra metrics

### LINKED ISSUE #5353: IndexFlat(METRIC_L1).range_search throws "metric type not supported" although k-NN search supports L1
## Description

`IndexFlat` supports `METRIC_L1` for k-NN search, but `IndexFlat::range_search` rejects the same metric with:

```text
metric type not supported
```

This creates an inconsistency in the same exact flat index: `search()` computes correct L1 distances, while `range_search()` refuses to run for `METRIC_L1`.

## How to reproduce this bug?

This bug can be reproduced with a plain CPU `IndexFlat` using `METRIC_L1`.

```python
import faiss
import numpy as np
import platform
import sys

print("FAISS version:", getattr(faiss, "__version__", "unknown"))
print("Python version:", sys.version.split()[0])
print("OS:", platform.platform())

xb = np.array([
    [0.0, 0.0],  # L1 distance 0
    [2.0, 0.0],  # L1 distance 2
    [0.0, 3.0],  # L1 distance 3
    [5.0, 0.0],  # L1 distance 5
], dtype=np.float32)

xq = np.array([[0.0, 0.0]], dtype=np.float32)
radius = 3.0

index = faiss.IndexFlat(2, faiss.METRIC_L1)
index.add(xb)

D, I = index.search(xq, 4)

expected_distances = np.abs(xb - xq[0]).sum(axis=1)
expected_labels = np.flatnonzero(expected_distances < radius).astype(np.int64)
expected_D = expected_distances[expected_labels]

print("kNN labels:", I.tolist())
print("kNN distances:", D.tolist())

print("Expected range labels:", expected_labels.tolist())
print("Expected range distances:", expected_D.tolist())

try:
    lims, D2, I2 = index.range_search(xq, radius)
    print("Actual range lims:", lims.tolist())
    print("Actual range labels:", I2.tolist())
    print("Actual range distances:", D2.tolist())
except RuntimeError as e:
    print("Actual exception:", str(e))
```

## What is the expected behavior?

`IndexFlat(2, METRIC_L1).range_search(xq, 3.0)` should return the database vectors whose L1 distance from the query is strictly smaller than the radius.

For the example above, the L1 distances are:

```text
[0.0, 2.0, 3.0, 5.0]
```

Since FAISS range search uses `distance < radius`, the vector at exact distance `3.0` should be excluded.

Expected labels:

```text
[0, 1]
```

Expected distances:

```text
[0.0, 2.0]
```

The same `IndexFlat` object already computes the correct L1 distances through k-NN search:

```text
kNN labels: [[0, 1, 2, 3]]
kNN distances: [[0.0, 2.0, 3.0, 5.0]]
```

Therefore, range search on the same exact flat L1 index should return the vectors within the L1 radius instead of rejecting the metric.

## What is the actual behavior?

`IndexFlat.search` correctly returns L1 distances:

```text
kNN labels: [[0, 1, 2, 3]]
kNN distances: [[0.0, 2.0, 3.0, 5.0]]
```

But `IndexFlat.range_search` throws:

```text
Actual exception: metric type not supported
```

So the metric is accepted by `search()` but rejected by `range_search()` on the same index.

## Why this is a bug?

FAISS documents/comments flat/brute-force indices as supporting additional metric types for vector comparison, including `METRIC_L1`.

The reproducer uses a plain exact `IndexFlat` with legal `float32` vectors. There is no ANN recall variation, no quantization, no randomness, no NaN/Inf, and no boundary ambiguity: the radius is `3.0`, and the expected returned distances are `0.0` and `2.0`.

## Server Version / FAISS Version

FAISS version: 1.14.2
Python version: 3.13.9
OS: Windows 11
Installation: pip `faiss-cpu`
Index type: `IndexFlat`
Metric: `METRIC_L1`
Component: `faiss/IndexFlat.cpp`

SakshamKapoor2911:
Summary:
- add range_search support for IndexFlat metrics handled by VectorDistance, including METRIC_L1
- preserve existing range result handler semantics for distance and similarity metrics, including ID selectors
- add a METRIC_L1 regression test for the strict radius boundary from #5353

Fixes #5353.

Test Plan:
- PYTHONPATH=tests python -m pytest -q tests/test_index.py::TestRangeSearch tests/test_extra_distances.py::TestKNN_NONE
- manual issue reproducer: IndexFlat(2, METRIC_L1).range_search(xq, 3.0) returns labels [0, 1] and distances [0.0, 2.0]
- git diff --check

### DIFF
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -77,6 +77,7 @@ void IndexFlat::range_search(
             range_search_L2sqr(x, get_xb(), d, n, ntotal, radius, result, sel);
             break;
         default:
+            metric_type_from_int(static_cast<int>(metric_type));
             range_search_extra_metrics(
                     x,
                     get_xb(),
--- a/faiss/IndexFlat.cpp
+++ b/faiss/IndexFlat.cpp
@@ -77,7 +77,17 @@ void IndexFlat::range_search(
             range_search_L2sqr(x, get_xb(), d, n, ntotal, radius, result, sel);
             break;
         default:
-            FAISS_THROW_MSG("metric type not supported");
+            range_search_extra_metrics(
+                    x,
+                    get_xb(),
+                    d,
+                    n,
+                    ntotal,
+                    metric_type,
+                    metric_arg,
+                    radius,
+                    result,
+                    sel);
     }
 }
 
--- a/faiss/utils/extra_distances.cpp
+++ b/faiss/utils/extra_distances.cpp
@@ -12,6 +12,7 @@
 
 #include <omp.h>
 #include <algorithm>
+#include <limits>
 
 #include <faiss/impl/AuxIndexStructures.h>
 #include <faiss/impl/DistanceComputer.h>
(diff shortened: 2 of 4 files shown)


mnorris11:
sorry I just now saw this. I added the fallback in #5391, so I will go ahead and close this one

PR #5383, opened 2026-07-07
fix: bs add n fix valid counter skip

lyang24:
isn't valid counter == 0 means all vectors are filtered out?


mnorris11:
Thanks for the fix, merging it

PR #5385, opened 2026-07-07
Vectorize RaBitQ fast-scan handler adjustment (1 bit, no filter)

lyang24:
This patch adds an AVX512 fast path for the 1-bit IndexIVFRaBitQFastScan result handler.

The PQ4 fast-scan kernel already computes distances for 32 candidates at a time, but the RaBitQ handler was converting those 32 uint16 distances back into a scalar per-candidate loop to apply the 1-bit distance correction and heap threshold check.

  For the common 1-bit, non-centered, no-selector, full-block path, this patch keeps the correction step vectorized:

  - converts 16 uint16 distances to float at a time
  - loads/deinterleaves the per-vector RaBitQ aux factors
  - computes adjusted distances with AVX512
  - compares all lanes against the current heap threshold
  - drains only surviving lanes into the scalar heap update path

  The scalar fallback is unchanged for multibit, selectors, partial blocks, centered mode, and non-AVX512 builds.

  ## Validation

  Tested on AWS Sapphire Rapids with a 1M Cohere IVF_RABITQ_FASTSCAN index.

  Config:

  - nprobe=64
  - k=100
  - nq=1000
  - 5 rounds
  - 8 search threads

  Results:
  baseline 312.87
  patched 322.61

  This is about a 3% end-to-end QPS improvement with unchanged results.

### DIFF
--- a/faiss/IndexIVFRaBitQFastScan.h
+++ b/faiss/IndexIVFRaBitQFastScan.h
@@ -9,6 +9,9 @@
 
 #include <memory>
 #include <vector>
+#if defined(__AVX512F__) && defined(__AVX512BW__) && defined(__AVX512VL__)
+#include <immintrin.h>
+#endif
 
 #include <faiss/IndexIVFFastScan.h>
 #include <faiss/IndexIVFRaBitQ.h>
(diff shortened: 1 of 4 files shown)


mdouze:
Please refer to https://github.com/facebookresearch/faiss/blob/main/faiss/docs/simd_dynamic_dispatch_migration.md on how to integrate SIMD code in Faiss.

lyang24:
> Please refer to https://github.com/facebookresearch/faiss/blob/main/faiss/docs/simd_dynamic_dispatch_migration.md on how to integrate SIMD code in Faiss.

Thanks will read and get back later

lyang24:
> Please refer to https://github.com/facebookresearch/faiss/blob/main/faiss/docs/simd_dynamic_dispatch_migration.md on how to integrate SIMD code in Faiss.

fixed but would like to wait for https://github.com/facebookresearch/faiss/pull/5396 and then run benchmark again

lyang24:
perf gain cannot be proven after repeated testing closing

PR #5396, opened 2026-07-10
Optimize RaBitQ fastscan query setup (openblas + gcc) only

lyang24:
## What

My results are specific to GCC 11.4, OpenBLAS pthread 0.3.20, AVX-512 dispatch, and one-query-per-call search. They should not be interpreted as a universal speedup. 

Two hot-path fixes for RaBitQ FastScan search, from profiling Cohere 1M (d=1024, nprobe=64):

1. **LUT/query quantization**: `compute_LUT_uint8` and `compute_query_factors` call
   `std::roundf` per element (~330k libm calls per query, 16.9% of cycles in
   `__roundf`+PLT alone — half-away-from-zero rounding has no x86 instruction, so it
   never inlines). Replaced with an inline `x + 0.5f` + truncate helper, valid since
   these inputs are non-negative and range-bounded by construction. (recall impact was small enough to be unnoticed)

2. **Single-query pretransform**: `LinearTransform::apply_noalloc` always calls `sgemm`,
   so for n=1 OpenBLAS repacks the whole rotation matrix every query
   (`sgemm_incopy` was 12.2% of cycles vs 2.7% for the actual kernel). Added an
   `sgemv` path for n=1.

Note: rounding is not bit-exact with `roundf` for inputs within 1-2 ulp of a
`k + 0.5` boundary (±1 quantization code); ties agree. gemv also differs from gemm
in accumulation order. Neither affects recall (below), 4/10000 result ids changed.

image

## Results

Cohere 1M, IVF_RABITQ_FASTSCAN, k=100, nprobe=64, single thread:

- QPS 299.5 → 546.4 (median of 5)
- perf stat: cycles -36.7%, instructions -36.4%, L1D misses -62.2%, IPC unchanged

Recall vs exact ground truth (recomputed by brute force; the HDF5-provided
neighbors are invalid for IP/COSINE): recall@1/10/100 = 0.95 / 0.924 / 0.6554,
identical before and after.


lyang24:
cc @alibeklfc @mnorris11 
may i get a review

alibeklfc:
Hi @lyang24, thanks for the contribution. We confirmed it is recall-neutral on SIFT1M and GIST1M.

We benchmarked it on SIFT1M and GIST1M (single thread) and did not see a significant speedup, so we were not able to reproduce the reported gain. Could you share how you built it: compiler and version, the flags used, and the SIMD variant?

Separately: our tests are almost all in Python. Is there a reason test_rabitq_utils.cpp needs to be C++? If not, please rewrite it in Python. Thank you!

lyang24:
> Hi @lyang24, thanks for the contribution. We confirmed it is recall-neutral on SIFT1M and GIST1M.
> 
> We benchmarked it on SIFT1M and GIST1M (single thread) and did not see a significant speedup, so we were not able to reproduce the reported gain. Could you share how you built it: compiler and version, the flags used, and the SIMD variant?
> 
> Separately: our tests are almost all in Python. Is there a reason test_rabitq_utils.cpp needs to be C++? If not, please rewrite it in Python. Thank you!

Hi @alibeklfc 

Thanks for checking. I re-ran this from clean, isolated baseline and patched builds. The important reproduction condition is single-query search: the SGEMV change is guarded by LinearTransform::apply_noalloc(n == 1), so a batched index.search(xq, k) benchmark continues to use SGEMM and bypasses the main optimization.

For transparency, my benchmark is currently a Knowhere Python script rather than an upstream Faiss test. It loads HDF5 train/test arrays and invokes Search once per query:

      qds = [knowhere.ArrayToDataSet(xq[i:i + 1]) for i in range(xq.shape[0])]
      for ds in qds:
          ans, _ = idx.Search(ds, cfg_json, null_bitset)

I build the index once with the baseline binary, then reuse that exact serialized index for both binaries:

      env OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 \
        LD_LIBRARY_PATH=$BASE/build/Release:$COMMON \
        PYTHONPATH=$PY \
        python3 bench_cohere_qps_saved.py \
          --mode build \
          --dataset cohere-wiki-1024-ip.hdf5 \
          --index-name IVF_RABITQ_FASTSCAN \
          --index-file cohere_1m_ivf_rabitq_fastscan_nlist1024.index \
          --metric-type IP \
          --nb 1000000 \
          --nlist 1024

 Then, for each binary, I run:

      env OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 \
        LD_LIBRARY_PATH=$VARIANT/build/Release:$COMMON \
        PYTHONPATH=$PY \
        python3 bench_cohere_qps_saved.py \
          --mode search \
          --dataset cohere-wiki-1024-ip.hdf5 \
          --index-name IVF_RABITQ_FASTSCAN \
          --index-file cohere_1m_ivf_rabitq_fastscan_nlist1024.index \
          --metric-type IP \
          --nq 1000 \
          --k 100 \
          --nprobe 64 \
          --refine-k 1.0 \
          --warmup 100 \
          --rounds 1

 I alternate runs in this order:

      base, patch, patch, base, base, patch

  Environment (**aws c7i 2xlarge**):
  - Intel Xeon Platinum 8488C
  - Ubuntu 22.04
  - GCC 11.4.0
  - OpenBLAS pthread
  - AVX2/AVX-512 runtime-dispatch build

  Clean A/B mean QPS:
  - Cohere 1M, IP, d=1024: 276.4 -> 431.0 (+55.9%)
  - GIST1M, L2, d=960: 288.3 -> 374.9 (+30.1%)
  - SIFT1M, L2, d=128: 915.4 -> 986.3 (+7.7%)

The result checksum matched for every run. The smaller SIFT gain is expected because the LUT/query quantization work scales with dimension, while FastScan scan and heap overhead does not.

historical profiles of this workload showed __roundf at about 13.1% of cycles and SGEMM input packing at about 12.2%; after the patch, roundf disappears and the n==1 transform uses SGEMV instead.

lyang24:
i might have an idea why this cannot be reproduced, after some digging i think the difference maybe i used an older gcc version? and for linear algebra i used openBLAS and meta uses mkl?

The SGEMV path only activates when LinearTransform::apply_noalloc receives n == 1. A batched search call uses SGEMM instead, so batching can eliminate the SGEMV benefit. This is relevant to online serving, where queries are commonly issued one at a time.

Batching does not explain the rounding optimization. compute_LUT_uint8 still processes each query independently in the batched path. The relevant condition there is compiler/code generation: in my GCC 11.4 build without fast-math, std::roundf remained a libm call and was a significant baseline hotspot. A compiler or flag set that inlines or vectorizes roundf has substantially less headroom, so this part of the gain is compiler-dependent.

My results are specific to GCC 11.4, OpenBLAS pthread 0.3.20, AVX-512 dispatch, and one-query-per-call search. They should not be interpreted as a universal speedup.

PR #5406, opened 2026-07-11
Parse IVF HNSW coarse quantizer with any M in get_code_size

Vinay152003:
## Summary

`contrib.factory_tools.get_code_size` matches only `_HNSW32` in its IVF regex, so it raises `cannot parse` for valid factory strings whose IVF coarse quantizer is an HNSW with a different `M`, e.g. `IVF100_HNSW16,Flat`. These names are produced by `reverse_index_factory` in the same file (`IVF{nlist}_HNSW{M}`), so the two functions are inconsistent.

## Reproduction

```python
from faiss.contrib import factory_tools
factory_tools.get_code_size(64, "IVF100_HNSW32,Flat")  # 256
factory_tools.get_code_size(64, "IVF100_HNSW16,Flat")  # RuntimeError: cannot parse IVF100_HNSW16,Flat
```

## Fix

Match `_HNSW\d+` instead of `_HNSW32`. The coarse quantizer does not contribute to the per-vector code size (that is determined by the sub-index after the comma), so the result is unchanged for the default `_HNSW32` and correct for other `M`.

## Tests

Adds `test_ivf_hnsw_quantizer` to `TestCodeSize`, covering `_HNSW16/32/64` with a `Flat` sub-index and an `IVF..._HNSW16,PQ8` case. Existing `get_code_size` cases are unaffected.


mnorris11:
It got handled in https://github.com/facebookresearch/faiss/pull/5416, apologies, it was done before we got to reviewing this PR.

PR #5412, opened 2026-07-12
Fuse RaBitQ AND-dot and popcount scan

lyang24:
Fuse RaBitQ’s AND-dot and doc-side popcount into a single SIMD pass, avoiding a second scan of the same binary code.


alexanderguzhva:
the optimization makes sense

Issue #5276, opened 2026-06-04
[Bug] Data race in NNDescent::update() — unlocked read of other.pool.back() before acquiring lock


wonder-fly:
# Summary
In NNDescent::update(), other.pool.back() is read before acquiring other.lock, while another thread may be concurrently modifying other.pool inside Nhood::insert(). This is a data race that can cause undefined behavior in multithreaded builds.

# Affected File
faiss/impl/NNDescent.cpp

# Root Cause
Nhood::insert() modifies pool under lock:

```cpp
void Nhood::insert(int id, float dist) {
    LockGuard guard(lock);       // lock acquired here
    ...
    std::pop_heap(pool.begin(), pool.end());
    pool[pool.size() - 1] = Neighbor(id, dist, true);
    std::push_heap(pool.begin(), pool.end());  // pool structure changes
}
Meanwhile, NNDescent::update() reads other.pool.back() without holding other.lock:

```cpp
auto& other = graph[nn.id];

if (nn.distance > other.pool.back().distance) {  // ❌ unlocked read
    LockGuard guard(other.lock);                  // lock acquired too late
    other.rnn_new.push_back(n);
}
```
# Race timeline
Thread A (update)                     Thread B (join → insert)
────────────────────────────────────  ────────────────────────────────────
read other.pool.back()                lock(other.lock)
                                      pop_heap(other.pool)  ← pool mutated
other.pool.back() is now stale ❌     push_heap(other.pool)
                                      unlock(other.lock)
std::pop_heap / std::push_heap rearrange elements inside the vector, so the reference returned by back() can be invalidated while Thread A is reading it.

The same issue exists in both branches of the nn.flag check (lines covering rnn_new and rnn_old).

# Affected Code Locations
Both occurrences are in the for (int l = 0; l < node.M; ++l) loop inside NNDescent::update():

## Occurrence 1 — nn.flag == true branch:

```cpp
if (nn.distance > other.pool.back().distance) {  // ❌ no lock
    LockGuard guard(other.lock);
    if (other.rnn_new.size() < static_cast<size_t>(R)) {
        other.rnn_new.push_back(n);
    } else {
        int pos = rng() % R;
        other.rnn_new[pos] = n;
    }
}
```
## Occurrence 2 — nn.flag == false branch:

```cpp
if (nn.distance > other.pool.back().distance) {  // ❌ no lock
    LockGuard guard(other.lock);
    if (other.rnn_old.size() < static_cast<size_t>(R)) {
[...]

PR #5413, opened 2026-07-12
Fix data race in NNDescent::update()

RohanMali2003:
other.pool.back() was read before acquiring other.lock, racing with concurrent pop_heap and push_heap calls in Nhood::insert(). Moved the lock acquisition before the read in both rnn_new and rnn_old branches. Fixes issue #5276.

I was not able to build the project locally, so this change has not been compiled or run through the test suite. The fix mirrors the reordering proposed in the linked issue exactly, as proposed by reporter @wonder-fly : moving LockGuard guard(other.lock) before the other.pool.back() read in both the rnn_new and rnn_old branches. Happy to iterate based on CI results or review feedback. Thanks!


RohanMali2003:
Heya, @alibeklfc , just checking in to see if there are any updates on the internal review or if you need any adjustments from my side. Thanks!

PR #5487, opened 2026-07-30
Fix IndexIVFPQ losing polysemous_ht/do_polysemous_training on I/O round-trip

### LINKED ISSUE #2120: polysemous_ht resets to 0 when loading from disk
# Summary

I run a hyperparameter sweep for an IVF,PQ index which provides optimal values for nprobe,ht.  I then set those
parameters for the index and save it to disk. When I read from index, the nprobe is still the same, but the ht has
been reset to 0.

# Platform
OS: Ubuntu 20.04.3 LTS

Faiss version: 1.7.1

Installed from: anaconda, pytorch channel

Running on:
- [X] CPU
- [ ] GPU

Interface: 
- [ ] C++
- [X] Python

# Reproduction instructions
import faiss

n = 10000
d = 64
nlist = 100
X = faiss.randn((n, d))

index = faiss.index_factory(d, f"IVF{nlist},PQ8")
print(f"{index.do_polysemous_training=:}")
print(f"{index.by_residual=:}")
index.train(X)
index.add(X)

params = faiss.ParameterSpace()
params.set_index_parameters(index, "nprobe=10,ht=32")
faiss.write_index(index, "testing.index")

index2 = faiss.read_index("testing.index")
assert index.nprobe == index2.nprobe, \
    f"{index.nprobe=:}, {index2.nprobe=:}"
assert index.polysemous_ht == index2.polysemous_ht, \
    f"{index.polysemous_ht=:}, {index2.polysemous_ht=:}"

"""
index.do_polysemous_training=True
index.by_residual=True
Traceback (most recent call last):
  File "faiss_bug.py", line 21, in <module>
    assert index.polysemous_ht == index2.polysemous_ht, \
AssertionError: index.polysemous_ht=32, index2.polysemous_ht=0
"""

Juanpacol:
## Summary

IndexIVFPQ never serialized its `polysemous_ht` and `do_polysemous_training` fields. The `write_index()` function wrote only the fields common to the IVFPQ family (`by_residual`, `code_size`, the ProductQuantizer, and the inverted lists) under the "IwPQ"/"IwQR" fourcc, so `read_index()` had no data to restore these two members. They silently reverted to their constructor defaults (`do_polysemous_training = false`, `polysemous_ht = 0`) on every load.

A caller who trained polysemous codes and tuned a Hamming threshold via ParameterSpace (`nprobe=10,ht=32`) would get back an index that searches as if polysemous filtering were never configured, with no error or warning to indicate the parameters were dropped.

### Reproduction (Issue #2120)

```python
import faiss
n, d, nlist = 10000, 64, 100
X = faiss.randn((n, d))
index = faiss.index_factory(d, f"IVF{nlist},PQ8")
index.train(X)
index.add(X)

params = faiss.ParameterSpace()
params.set_index_parameters(index, "nprobe=10,ht=32")
faiss.write_index(index, "testing.index")

index2 = faiss.read_index("testing.index")
assert index.polysemous_ht == index2.polysemous_ht  # FAILED: 32 != 0
```

### Solution

Introduced new fourcc tags:
- **"IwPh"** for IndexIVFPQ (replacing "IwPQ")
- **"IwQh"** for IndexIVFPQR (replacing "IwQR")

These new tags additionally write `do_polysemous_training` and `polysemous_ht` right after the ProductQuantizer, before the inverted lists.

The `read_ivfpq()` function recognizes the new tags and reads the two extra fields only when present. Existing "IwPQ"/"IwQR"/"IvPQ"/"IvQR" files continue to load exactly as before (fields default to `false`/`0`, same as pre-fix behavior), preserving full backward and forward binary compatibility with indexes already persisted to disk.

### Testing & Verification

- **C++ suite**: 284/285 tests pass (1 unrelated SIMD-hardware skip)
- **Python suite** (`tests/test_io.py`, `tests/test_index.py`): 79 passed
- **New regression test**: `TestIORoundTrip.test_index_ivfpq_polysemous_ht()` validates that polysemous parameters survive a serialize/deserialize cycle
- **Backward compatibility**: Legacy-format regression test `TestIVFPQRead.test_reader` confirms old on-disk indexes deserialize correctly
- **Build**: Verified against libfaiss built from source (CMake + SWIG, CPU-only, Accelerate BLAS/LAPACK)

Closes #2120

### DIFF
--- a/faiss/impl/index_read.cpp
+++ b/faiss/impl/index_read.cpp
@@ -1550,10 +1550,13 @@ static std::unique_ptr<IndexIVFPQ> read_ivfpq(
         uint32_t h,
         int io_flags) {
     bool legacy = h == fourcc("IvQR") || h == fourcc("IvPQ");
+    // "IwPh"/"IwQh" are the only formats that store the polysemous
+    // training parameters (see GH issue #2120).
+    bool has_polysemous = h == fourcc("IwPh") || h == fourcc("IwQh");
 
     IndexIVFPQR* ivfpqr = nullptr;
     std::unique_ptr<IndexIVFPQ> ivpq;
-    if (h == fourcc("IvQR") || h == fourcc("IwQR")) {
+    if (h == fourcc("IvQR") || h == fourcc("IwQR") || h == fourcc("IwQh")) {
         ivpq = std::make_unique<IndexIVFPQR>();
         ivfpqr = static_cast<IndexIVFPQR*>(ivpq.get());
     } else {
@@ -1565,6 +1568,11 @@ static std::unique_ptr<IndexIVFPQ> read_ivfpq(
     READ1_BOOL(ivpq->by_residual);
     READ1(ivpq->code_size);
     read_ProductQuantizer(&ivpq->pq, f);
+
+    if (has_polysemous) {
+        READ1_BOOL(ivpq->do_polysemous_training);
+        READ1(ivpq->polysemous_ht);
+    }
 
     if (legacy) {
         ArrayInvertedLists* ail = set_array_invlist(ivpq.get(), ids);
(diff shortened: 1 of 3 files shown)


mnorris11:
Sorry for the delays here. I will remove the lock based HNSW path soon in the next couple weeks, then we can make a new release which can unblock various serialization PRs like this one. We eventually want to get rid of the CMake write --> Conda read test which is more of a "forward compatibility" test, which doesn't make sense.

PR #5476, opened 2026-07-28
Fix/win simd detection

Binozo:
Hi,

while testing on windows x64 systems I've noticed that the `dd` optimization compilation target doesn't work properly. Causing crashes on non-avx2 systems for instance.

This pr fixes the SIMD level detection.
MSVC doesn't define `__x86_64__` but does define `_M_X64`. This is what this patch is based on.

I got some nice [GitHub actions](https://github.com/Binozo/Faiss.NET/actions/runs/30251011773/job/89928775910) and tested the patch on it.

PR #5497, opened 2026-08-03
Fix: Windows SIMD detection

Binozo:
Hi,

while testing on windows x64 systems I've noticed that the `dd` optimization compilation target doesn't work properly. Causing crashes on non-avx2 systems for instance.

This pr fixes the SIMD level detection.
MSVC doesn't define `__x86_64__` but does define `_M_X64`. This is what this patch is based on.

I got some nice [GitHub actions](https://github.com/Binozo/Faiss.NET/actions/runs/30251011773/job/89928775910) and tested the patch on it.

##### Based on #5476 but reopened because I made a mistake while rebasing


faiss/utils/simd_levels.cpp
```diff
@@ -67,42 +73,78 @@ static bool has_sve() {
 }
 #endif
 
+namespace {
+
+#if defined(__x86_64__) || defined(_M_X64)
```
mnorris11, line 78: IIUC, the minimal solution is to add the _M_X64 and _MSC_VER checks like this line, is that right? If so, should we revert the new functions and `regs` refactor etc? then it would be easier to merge.

Binozo: Indeed the `_M_X64` and `_MSC_VER` checks are key but they are not sufficient as widening the guard would not compile under MSVC at all because GNU asm is rejected by MSVC.

Also, `cpuid_count` has 5 call sites and `xgetbv0` has 2 (including `detect_x86_uarch_flags` which is used in static builds too), so the MSVC intrinsic version would otherwise need to be duplicated inline at 7 sites. So I decided to add a small wrapper to avoid that duplication

tests/test_residual_quantizer.py
```diff
@@ -849,7 +849,8 @@ def test_search_L2(self):
                 self.assertLess((Iref != I2).sum(), Iref.size * 0.05)
             else:
                 inter_2 = faiss.eval_intersection(I2, gt)
-                self.assertGreaterEqual(inter_ref, inter_2)
+                # many near-ties, quantized norms can marginally beat the reference
```
mnorris11, line 852: I think the simd_levels.cpp change is OK, but why is this needed?

Binozo: `test_search_L2` never ran in windows avx2 mode before this patch. Running it now uncovers a too-exact assumption that was made here which breaks due to the different calculation order of SIMD optimizations instead of the scalar ones.

Other tests like [this](https://github.com/facebookresearch/faiss/blob/6457e941f5f4808303bb480679cb78df0df51934/tests/test_residual_quantizer.py#L849) and [this](https://github.com/facebookresearch/faiss/blob/6457e941f5f4808303bb480679cb78df0df51934/tests/test_fast_scan.py#L88) one contain basically that loosened assumption.

PR #5502, opened 2026-08-05
harden: sanitize subprocess call in cmp_with_scann.py...

anupamme:
## Summary
Harden input handling in `benchs/bench_all_ivf/cmp_with_scann.py` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | python.lang.security.audit.dangerous-system-call-tainted-env-args.dangerous-system-call-tainted-env-args |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `python.lang.security.audit.dangerous-system-call-tainted-env-args.dangerous-system-call-tainted-env-args` |
| **File** | `benchs/bench_all_ivf/cmp_with_scann.py:94` |
| **Assessment** | Defensive hardening |

**Description**: Found user-controlled data used in a system call. This could allow a malicious actor to execute commands. Use the 'subprocess' module instead, which is easier to use without accidentally exposing a command injection vulnerability.

## Threat Model Context

This is a Python library - vulnerabilities affect applications that import this code.

## Changes
- `benchs/bench_all_ivf/cmp_with_scann.py`

> **Note**: The following lines in the same file use a similar pattern and may also need review: `benchs/bench_all_ivf/cmp_with_scann.py:79`, `benchs/bench_all_ivf/cmp_with_scann.py:196`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---

### DIFF
--- a/benchs/bench_all_ivf/cmp_with_scann.py
+++ b/benchs/bench_all_ivf/cmp_with_scann.py
@@ -6,6 +6,8 @@
 import time
 import sys
 import os
+import shlex
+import shutil
 import argparse
 
 import numpy as np
@@ -91,7 +93,8 @@ def main():
         ds = load_dataset(args.db, download=args.download)
         print(ds)
         # store for SCANN
-        os.system(f"rm -rf {cache_dir}; mkdir -p {cache_dir}")
+        shutil.rmtree(cache_dir, ignore_errors=True)
+        os.makedirs(cache_dir, exist_ok=True)
         tosave = {
             "xb": ds.get_database(),
             "xq": ds.get_queries(),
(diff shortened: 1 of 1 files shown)


mnorris11:
The script is just benchmark code, and a user can just run these commands anyway without going through the benchmark.

PR #5508, opened 2026-08-09
Fix IDSelector handling in unbounded HNSW search

### LINKED ISSUE #5504: IndexHNSW ignores SearchParameters.sel when bounded_queue=False
## Summary

When searching an `IndexHNSW` index with `SearchParameters.sel`, the ID selector works correctly when `bounded_queue=True`.

However, when `bounded_queue=False`, the selector is ignored and the search may return IDs that are explicitly excluded.

This violates the contract of `SearchParameters.sel`: when it is non-null, only IDs accepted by the selector should be returned.

## Bug type

Incorrect result

## Environment

```text
FAISS version: 1.14.2
Python: 3.x
OS: Windows
Installation: pip faiss-cpu
Index: IndexHNSWFlat
Metric: METRIC_L2
```

## Minimal reproducer

```python
import faiss
import numpy as np

xb = np.array(
    [
        [0.0, 0.0],
        [10.0, 0.0],
        [20.0, 0.0],
        [30.0, 0.0],
    ],
    dtype=np.float32,
)

xq = np.array([[0.0, 0.0]], dtype=np.float32)

index = faiss.IndexHNSWFlat(2, 4, faiss.METRIC_L2)
index.hnsw.efConstruction = 40
index.add(xb)

# Only ID 1 is allowed.
selector = faiss.IDSelectorRange(1, 2)

params = faiss.SearchParametersHNSW(
    efSearch=40,
    bounded_queue=False,
    sel=selector,
)

D, I = index.search(xq, 1, params=params)

print("labels:", I.tolist())
print("distances:", D.tolist())

assert I.tolist() == [[1]]
assert D.tolist() == [[100.0]]
```

## Expected behavior

Because the selector only allows ID 1, the result should be:

```text
labels: [[1]]
distances: [[100.0]]
```

Every returned label must satisfy:

```text
sel->is_member(label)
```

This requirement should not depend on the value of `bounded_queue`.

## Actual behavior

With `bounded_queue=False`, the result is:

```text
labels: [[0]]
distances: [[0.0]]
```

ID 0 is not accepted by the selector, but it is still returned.

## Comparison

Using the same index, data, query, and selector, but changing the parameter to:

```python
bounded_queue=True
```

returns the correct result:

```text
labels: [[1]]
distances: [[100.0]]
```

## Why this is a bug

This is not normal approximate-nearest-neighbor recall behavior.

Even if HNSW fails to find the best allowed candidate, it must not return an ID that is explicitly rejected by the selector. The selector is a hard validity constraint on returned results, not a ranking preference.

In this example:

* ID 0 is explicitly excluded;
* ID 1 is the only allowed ID;
* filtering works when `bounded_queue=True`;
* filtering is ignored when `bounded_queue=False`.

This indicates a queue-path-specific ID filtering bug.

## Impact

Applications that rely on `SearchParameters.sel` for permission filtering, tenant isolation, metadata filtering, or deleted-record filtering may receive IDs that should not be returned when `bounded_queue=False`.

This can cause filtering rules to be bypassed and restricted or cross-tenant data to appear in search results.

YingzuoLiu:
## Summary

- apply `SearchParametersHNSW.sel` to the unbounded HNSW search path
- keep filtered nodes available for graph traversal while excluding them from the result set
- add regression coverage for L2 and inner-product search, restrictive `efSearch`, and an empty selector

## Root cause

The bounded HNSW path applies the selector through its result handler. The unbounded path did not receive the selector and used the same heap for traversal pruning and final results, so IDs rejected by `SearchParameters.sel` could be returned.

## Implementation

When a selector is present, the unbounded path maintains a separate result heap containing only accepted IDs. The existing candidate heap remains unchanged and can still contain rejected nodes, preserving traversal through them. When no selector is present, the existing unfiltered result behavior is preserved.

## Validation

- issue reproducer: unbounded search now returns ID 1 at distance 100.0 instead of rejected ID 0
- `tests/test_search_params.py`: 52 passed
- HNSW C++ tests: 14 passed
- full CPU C++ test suite: 278 passed, 5 skipped
- clang-format and `git diff --check`: passed

Fixes #5504

### DIFF
--- a/faiss/impl/HNSW.cpp
+++ b/faiss/impl/HNSW.cpp
@@ -1708,6 +1708,8 @@ TopCandidatesQueue<C> search_from_candidate_unbounded_fixVT(
 
         auto add_to_heap = [&](const size_t idx, const float dis) {
             if constexpr (use_selector) {
+                // Check the size before top(): the starting node may be
+                // rejected, leaving result_candidates empty.
                 if (sel->is_member(idx) &&
                     (result_candidates.size() < static_cast<size_t>(ef) ||
                      C::cmp(result_candidates.top().first, dis))) {
--- a/faiss/impl/HNSW.cpp
+++ b/faiss/impl/HNSW.cpp
@@ -1644,17 +1644,29 @@ void reservePriorityQueue(
 /// Templated body of `search_from_candidate_unbounded`. The choice of
 /// max-heap vs min-heap for both `top_candidates` and `candidates` is
 /// derived from C via `TopCandidatesQueue` / `CandidatesQueue`.
-template <typename VTType, class C>
+template <typename VTType, class C, bool use_selector>
 TopCandidatesQueue<C> search_from_candidate_unbounded_fixVT(
         const HNSW& hnsw,
         const HNSW::Node& node,
         DistanceComputer& qdis,
         int ef,
         VTType& vt,
-        HNSWStats& stats) {
+        HNSWStats& stats,
+        const IDSelector* sel) {
     int ndis = 0;
     TopCandidatesQueue<C> top_candidates;
     reservePriorityQueue(top_candidates, ef);
+
+    TopCandidatesQueue<C> result_candidates;
+    if constexpr (use_selector) {
+        FAISS_ASSERT(sel);
+        // Keep rejected nodes in top_candidates so they can still be used
+        // for graph traversal, but never return them as search results.
+        reservePriorityQueue(result_candidates, ef);
+        if (sel->is_member(node.second)) {
+            result_candidates.push(node);
+        }
+    }
 
     CandidatesQueue<C> candidates;
     reservePriorityQueue(candidates, ef);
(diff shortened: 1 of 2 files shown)


faiss/impl/HNSW.cpp
```diff
@@ -1310,18 +1310,30 @@ void reservePriorityQueue(
 /// Templated body of `search_from_candidate_unbounded`. The choice of
 /// max-heap vs min-heap for both `top_candidates` and `candidates` is
 /// derived from C via `TopCandidatesQueue` / `CandidatesQueue`.
-template <typename VTType, class C>
+template <typename VTType, class C, bool use_selector>
```
mnorris11, line 1313: Instead of `use_selector` template, can we just check `sel`?

The pattern elsewhere in the file is `!sel || sel->is_member(v1)` we can follow that.

YingzuoLiu: Updated — I removed the use_selector template parameter and now use the nullable sel directly, following the existing pattern in HNSW.cpp. The traversal/result-heap separation is unchanged, and the regression and HNSW test suites still pass.

YingzuoLiu:
Hi @mnorris11, just a quick follow-up — I’ve addressed the review comment and rebased the PR onto the current `main`.
Please let me know if there’s anything else needed from my side. Thanks!

mnorris11:
Hi @YingzuoLiu , I am merging this now, thanks for the contribution. I did some minor refactor on top of it on the internal import so the committed code will be slightly different than the PR, but the functionality is the same.

PR #5510, opened 2026-08-11
Detect RISC-V RVV support from target flags

KimmyXYC:
## Summary

- Detect RVV and Zvfhmin support using the active compiler target flags.
- Enable the RISC-V vector kernels only when the compiler probe succeeds.
- Fall back to scalar implementations for lower RISC-V target baselines.
- Remove per-source `-march=rv64gcv_zvfhmin` and `-mabi=lp64d` overrides.

## Motivation

Faiss currently selects RVV kernels for every RISC-V target and forces a specific architecture through per-source compiler options. This can make binaries incompatible with lower target baselines and bypass distribution compiler flags.

Automatic detection keeps lower baselines portable while still enabling the RVV implementations when the selected target provides both RVV and Zvfhmin.

## Testing

Tested through openRuyi OBS builds:

- RVA20/RISC-V: the probe failed as expected, scalar implementations were used, and the build and checks succeeded.
- RVA23/RISC-V: the probe succeeded, RVV sources were compiled, and the build and checks succeeded.
- AMD64: the RISC-V probe was skipped and the build and checks succeeded.

### DIFF
--- a/faiss/CMakeLists.txt
+++ b/faiss/CMakeLists.txt
@@ -74,6 +74,32 @@ set(FAISS_SIMD_RVV_SRC
   utils/hamming_distance/hamming_rvv.cpp
   utils/simd_impl/rabitq_rvv.cpp
 )
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "(riscv64|riscv)")
+  include(CheckCXXSourceCompiles)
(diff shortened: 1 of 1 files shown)


mnorris11:
Hi @KimmyXYC  can you check the compilation error?

PR #5518, opened 2026-08-13
Add missing headers

rupprecht:
When using libc++ after https://github.com/llvm/llvm-project/commit/ada3786e91ca2058f3ac8255a024c12ae7d263ee, `#include <random>` does not imply `#include <cmath>`, so we need to directly import it to avoid breakages, e.g.

```
faiss/utils/random.cpp:140:40: error: use of undeclared identifier 'log'; did you mean 'long'?
  140 |                 x[i] = a * sqrt(-2.0 * log(s) / s);
```

While `#include <cmath>` is the only one necessary to fix a breakage from that LLVM commit, there are a few other C++ headers not being directly included, so I included those to prevent future breakages from refactorings.

### DIFF
--- a/faiss/utils/random.cpp
+++ b/faiss/utils/random.cpp
@@ -9,6 +9,12 @@
 
 #include <faiss/utils/random.h>
 
+#include <cmath>
+#include <cstddef>
+#include <limits>
+#include <utility>
+#include <vector>
+
 extern "C" {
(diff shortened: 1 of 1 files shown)


alibeklfc:
Hi, thanks for the contribution. Can you please sign Contributor License Agreement?

PR #5525, opened 2026-08-17
SVS: allow option to disable vector storing

ethanglaser:
Allow avoiding storing original vectors, which may not necessarily be used, in which case they would be wasting memory


ethanglaser:
@mnorris11 FYI

mnorris11:
@ethanglaser  does this require bumping an SVS version?

ethanglaser:
> @ethanglaser does this require bumping an SVS version?

No SVS change, it would just enhance users by allowing avoiding memory waste in Milvus (ie https://github.com/mihaic/knowhere/commit/2163017e2a82c63f2ab1c96cb90ba17945584406) and provide a similar option for faiss users

PR #5526, opened 2026-08-18
Add staged RaBitQ storage to HNSW

lyang24:


faiss/impl/HNSW.h
```diff
@@ -305,18 +311,24 @@ struct HNSWStats {
             0; /// number of queries for which the candidate list is exhausted
     size_t ndis = 0;  /// number of distances computed
     size_t nhops = 0; /// number of hops aka number of edges traversed
+    size_t n_rabitq_1bit = 0;
```
mnorris11: We should keep the rabitq-specific things out of the general HNSWStats. I see there is RaBitQHNSWStats below anyway?

faiss/impl/HNSW.h
```diff
@@ -157,6 +157,12 @@ struct HNSW {
     /// use Panorama progressive pruning in search
     bool is_panorama = false;
 
+    /// use staged RaBitQ search: a 1-bit estimate for every neighbor, and the
+    /// full multi-bit distance only where the error bound leaves the candidate
+    /// able to enter the result heap. Requires an IndexRaBitQ storage with
+    /// nb_bits >= 2 (1-bit codes carry no f_error, so there is no bound).
+    bool is_rabitq = false;
```
mnorris11: These two are sort of "search methods". We can make this an enum and switch on it instead.

faiss/impl/HNSW.cpp
```diff
@@ -1290,6 +1292,171 @@ int search_from_candidates_dispatch(
...
+ * difference is attributable to the staged distance rather than to a different
+ * traversal.
+ *
+ * Both the beam and the result heap receive the same value per candidate, which
+ * means the beam mixes 1-bit estimates with refined distances. That is the
+ * reference implementation's behavior and it is why recall is not guaranteed to
+ * match vanilla HNSW.
+ */
+template <typename VTType, class C>
+int search_from_candidates_rabitq_fixVT(
```
mnorris11: This function seems to have duplication with `search_from_candidates_fixVT`, can it be deduplicated?

faiss/IndexHNSW.cpp
```diff
@@ -1128,6 +1137,107 @@ IndexHNSWSQ::IndexHNSWSQ(
...
+}
+
+void IndexHNSWRaBitQ::add(idx_t n, const float* x) {
+    FAISS_THROW_IF_NOT_MSG(
+            build_storage,
+            "index is finalized; rebuild it from scratch to add more vectors");
+    FAISS_THROW_IF_NOT(is_trained);
+    size_t n0 = ntotal;
+    storage->add(n, x);
+    build_storage->add(n, x);
```
mnorris11: IIUC, this would mean the caller has the below in memory during add():
- fp32 copy of their database vectors
- build_storage, basically a copy of their database vectors (IIUC?)
- HNSW graph
- RaBitQ storage

Is this the right understanding?

faiss/IndexHNSW.h
```diff
@@ -216,6 +216,47 @@ struct IndexHNSWSQ : IndexHNSW {
...
+ */
+struct IndexHNSWRaBitQ : IndexHNSW {
+    /// exact vectors, needed only to build the graph; null once finalized
+    IndexFlat* build_storage = nullptr;
+
+    IndexHNSWRaBitQ();
+    IndexHNSWRaBitQ(
+            int d,
+            int M,
+            uint8_t nb_bits = 4,
```
mnorris11: Making the default 4 here is a sensible default,  but for serialization, omitting it will diverge from existing behavior: 
- default: 1 bit
- this approach: 4 bits

We should keep in line with existing serialization

faiss/impl/HNSW.h
lyang24: yep thanks for catching

faiss/impl/HNSW.cpp
lyang24: Fixed. RaBitQ and the default path now share the same search_from_candidates_fixVT traversal, with only the distance evaluator differing

faiss/IndexHNSW.cpp
lyang24: Yep, that’s right for a single add() containing the full dataset. The only nuance is that with batched adds, the caller only holds the current batch, while build_storage still accumulates all vectors. It’s temporary—finalize() drops it, leaving the graph + RaBitQ codes.

faiss/IndexHNSW.h
lyang24: right set default to 1 bit

faiss/IndexHNSW.cpp
mnorris11: Unfortunately this build_storage copy will be a problem since it doubles memory during add(), and also no other index has the finalize() requirement so the API diverges. The main issue seems to be RaBitQ not having `symmetric_dis()`. We can just implement that for 1 bit. It threw because it was not needed before.

I tried it quickly locally to compare build_storage approach with symmetric_dis for 1 bit, and symmetric_dis multibit and swept efSearch. 1 bit is only 1-2% worse over all efSearch than build_storage and multibit, and multibit is way slower. So I think we can drop build_storage and just do a 1-bit symmetric_dis. 
```
  recall@10 (mean of 2 builds, ± half-spread)

  ┌──────────┬────────────────┬────────────────┬────────────────┬──────────┬─────────┐
  │ efSearch │ build_storage  │  sym_multibit  │    sym_1bit    │ multi−bs │ 1bit−bs │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 10       │ 0.7791 ±0.0008 │ 0.7661 ±0.0112 │ 0.7555 ±0.0046 │ −0.0130  │ −0.0236 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 16       │ 0.8276 ±0.0013 │ 0.8154 ±0.0066 │ 0.8080 ±0.0068 │ −0.0122  │ −0.0196 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 24       │ 0.8531 ±0.0017 │ 0.8516 ±0.0048 │ 0.8362 ±0.0035 │ −0.0016  │ −0.0170 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 32       │ 0.8717 ±0.0037 │ 0.8666 ±0.0030 │ 0.8589 ±0.0029 │ −0.0051  │ −0.0128 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 48       │ 0.8910 ±0.0015 │ 0.8873 ±0.0036 │ 0.8802 ±0.0010 │ −0.0036  │ −0.0107 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 64       │ 0.8972 ±0.0011 │ 0.8970 ±0.0017 │ 0.8904 ±0.0014 │ −0.0002  │ −0.0068 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 128      │ 0.9084 ±0.0023 │ 0.9119 ±0.0034 │ 0.9079 ±0.0009 │ +0.0036  │ −0.0004 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 256      │ 0.9178 ±0.0010 │ 0.9185 ±0.0003 │ 0.9160 ±0.0010 │ +0.0007  │ −0.0018 │
  ├──────────┼────────────────┼────────────────┼────────────────┼──────────┼─────────┤
  │ 512      │ 0.9199 ±0.0000 │ 0.9199 ±0.0007 │ 0.9180 ±0.0006 │ +0.0000  │ −0.0018 │
  └──────────┴────────────────┴────────────────┴────────────────┴──────────┴─────────┘
```

AI description of the proposed change, Claude can probably figure it out for you from some prompt similar to this:

  ▎ Building the graph from the codes instead of an fp32 copy
  ▎
  ▎ build_storage keeps a full fp32 copy alive for the whole build. At d=128, nb_bits=4 a RaBitQ code is 84 bytes against 512 for fp32, so the temporary copy
  ▎ is ~6x the size of the index it's there to build — 2.56 GB vs 420 MB at 5M vectors. It exists only because RaBitQDistanceComputer::symmetric_dis() 
  ▎ threw, and HNSW needs code-to-code distances for neighbor-diversity pruning.
  ▎
  ▎ Implementing symmetric_dis removes the need for it. The estimator uses the exact stored norms with a reconstructed cross term:
  ▎
  ▎ ||r_i - r_j||^2  ~=  ||r_i||^2 + ||r_j||^2 - 2 <r_i^, r_j^>,   r = o - centroid
  ▎
  ▎ or_minus_c_l2sqr supplies both norms exactly, so only the cross term is approximated, and the centroid cancels. Two variants:
  ▎
  ▎ - 1-bit: r^ = alpha * sign(r), so the cross term is alpha_i * alpha_j * (d - 2*popcount(c_i XOR c_j)) — one popcount over the sign bytes, no 
  ▎ reconstruction, no scratch memory.
  ▎ - multi-bit: reconstruct both residuals from sign + ex bits (r^ = -0.5 * f_rescale_ex * w, the same weight w the full-distance kernel uses) and take an 
  ▎ inner product. O(d) with two float[d] scratch buffers.
  ▎
  ▎ One derivation trap worth calling out. The obvious magnitude is dp_multiplier / sqrt(d), but that is the IP-unbiased reconstruction — the same one that 
  ▎ makes decode() unsuitable for L2. It overshoots by ~pi/2 for Gaussian residuals, which inflates the cross term by ~2.5x and clamps near-duplicate pairs 
  ▎ to zero 100% of the time. The L2-optimal magnitude is ||r||_1 / d, recoverable from the same two stored floats as or_minus_c_l2sqr / (sqrt(d) * 
  ▎ dp_multiplier). With that, both variants are unbiased (multi-bit: 0.004% bias, 2.0% rms, 96% pair-order agreement; 1-bit: 1.0% bias, 10% rms, 81% 
  ▎ agreement) and neither clamps.
  ▎
  ▎ API surface
  ▎
  ▎ On RaBitQDistanceComputer:
  ▎ - symmetric_dis(i, j) — replaces the throwing stub. L2 only; IP still throws, because the stored factors fold in ||o||^2 and <o-c, c> isn't recoverable 
  ▎ from them.
  ▎ - bool symmetric_1bit_only — selects the cheap cross term.
  ▎ - std::vector<float> symmetric_buf — per-computer scratch for the multi-bit path. Safe because each build thread constructs its own distance computer.
  ▎
  ▎ On IndexHNSWRaBitQ:
  ▎ - new ctor parameter and field bool exact_build = true; when false, build_storage is never allocated and add() builds straight on the codes.
  ▎ - the train() / add() guards now distinguish "never had a build storage" from "finalized", instead of treating a null pointer as finalized.
  ▎
  ▎ No changes to HNSW.cpp or HNSW.h. shrink_neighbor_list and add_link already call qdis.symmetric_dis(); they just stop throwing. That also fixes an 
  ▎ existing failure mode: building an HNSW over IndexRaBitQ storage today aborts the process rather than raising, because the throw escapes an OpenMP 
  ▎ region.

lyang24: Good point — I reworked this to build directly from the RaBitQ codes (fyi quite a big change). symmetric_dis() now uses the 1-bit sign plane for L2 pruning, including with multi-bit storage, so build_storage and finalize() are gone. This brings add() and serialization back in line with the other HNSW variants and removes the temporary FP32 copy. I limited IndexHNSWRaBitQ to L2 for now since the stored factors don’t support a reliable symmetric IP estimate. Thanks for the benchmark and derivation

PR #5527, opened 2026-08-18
  Speed up HNSW RaBitQ search with code prefetching

lyang24:
Prefetch RaBitQ sign bits and error factors four neighbors ahead during
  level-0 HNSW traversal.

  RaBitQ codes are accessed through irregular graph edges. Prefetching the bytes
  needed by the 1-bit filter hides part of this memory latency without loading the
  extra multi-bit code that most candidates never refine into.

  This does not change search results or distance calculations.

  ## Benchmark

  Cohere 1M, d=1024, IP, HNSW32,RaBitQ4, k=10, single thread.

  AWS Graviton4 (m8g.4xlarge), warm runs interleaved across control and patched
  builds for five rounds:
     efSearch      Before       After    Improvement
   16    1905 QPS    2093 QPS          +9.8%
   64    1224 QPS    1513 QPS         +23.6%
   256     564 QPS     794 QPS         +40.7%

  Lookahead 4 also outperformed 8 and 16 on the same workload.

  Complete result ID and distance arrays were bit-identical between builds.
  The implementation uses Faiss's portable prefetch_L1 helper and was validated
  on both AArch64 and x86-64.

### DIFF
--- a/faiss/IndexHNSW.cpp
+++ b/faiss/IndexHNSW.cpp
@@ -7,6 +7,8 @@
 
 #include <faiss/IndexHNSW.h>
 
+#include <faiss/IndexRaBitQ.h>
+
 #include <omp.h>
 #include <atomic>
 #include <cinttypes>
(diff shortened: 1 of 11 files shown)


lyang24:
prefetching is case by case

PR #5531, opened 2026-08-20
Add AVX512 VPOPCNT dynamic dispatch level

lyang24:
## What

  Split the existing VPOPCNTDQ RaBitQ and Hamming kernels from the full Sapphire Rapids SIMD level.

  DD builds now expose an `AVX512_VPOPCNT` capability for CPUs such as Ice Lake and Zen 4. The fallback chain is:

  `AVX512_SPR -> AVX512_VPOPCNT -> AVX512 -> AVX2 -> NONE`

  The ordinary static AVX512 target remains unchanged and contains no VPOPCNT instructions.

  ## Why

  The VPOPCNT kernels only require baseline AVX-512 plus `AVX512_VPOPCNTDQ`. Tying them to the SPR level unnecessarily excluded CPUs that support VPOPCNTDQ but not AVX512-FP16, notably Zen 4.

  ## Validation

  Tested on:

  - AMD Zen 4: automatically selects `AVX512_VPOPCNT`, not `AVX512_SPR`
  - Intel Sapphire Rapids: continues to select `AVX512_SPR`
  - DD, static AVX512, and static AVX512_SPR builds
  - RaBitQ L2/IP across multiple dimensions, code widths, query widths, and SIMD tails
  - Hamming VPOPCNT paths

  Cohere 1M IVFRaBitQ4, `nprobe=64`, Zen 4:

  - 1 thread: approximately 8% higher QPS
  - 8 threads: 29% higher QPS

  Recall, result hashes, and refinement counts were identical

### DIFF
--- a/benchs/bench_rabitq.py
+++ b/benchs/bench_rabitq.py
@@ -23,8 +23,8 @@ from faiss.contrib.datasets import SyntheticDataset
 #   d=512  -> size=64  -> 512-bit zmm, 1 iteration per bit-plane
 #   d=768  -> size=96  -> 512-bit zmm + 256-bit tail
 #   d=1024 -> size=128 -> 512-bit zmm only, 2 iterations per bit-plane
-# Sweeping these is useful for verifying the AVX512_SPR (vpopcntdq)
-# specialization in faiss/utils/simd_impl/rabitq_avx512_spr.cpp and for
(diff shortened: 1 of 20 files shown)


mnorris11:
I am renaming some SIMD functions to be easier to understand, so I will take a look at this one soon

PR #5534, opened 2026-08-21
Fix HNSW concurrent race segfault on RISC-V

anthony-zy:
On RISC-V, GCC -O2 optimizes the `while (i < end) neighbors[i++] = -1` loop in add_link_tpl into a memset(0xFF) call, which writes byte-by-byte (`sb`). The reader in search_neighbors_to_add_fixVT does a lock-free `lw` on the same neighbor slot while the writer holds the other_id lock but the reader does not. A torn read yields 0x00FFFFFF (a half-finished write), which passes the `nodeId < 0` check and is used as a valid id, eventually driving an out-of-bounds vector access and segfault.

Fix the writer to store -1 with a relaxed atomic store so the reader sees either the old valid id or a complete -1, and add a bounds check on the reader as a defensive backstop. Both guarded by __riscv so other architectures are untouched.

### DIFF
--- a/faiss/impl/HNSW.cpp
+++ b/faiss/impl/HNSW.cpp
@@ -419,7 +419,13 @@ void add_link_tpl(
     }
     // they may have shrunk more than just by 1 element
     while (i < end) {
+#ifdef __riscv
+        __atomic_store_n(
+                &hnsw.neighbors[i], (storage_idx_t)-1, __ATOMIC_RELAXED);
+        i++;
+#else
         hnsw.neighbors[i++] = -1;
+#endif
     }
 }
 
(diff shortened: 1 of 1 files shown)


mnorris11:
@anthony-zy do you also see this with deterministic HNSW on? It should not go down this code path. See https://github.com/facebookresearch/faiss/pull/5486 for how to turn it on. We will be deleting the existing lock-based non-deterministic path by the next release.

anthony-zy:
@mnorris11 Thanks. Confirmed with hnsw_deterministic_build = true: no crash, so the
deterministic path is clean.

But the lock-based path is still the default (flag defaults to false), so
users who don't opt in still hit the segfault today. My patch is a
stopgap for that default path until it's removed. Want me to keep it, or
flip the default to deterministic now?

mnorris11:
Great to hear it is resolved by deterministic flag. 

We're aiming to delete the lock-based path in a few weeks and make deterministic path the only path, so we can hold off on the patch.

PR #5535, opened 2026-08-21
feat: implement RVV optimized scalar quantizer codecs and distance computation

ihb2032:
## Summary

This PR adds RISC-V Vector Extension SIMD support for Faiss scalar quantizer.

The implementation introduces runtime vector-length aware RVV kernels using the m8 vector configuration, instead of assuming a fixed vector width. This matches the variable-length nature of RVV hardware and allows the implementation to scale across different RISC-V vector implementations.

## Changes

- Add RVV implementations for scalar quantizer codecs:
  - 8-bit codec decoding
  - 4-bit codec decoding
  - (6-bit stays scalar: the RVV gather decoder measured 12-22% slower
    than the scalar path, so QT_6bit falls back until a faster decoder
    exists)

- Add RVV quantizer reconstruction support for:
  - Uniform quantizers
  - Non-uniform quantizers
  - FP16 (Zvfhmin is the enforced minimum ISA of the RISCV_RVV level; a
    build without it fails compilation rather than silently falling back)
  - BF16
  - Direct 8-bit quantization
  - Signed 8-bit quantization
  - LloydMax quantizers (1/2/3/4/8-bit)

- Add RVV optimized similarity implementations:
  - L2 distance
  - Inner product


ihb2032:
I believe this CI failure is caused by the RISC-V GCC version.
The same code builds successfully with g++-14 and g++-15.1 locally. The CI uses riscv64-linux-gnu-g++-13, which seems to lack support for the RVV FP16 intrinsics (vfloat16m4_t, __riscv_vfwcvt_f_f_v_f32m8) used here.
Updating the CI RISC-V toolchain to GCC 14+ should resolve this issue.

mnorris11:
@ihb2032 feel free to make the change in this PR to the gcc version for the riscv job

ihb2032:
@mnorris11 Updated the RISC-V toolchain version and confirmed the CI build is successful.

ihb2032:
Hi @mnorris11, here are the benchmark results for this PR, measured on a **SpacemiT K3** (RISC-V, RVV) board. Two regressions need your call before merge.

## Setup

- Benchmark: `perf_tests/bench_scalar_quantizer_distance` (+ encode/decode/accuracy), d=128, n=2000, 20 iterations, single-threaded
- Baseline: `main` (this branch's merge-base), Optimized: this PR
- All distance numbers in ns per benchmark iteration (cpu_time)

## sq-distance results (K3)

| benchmark | baseline | this PR | speedup |
|---|---:|---:|---:|
| QT_fp16 IP | 1,871,406,610 | 275,812,986 | **6.79x** |
| QT_fp16 L2 | 1,959,386,090 | 303,159,022 | **6.46x** |
| QT_bf16 IP | 1,381,156,579 | 317,210,834 | **4.35x** |
| QT_bf16 L2 | 1,468,961,832 | 342,725,949 | **4.29x** |
| QT_8bit_direct IP | 1,385,076,715 | 315,899,828 | **4.38x** |
| QT_8bit_direct L2 | 1,470,483,012 | 342,645,008 | **4.29x** |
| QT_8bit_direct_signed IP | 1,474,115,501 | 344,632,385 | **4.28x** |
| QT_8bit_direct_signed L2 | 1,561,504,202 | 375,516,567 | **4.16x** |
| QT_8bit_uniform L2 | 2,827,837,050 | 1,461,880,446 | **1.93x** |
| QT_8bit_uniform IP | 2,710,494,295 | 1,431,843,834 | **1.89x** |
| QT_8bit L2 | 2,700,351,140 | 1,601,147,322 | **1.69x** |
| QT_8bit IP | 2,585,627,344 | 1,532,610,677 | **1.69x** |
| QT_4bit L2 | 2,771,739,049 | 1,638,479,753 | **1.69x** |
| QT_4bit IP | 2,421,483,649 | 1,578,416,337 | **1.53x** |
| QT_4bit_uniform IP | 2,425,139,614 | 1,537,596,652 | **1.58x** |
| **QT_4bit_uniform L2** | **347,312,659** | **1,563,596,092** | **0.22x (regression)** |
| **QT_6bit L2** | **2,955,453,643** | **3,362,615,662** | **0.88x (regression)** |
| **QT_6bit IP** | **2,585,466,583** | **3,315,402,733** | **0.78x (regression)** |

sq-encode / sq-decode / sq-accuracy are flat (within ±2%, reconstruction errors identical), as expected — encode/decode intentionally stay scalar (the RVV quantizers inherit `encode_vector`/`decode_vector` from the `NONE` base).

## Regression 1: QT_4bit_uniform L2 (0.22x)

Root cause: `main` has a dedicated full specialization `DCTemplate<QuantizerTemplate<Codec4bit<RISCV_RVV>, UNIFORM, RISCV_RVV>, SimilarityL2<RISCV_RVV>, RISCV_RVV>` that works entirely in the integer domain: the query is pre-quantized to nibbles at `set_query` time (`q_lo`/`q_hi`), then each code is scored with u8 sub / u16 widening square / u32 widening reduce — no float decode at all.

This PR replaces it with the generic float path (`reconstruct_m8_components` + `vfmacc`), hence the slowdown.

Note the semantic difference: the old kernel is **approximate** (the query itself is quantized to 4 bits before distance computation); the new generic path is **exact** (identical semantics to the scalar `NONE` path).

**Options:**
- **(a)** Restore the old specialization alongside the new generic code (the full specialization wins over the `requires`-constrained partial one, so they coexist cleanly) — recovers the ~7x advantage over scalar for this case, but keeps the approximate quantized-query behavior.
- **(b)** Keep this PR's exact semantics and accept the regression (still 4.5x slower than the old kernel, though comparable to the other exact kernels).

## Regression 2: QT_6bit L2/IP (0.88x / 0.78x)

Root cause: `Codec6bit<RISCV_RVV>::decode_m8_components` issues 3x `vluxei32` (indexed/gather loads) per iteration to fetch the 3 packed bytes per 4 components. Gather loads are micro-coded and slow on the K3; together with ~20 vector ops (4 mask compares + 3 merges) the vector decode loses to the scalar per-component loop. perf confirms it: the 6-bit `DCTemplate::query_to_code` tops the sq-distance profile at ~31% of samples.

**Options:**
- **(a)** Rework the decode to use `vlseg3e8` segment loads (contiguous memory) instead of gathers. Lane order becomes field-grouped, so the query vector must be permuted the same way at `begin_m8`/`set_query` time — legal because the distance accumulation is order-agnostic. Bigger change.
- **(b)** Drop `decode_m8_components` from `Codec6bit<RISCV_RVV>` and add a `requires` constraint on `QuantizerTemplate<..., RISCV_RVV>::reconstruct_m8_components` so `has_reconstruct_m8_v` becomes false for 6-bit and it falls back to the scalar `NONE` path — i.e. baseline performance, no regression.

juancarpio27:
Thanks for extending the RVV scalar-quantizer support. The reconstruction-based approach avoids the query re-quantization and integer-overflow issues identified on https://github.com/facebookresearch/faiss/pull/5539. However, I found three issues to address before landing:

1. The tail-handling issue from https://github.com/facebookresearch/faiss/pull/5539 is repeated here. The final short iteration uses tail-agnostic vfmacc, so inactive accumulator lanes may become undefined before the original vector length is reduced. Please use the _tu variants, as in faiss/utils/simd_impl/distances_rvv.cpp, or reduce each chunk separately.

2. The packed decoders use 8-bit lane IDs as gather indices. On valid RVV implementations with VLEN greater than 1024 bits, e32m8 can contain more than 256 lanes, causing those indices to wrap and decode incorrect bytes. This affects the 4-bit codec and the Lloyd-Max 1/2/4-bit implementations. Please use wider indices, limit the chunk size, or explicitly enforce a supported VLEN limit.

3. The QEMU job only runs SIMD configuration tests, so none of these new scalar-quantizer kernels are exercised. The existing scalar-quantizer parity helper also omits RISCV_RVV. Please add and run RVV-versus-scalar tests covering the newly supported quantizers, both L2 and inner product, the distance APIs, and dimensions around VLMAX boundaries.

juancarpio27:
Thanks for the comprehensive update. The previous tail-handling issue is fixed, the packed gathers now use wide indices, and the new RVV-versus-scalar coverage is a significant improvement.

I found one remaining correctness issue before approval. The packed 4-bit and Lloyd-Max 1/2/4-bit decoders assume every vector chunk starts on a byte boundary. RVV permits vsetvl to return a non-maximal, non-pack-aligned VL when AVL is between VLMAX and 2×VLMAX. A following chunk can therefore start within a byte while the decoder treats it as starting at bit/nibble zero.

Please either calculate packed positions from the absolute component index (i + lane) or force intermediate chunks to an exact, pack-aligned VLMAX. QEMU’s usual VL-selection policy masks this case.

I would also add a d == 0 guard and clarify how the greater-than-256-lane index fix was validated, since VLEN 1024 reaches exactly 256 lanes. After those changes and a final passing CI run, this should be close.

ihb2032:
Thanks for the careful review @juancarpio27. The remaining items are addressed in the latest commit:

1. **Mid-byte chunk starts**: the packed 4-bit and Lloyd-Max 1/2/4-bit decoders now derive byte offsets and nibble/bit positions from the absolute component index (`i + lane`) instead of the in-chunk lane index, matching what the 6-bit and Lloyd-Max 3-bit decoders already did. A non-maximal, non-pack-aligned `vl` (AVL between VLMAX and 2×VLMAX) now decodes correctly. As a side benefit the loaded byte range is computed exactly, so the final chunk no longer touches bytes past the last used one. I validated the index arithmetic with a scalar re-implementation of the new addressing exercised over every chunk split — both QEMU's `min(AVL, VLMAX)` policy and a hostile policy returning odd, non-pack-aligned `vl` in the (VLMAX, 2×VLMAX) window — across d = 1..513; results match the scalar decoders bit-exactly. (QEMU itself cannot surface this case, as you noted.)

2. **d == 0**: both RVV `DCTemplate` distance loops now return 0.0f early for empty vectors instead of relying on `vsetvl(0)`/zero-length reduction semantics.

3. **>256-lane index fix validation**: to clarify — this fix is by construction rather than by test. The gathers now use widened 32-bit lane indices, so no wrap is possible at any VLEN. You are right that VLEN=1024 reaches exactly 256 lanes for e32m8 (indices 0..255, no wrap), and QEMU caps at VLEN=1024, so the wrap case (VLEN > 1024) cannot be exercised in CI; the pre-fix code would only have failed on VLEN ≥ 2048 hardware. The parity test dimensions now also straddle the VLMAX=512 chunk boundary of VLEN=2048 implementations (511/512/513) so such hardware gets boundary coverage when available.

CI is re-running on the rebased branch.

juancarpio27:
Thanks for the detailed update. I re-reviewed the latest revision, and the previously reported issues are addressed in the implementation:

- The packed 4-bit and Lloyd-Max 1/2/4-bit decoders now derive byte and bit positions from the absolute component index, so non-pack-aligned chunk starts are handled correctly.
- Gather indices are widened to 32 bits, avoiding the greater-than-255-lane wraparound.
- The distance accumulators use tail-undisturbed operations before the full-width reduction.
- Both distance paths now return early for `d == 0`.
- The RVV-versus-scalar test is registered and covers the supported quantizers, both metrics, the distance APIs, and several VLMAX boundaries.

I found one remaining blocking portability/CI issue:

### `QT_fp16` requires `Zvfhmin`, but dispatch and QEMU enable only generic RVV

The RVV translation unit is compiled with `-march=rv64gcv_zvfhmin`, so the `QT_fp16` specialization emits `Zvfhmin` instructions. Runtime dispatch, however, marks `RISCV_RVV` available based only on generic RVV support; it does not check for `Zvfhmin`. A valid `rv64gcv` CPU without that optional extension can therefore select the FP16 RVV kernel and trap with an illegal instruction.

The new QEMU jobs have the same mismatch. Both use `-cpu rv64,v=true`, while QEMU 8.2's `x-zvfhmin` CPU property defaults to false. QEMU implementing the extension does not mean that it is enabled for the selected CPU, so the `QT_fp16` parity case is not configured to run safely.

Please either keep `QT_fp16` on the scalar path unless runtime `Zvfhmin` detection is available, or make `Zvfhmin` an explicit minimum ISA requirement and align the runtime contract and tests with it. For QEMU 8.2, the test CPU configuration should explicitly enable the corresponding property (for example, `x-zvfhmin=true`) in both parity runs and verify that the FP16 case executes.

Two smaller follow-ups:

- The new `d == 0` guards are not covered because the parity dimension list starts at 1. Please add a direct zero-dimensional regression for `query_to_code`, batch-of-four, and `symmetric_dis` if zero-dimensional quantizers are supported.
- The broad RVV matrix includes the EDEN Lloyd-Max types but not the `QT_*_tqmse` aliases. Those currently share the same implementation, but the existing TQMSE parity coverage is only `d = 32`, L2, and does not cover every distance API. Adding the TQMSE types to the new matrix would make that dispatch coverage explicit.
- `sq-rvv.cpp` uses `std::min` but does not include `<algorithm>` directly; please add the include instead of relying on a transitive header.

ihb2032:
Thanks for the detailed review @juancarpio27. The blocking item and the three follow-ups are addressed in the latest commits:

**Blocking: QT_fp16 / Zvfhmin** (2b278ef9) — took option (b), Zvfhmin is now the explicit minimum ISA of the RISCV_RVV level:
- The rv64gcv + Zvfhmin contract is documented on the `SIMDLevel` enum, at the availability detection in `simd_levels.cpp`, and on the CMake baseline. Since the RVV translation units are compiled with `-march=rv64gcv_zvfhmin`, the build itself requires the extension, so a runtime dispatch check could not add safety — a CPU without Zvfhmin cannot run these binaries at all.
- Both QEMU parity runs now pass `x-zvfhmin=true` explicitly; as you noted, QEMU 8.2 implements the extension but defaults the property to false on the rv64 CPU.
- QT_fp16 is split into its own `ScalarQuantizer.RVVFP16DistancePathParity` test so the FP16 case is individually visible in the CI log — it would die with SIGILL if Zvfhmin were not enabled, which makes its execution directly verifiable.

**Follow-ups** (138f53a7, 3301bb4c):
- **d == 0**: new `ScalarQuantizer.RVVZeroDimDistancePathParity` constructs zero-dimensional distance computers for every RVV-supported quantizer (L2 and IP) and checks `query_to_code`, `query_to_codes_batch_4` and `symmetric_dis` all return exactly 0, matching the scalar path. Zero-dimensional quantizers are constructible: the DC constructors only store the trained values/pointers, and the early returns fire before any query/code access.
- **TQMSE**: the `QT_*_tqmse` aliases are added to the RVV parity matrix (both metrics, all VLMAX-boundary dims, every distance API), making that dispatch coverage explicit.
- **Include**: `sq-rvv.cpp` now includes `<algorithm>` directly for `std::min`.

CI is re-running.

juancarpio27:
Thanks — the earlier correctness issues are addressed. Two items remain before approval:

  - `QT_6bit` still takes the slower RVV gather path despite the measured 12–22% regression. Please fall back to the scalar implementation until the decoder is faster.
  - The Zvfhmin parity test does not prove that the RVV FP16 kernel executed. If the compiler macro is absent, `reconstruct_m8_components` disappears and dispatch silently falls back to scalar. Since Zvfhmin is now the minimum ISA, please enforce it at compile time and make the FP16 RVV implementation unconditional.

ihb2032:
Both items are addressed in the latest commits (4bf25c73, 15b38f04):

1. **QT_6bit scalar fallback**: `Codec6bit<RISCV_RVV>` is now a scalar marker specialization (`: Codec6bit<NONE> {}`, as on main), and the `QuantizerTemplate` partial specializations for `RISCV_RVV` are constrained to codecs that provide `decode_m8_components`; codecs without one resolve to a scalar fallback partial specialization. `has_reconstruct_m8_v` is therefore false for 6-bit, so `DCTemplate` also selects the scalar path — no dispatch-site changes needed. QT_6bit is dropped from the RVV parity matrix since there is no RVV kernel left to parity-check; it stays in the zero-dimensional test so the d == 0 contract of the RVV-level dispatch remains covered through the fallback.

2. **Zvfhmin compile-time enforcement**: `sq-rvv.cpp` now `#error`s when neither `__riscv_zvfhmin` nor `__riscv_zvfh` is defined, and the FP16 `reconstruct_m8_components` is compiled unconditionally, so the kernel can no longer disappear silently — if the FP16 parity test executes, the RVV kernel is what ran. I verified the macro behavior with clang 22: `__riscv_zvfhmin` is defined under `-march=rv64gcv_zvfhmin` and `_zvfh`, and absent under plain `rv64gcv`; GCC defines it per the RISC-V C API since the extension landed (GCC 14, which CI uses).

CI is re-running on the branch.

mnorris11:
Hi @ihb2032 could you resolve merge conflicts? we are still working through the backlog of PRs

PR #5575, opened 2026-09-05
Let the binary Hamming paths reach the VPOPCNT kernels (#5575)

mnorris11:
Summary:

**TL;DR:** Points `IndexBinaryIVF` at the VPOPCNT level, instantiates its scanner in the translation unit compiled with `-mavx512vpopcntdq`, and documents the `scan_codes` heap contract.

Two things kept the binary Hamming paths on scalar popcount, whatever the host.

`IndexBinaryIVF` builds its scanner and runs its searches through `with_simd_level`, which uses the BASE level mask. That mask holds no `AVX512_VPOPCNT` bit, so on a host that reports that level the dispatch falls through to `AVX512`, where the wide computers use scalar popcount.

Two build variants, `faiss_omp_mock` and `faiss_no_multithreading`, hardcoded dynamic dispatch off. They compiled the AVX2 kernels only, so a consumer of either variant could not use runtime dispatch at all, whatever the build setting said.

This change:

- Moves the three dispatch sites in `IndexBinaryIVF.cpp` to `with_simd_level_with_vpopcnt`. Below that level the dispatch falls through to `AVX512`, as before.
- Instantiates `IndexBinaryIVF_impl.h` in `hamming_avx512_vpopcnt.cpp`. That is the one translation unit compiled with `-mavx512vpopcntdq`, which the VPOPCNT computers need. The include must follow the computer specializations.
- Lets both build variants honour the dynamic dispatch setting.
- Documents the `scan_codes` heap contract, and adds the first test for it.

The contract needed writing down because a caller that owns its heap has no way to discover it from the code. The heap top is the only bound, so a caller that wants the k nearest codes inside a radius seeds every slot with that radius in place of the neutral value. The heap must also be a max-heap: the scan accepts a code below `distances[0]` and calls `heap_replace_top` with `CMax`. The previous text said "min- or max-heap", which no implementation honours.

`scan_codes` also reads the heap top once into a local instead of on every iteration. Only an accepted code can lower it, so on a scan that rejects almost everything this turns a load into a register read per code.

`HammingComputer16` and `HammingComputer20` still inherit the scalar body at this level, so a 16-byte or 20-byte code gains nothing here.

Differential Revision: D118874433

### DIFF
--- a/benchs/bench_rabitq.py
+++ b/benchs/bench_rabitq.py
@@ -23,8 +23,8 @@ from faiss.contrib.datasets import SyntheticDataset
 #   d=512  -> size=64  -> 512-bit zmm, 1 iteration per bit-plane
 #   d=768  -> size=96  -> 512-bit zmm + 256-bit tail
 #   d=1024 -> size=128 -> 512-bit zmm only, 2 iterations per bit-plane
-# Sweeping these is useful for verifying the AVX512_SPR (vpopcntdq)
-# specialization in faiss/utils/simd_impl/rabitq_avx512_spr.cpp and for
+# Sweeping these is useful for verifying the AVX512_VPOPCNT specialization in
+# faiss/utils/simd_impl/rabitq_avx512_vpopcnt.cpp and for
 # profiling perf-record annotations across SIMD-width tiers.
 DIMENSIONS = [256, 512, 768, 1024]
 nlist: int = 1000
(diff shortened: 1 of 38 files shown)


mnorris11:
this merged in 3c7042f17a467c38d99f490ef39e4dfa615dc510, unsure why it didn't close

PR #5621, opened 2026-09-10
High Performance RaBitQ - HNSW RaBitQ support FP32 graph construction

lyang24:
this builds HNSW with FP32 distances so the graph has better neighbors, while search still uses the same quantized RaBitQ navigation. High quality graph lets queries reach the same recall with a lower efSearch

### DIFF
--- a/faiss/IndexHNSW.cpp
+++ b/faiss/IndexHNSW.cpp
@@ -69,7 +69,8 @@ void hnsw_add_vertices(
         size_t n,
         const float* x,
         bool verbose,
-        bool preset_levels = false) {
+        bool preset_levels = false,
+        const Index* construction_storage = nullptr) {
     size_t d = index_hnsw.d;
     HNSW& hnsw = index_hnsw.hnsw;
     size_t ntotal = n0 + n;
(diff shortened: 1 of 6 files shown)


faiss/IndexHNSW.cpp
```diff
@@ -1184,6 +1186,59 @@ IndexHNSWRaBitQ::IndexHNSWRaBitQ(
...
+    construction_storage.add(n, x);
+
+    storage->add(n, x);
+    ntotal = storage->ntotal;
+    // Set this before graph construction so an interrupted build cannot later
+    // append with a different construction-distance policy.
+    fp32_graph_built = n > 0;
+    bool preset_levels = hnsw.levels.size() == static_cast<size_t>(ntotal);
+
+    if (hnsw_deterministic_build) {
```
mnorris11, line 1215: @lyang24 btw, we just deleted the "non deterministic" build. It was intended to fall between releases so we don't leave the flag lying around. Does this mean the change to hnsw_add_vertices about the construction_storage needs to be added to the hnsw_add_vertices_deterministic?

lyang24: Rebased,  The deterministic builder already accepts distance-computer and query callbacks, so no signature change was needed.

PR #5622, opened 2026-09-10
High Performance RaBitQ - integer adc

lyang24:
This keeps the existing RaBitQ quantization but expands its codes into int8 values for faster search-time scoring. Queries are quantized once, and ARM CPUs use SDOT to score arbitrary document IDs efficiently. A scalar fallback and independent oracle tests verify the same arithmetic on other CPUs.


lyang24:
this one pre expand the sign bits and ex bits though throughput is good. The work will be superseded by https://github.com/facebookresearch/faiss/pull/5656  which does not carray 8 bits per dim overhead and still have a good edge over sq8 and sq4.

PR #5634, opened 2026-09-14
Add dimension-preserving block Hadamard transform

lyang24:
This adds a fast orthogonal transform that works for arbitrary vector dimensions without padding the full vector to a power of two. It can replace a dense random rotation when an index needs inexpensive feature mixing, and avoids increasing vector and code sizes for dimensions such as 768, 960, and 1536. This is useful for high-dimensional pipelines where a dense O(D²) transform is too expensive but preserving distances and the original dimensionality is important. The implementation includes inverse transforms, serialization, factory support, concurrent execution coverage, and numerical edge-case tests.


alexanderguzhva:
@lyang24 this one has weaker guarantees than a plain random matrix rotation or `2^n` FWHT.
Also, what is going to happen if the dimensionality is, say, 1000 ?

lyang24:
> this one has weaker guarantees than a plain random matrix rotation or `2^n` FWHT.
> Also, what is going to happen if the dimensionality is, say, 1000 ?

yea this one broken the randomess that makes the rabitq great
d=1000 is usual but 1000 = 512 + 256 + 128 + 64 + 32 + 8
the last 8 would give us problems

PR #5646, opened 2026-09-16
Update backward compatibility test to v1.15.1

alibeklfc:
## Summary
- Update backward compatibility test to use faiss-cpu=1.15.1

Note: the test was previously pinned to faiss-cpu=1.14.3, so this skips over 1.15.0.

**Do not land until faiss-cpu=1.15.1 is available on the pytorch conda channel** — the conda package is published by the tag-triggered release build. Landing earlier will break this workflow on main.

## Test plan
- [ ] CI passes
- [ ] Backward compatibility test uses correct version


alibeklfc:
**Unblocked — clear to land.**

The "do not land until faiss-cpu=1.15.1 is on the pytorch channel" note in the description is now satisfied. 1.15.1 published for linux-64 as `py3.12_hf65b397_0_cpu`, requiring `python >=3.12,<3.13.0a0` and `mkl >=2024.2.2,<2026` — matching what this workflow installs.

Both jobs are `runs-on: ubuntu-latest`, so only linux-64 is needed. The osx-arm64 conda package failed in the release run and is being rebuilt separately, but this workflow does not depend on it.

